PR #31, opened 2023-08-22
Biasdetection

j-space-b:
Added functionality for debias tests, using Dbias package as the rationale and this paper https://arxiv.org/pdf/2208.05777.pdf

Primarily just added bias_classifier.py and debias.md along with any dependencies 

Used same functionality as was in place for detoxify to keep consistent classifier functionality - also the paper seems to allow the 0.5 definition of bias in its description of incremental adversarial bias, so this makes sense.

Did not update test_utils.py or any other file since am not sure where this might fit in relation to other conversations of test cases against test datasets (and custom tests) - I just thought testing for bias was a cool addition here


ColabDog:
Thanks for this - bias detection in LLMs is definitely important. Thanks for the pull request. 
In future, I'll add a contributing.md but I'll clean it up for now and add tests.

PR #74, opened 2023-08-31
Add gpt synthetic data

penguine-ip:


ColabDog (changes requested):
Can you please: 
- Add a a test using OpenAI env variable to make sure it works. 
- Move hard-coded parameters to the top. 
- Build in re-try for OpenAI in case it fails

PR #118, opened 2023-09-22
Added answer_length.py

j-space-b:
Simply added the custom metric class for answer length to its own file in the metrics dir - didn't do anything to any other processes and not sure if it should be included anywhere other than here, for reference/use


j-space-b:
failed the lint formatting - lmk what you're using locally (if it matters) so I can edit and rerun but feel free to add anything you've been using on your end if no other changes

ColabDog:
@j-space-b Thanks! 
For linting (for now) we use black. Planning to do a major code clean-up including testing, mypy and ruff soon (ticketed up but unfortunately never got around to it).

j-space-b:
just reformatted - getting an error on the PR comment resource (below) - possibly due to git permissions - lmk what I can do to keep adding PRs, otherwise will let you take this one in Error: Resource not accessible by integration

ColabDog:
Thanks - should be able to ignore that github action - I'll look to remove it. It hasn't really been used and is more just a way to tell others what to install to test out deepeval!

ColabDog:
@j-space-b I see a few extra committed .idea files - are you able to remove those from the PR so we can merge this in? If it's easier - might be better to create a separate PR :)

j-space-b:
@ColabDog removed from this PR all good

ColabDog:
Thanks @j-space-b  

Just wondering with this implementation, it does a super simple count of the length of the number of characters. 
Curious on whether you think it's useful to also consider the following:

- If there's empty spaces before and after (or if this should be left to the user). I've seen lower-parameter GPT models have a lot of empty spaces before and after the text generation that could inflate this metric.
- Are there edge cases that length doesn't capture? I feel if code is being generated, len might not be the best metric? Could you specify in some documentation inside `docs/docs/metrics`  about the main use case of this? 

In addition, could you please also add the following?
- [ ] A test so we can capture the desired behaviour any ignore any undesired behaviour
- [ ] Any documentation outlining this example

j-space-b:
np ^ totally agree with the preceding/trailing spaces, good call, added that!  added documentation that stated original intent (to measure brevity along with any other metrics, but left to user to decide if relevant to other modalities e.g. code).  added to test_utils.py to test output.

ColabDog:
Oops - `test_utils` isn't really a test - I"ll make sure to rename that to avoid the confusion. All tests sit inside the `tests` folder. As an example  - I would recommend looking at `tests/test_bias.py` to see where the tests currently lie.

PR #167, opened 2023-10-02
add spport for 3-11

ColabDog:


ColabDog:
For python3.11 support to be added -will need to change Tensorflow installation

PR #177, opened 2023-10-05
Additional error handling for network issues in __init__py

### LINKED ISSUE #176: __init__.py issues when using behind vpn/firewall/corporate firewall
**Describe the bug**
When using the library with corporate firewall/vpn solutions (like fortinet) that do self-signed certificate-based traffic inspection, the `requests.get` request to pypi fails due to the self-signed certificate. (Most teams in these circumstances use a `pip.conf` file which solves this issue directly).

Additionally, this error would occur if you were using the library in a project where you don't actively have network connectivity (or in some CI environments that don't have access to the open internet). 

**To Reproduce**
1. use library behind corporate firewall or without internet access

**Expected behavior**
The library shouldn't break everything

Note: I have a PR linked to this issue

donaldwasserman:
Fixes #176

### DIFF
--- a/deepeval/__init__.py
+++ b/deepeval/__init__.py
@@ -11,7 +11,9 @@
 # )
 import warnings
 import re
+import ssl
 from ._version import __version__
+from requests.exceptions import SSLError, ConnectionError, HTTPError
 
 
 def compare_versions(version1, version2):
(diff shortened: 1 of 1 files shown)


ColabDog (approved):
Thanks! Didn't consider the edge case where `pypi` could be down too.

PR #185, opened 2023-10-06
Iimprove prompting flow for api key

penguine-ip:
Also removed implementation.

### DIFF
--- a/deepeval/cli/main.py
+++ b/deepeval/cli/main.py
@@ -26,10 +26,7 @@ def login(
         typer.Option(
             help="API key to get from https://app.confident-ai.com. Required if you want to log events to the server."
         ),
-    ] = "",
-    implementation_name: Annotated[
-        str, typer.Option(help="The name of your project")
-    ] = "",
+    ] = ""
 ):
     """Login to the DeepEval platform."""
     print("Welcome to [bold]DeepEval[/bold]!")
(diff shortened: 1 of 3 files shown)


deepeval/cli/main.py
```diff
@@ -26,42 +26,27 @@ def login(
...
         "Grab your API key here: [link=https://app.confident-ai.com]https://app.confident-ai.com[/link] "
     )
     if api_key == "":
-        api_key = KEY_FILE_HANDLER.fetch_api_key()
-        api_key = typer.prompt(
-            text="Paste it here (Hit enter if default is right)",
-            default=api_key,
-        )
+        while True:
+            api_key = input("Paste your API Key: ").strip()
```
jwongster2, line 38: Can we use the default api key that's stored using `KEY_FILE_HANDLER.write_api_key`?

ColabDog (changes requested):

deepeval/cli/main.py
penguine-ip: Why?

jwongster2: Hmm - wouldn't it be useful to have a default value if it's there? I'm not seeing why not.

penguine-ip: People run deepeval login to set their API key, if they wanted to just use the default value they wouldnt run the command in the first place, because the API key is already set. Also, including the default value means it will display this:
```
(venv) your/path
Paste your API Key [askdhjfbs87239r2g9]: 
```
The thing in [] is the API key which is confusing.

jwongster2: Got it - I understand.

PR #207, opened 2023-10-12
.

penguine-ip:


deepeval/test_quickstart.py
```diff
@@ -67,7 +67,7 @@ def test_customer_chatbot_simple():
         ],
     },
     {
-        "query": "Do you offer international shipping?",
+        "input": "Do you offer international shipping?",
         "output": "Yes, we offer international shipping to selected countries.",
```
jwongster2, line 71: Should also modify this to `actual_output`

PR #233, opened 2023-10-24
Keep local results file if not logged in

Andrea23Romano:
Local results file is deleted even if not logged in. In that case it could be useful to keep it.

### DIFF
--- a/deepeval/plugins/plugin.py
+++ b/deepeval/plugins/plugin.py
@@ -118,7 +118,7 @@ def pytest_sessionfinish(session: pytest.Session, exitstatus):
         print(
             "✅ Tests finished! View results on " f"[link={link}]{link}[/link]"
         )
-    os.remove(test_filename)
+        os.remove(test_filename)
 
 
 def pytest_terminal_summary(terminalreporter, exitstatus, config):
--- a/deepeval/plugins/plugin.py
+++ b/deepeval/plugins/plugin.py
@@ -1,4 +1,5 @@
 import pytest
+import shutil
 import os
 from rich import print
 from deepeval.api import Api, TestRun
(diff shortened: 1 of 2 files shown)


penguine-ip:
@Andrea23Romano thanks for the commit, I'm going to make this conditional on an environment variable to not pollute a user's directory

Andrea23Romano:
> @Andrea23Romano thanks for the commit, I'm going to make this conditional on an environment variable to not pollute a user's directory

I was thinking about adding one in deepeval.constants, is that ok?

penguine-ip:
@Andrea23Romano if that's the case and i'm not mistaken, a user will have to edit some deepeval source code to turn on/off saving the file, is that correct?

Andrea23Romano:
I don't think so, the user could simply add a statement like
```
deepeval.constants.PYTEST_LOCAL_RESULTS_FOLDER=<path>
```
in the test file. Could be None by default and, if not None, the file would be saved in that folder. I'm going to try it later and push it to this branch if everything works.

penguine-ip:
@Andrea23Romano can you please make it an env variable instead, constants are meant to be accessed by deepeval internally

Andrea23Romano:
Sure!

penguine-ip:
@Andrea23Romano thanks! appreciate it

Andrea23Romano:
@penguine-ip env variable and documentation example added. Let me know if it's ok or if I should change something!

penguine-ip:
@Andrea23Romano Looks good, going to merge after the tests completes, thanks! 😊

penguine-ip:
@Andrea23Romano I dont see why but deepeval login is failing in .github/workflows/deepeval-results.yml, any idea?

Andrea23Romano:
No idea, going to check it later.

Andrea23Romano:
Something's wrong with the action. In setup job "secret source" is set to None, so it can't find the deepeval api key. I'm not a github actions expert, but after some search seems like the secrets are not available for actions triggered by external forks, probably because of some "first time" PR restrictions you can probably modify. Check this, could be useful: https://docs.github.com/en/organizations/managing-organization-settings/disabling-or-limiting-github-actions-for-your-organization#configuring-required-approval-for-workflows-from-public-forks

penguine-ip:
@Andrea23Romano Hey, thanks for submitting the PR, join our discord if you haven't already, we're trying to build up the contributors community!

PR #237, opened 2023-10-25
Add: normalize_text function

Anindyadeep:
normalize_text is a very important function which is often used when calculating quasi metrics. Example: quasi_exact_match etc.

### DIFF
--- a/deepeval/utils.py
+++ b/deepeval/utils.py
@@ -1,4 +1,4 @@
-import re 
+import re
 import string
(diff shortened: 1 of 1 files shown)


penguine-ip:
The instructions for CONTRIBUTING.md isn't here yet but you'll have to run `black .` at the root directory for formatting.

Anindyadeep:
I see, sure.

PR #257, opened 2023-10-28
Exact Match metrics Tests

Anindyadeep:
- Adds the code for exact match metric.
- Adds tests for exact match metric.

### DIFF
--- a/deepeval/metrics/scores/calculator.py
+++ b/deepeval/metrics/scores/calculator.py
@@ -77,8 +77,19 @@ class StatisticalCalculator:
         )
 
     @classmethod
-    def exact_match(cls, target: str, prediction: str) -> float:
-        raise NotImplementedError()
+    def exact_match(cls, target: str, prediction: str) -> int:
+        """Metrics that calculates whether two sequences matches exactly or not. 
+
+        Args:
+            target (str): The target string.
+            prediction (str): The predicted string from the llm 
+
+        Returns:
+            int: The exact match score. 
+        """
+        if not prediction:
+            return 0
+        return 1 if prediction.strip() == target.strip() else 0
 
     @classmethod
     def quasi_exact_match(cls, target: str, prediction: str) -> float:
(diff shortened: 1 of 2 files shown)


penguine-ip:
I was thinking, exact match isn't a statistical method, would it make sense to have an additional class?

Anindyadeep:
> I was thinking, exact match isn't a statistical method, would it make sense to have an additional class?

Well, we can make, but you might not be having lot of methods under that class. Also experience wise, more the classes are, would that make more complex to use (users have to keep more number of classes in mind? just trying to keep things simple here), let me know what you think.

penguine-ip:
I think we either merge all the scores under one Calculator class, or make another class, since the one thing we wouldn't want is provide users with misleading information through the naming. I'd say one calculator class would be better, the way of computing scores can be in the docs and doesn't have to be reflected in the code.

Anindyadeep:
> I think we either merge all the scores under one Calculator class, or make another class, since the one thing we wouldn't want is provide users with misleading information through the naming. I'd say one calculator class would be better, the way of computing scores can be in the docs and doesn't have to be reflected in the code.

Yeah that quite makes sense. Since the way of doing the classification or doing the taxonomy is relative. I just want to somehow distinguish the fact that there are some metrics which uses some kind of models / or may be services and some which are deterministic functions. 

Because when metrics are deterministic functions, sometimes, people rely more on that. So another way of defining this notion could be:

For each function which is using a model it should have an Optional model parameter. And also, we can log that which model it is been using (if required), in that way people can still find that, which is using? 

There can be a third option and that is in documentation, where we can do the grouping.

penguine-ip:
Great idea, model as optional parameter + documentation is the way to go. There's also existing parts of the code we can abstract away using this approach, for example: https://github.com/confident-ai/deepeval/blob/main/deepeval/metrics/factual_consistency.py#L11

Anindyadeep:
> Great idea, model as optional parameter + documentation is the way to go. There's also existing parts of the code we can abstract away using this approach, for example: https://github.com/confident-ai/deepeval/blob/main/deepeval/metrics/factual_consistency.py#L11

Awesome :smile: , will go with this then

Issue #244, opened 2023-10-26
Metric Request: Bert score metrics


Anindyadeep:
By definition, we can define BERT score as:

> BERT Score assesses the quality of generated text by comparing it to a reference or a ground truth text. It considers the semantic similarity between the generated text and the reference text. BERT Score leverages BERT's contextual embeddings to capture the meaning and context of words and phrases in the text, which makes it more effective than traditional metrics like BLEU or ROUGE, which rely on n-grams and may not capture the nuances of language.

It will come under the ModelBasedMetric Category as defined on issue #238

PR #261, opened 2023-10-29
Feat: Added BertScore metrics 

Anindyadeep:
This PR adds:

- BERTScore as the model-based metrics.
- Adds tests for the same.

Addresses Issue #244


Anindyadeep:
Hey @penguine-ip, I added some new tests for I guess these CI checks might be using a big model (1.8 GB) to tests for BERTScore. Is that okay, before this will be for all the CI checks from now on.

Anindyadeep:
Oh, it runs lot of checks prior which also uses lot of models.

penguine-ip:
Yup it works :)

PR #265, opened 2023-11-02
update in depdency

Anindyadeep:


Anindyadeep:
Hey @penguine-ip, Can you please check the tests? Because although it is showing the pytest ran successfully, I am seeing there is one failed test which is not caught in the CI tests and marking that finally as a PASS.

penguine-ip:
Ye the test has to be rewritten, it is also not running the tests directory properly which is why you saw the errors earlier yesterday when running them locally. But it does seem like tensorflow and pytorch versions are troublesome atm, @Vasilije1990 is also working on the dependency issues and changing up the github test.yml files a bit as well

Anindyadeep:
Yes, that is awesome, no problem, right now let this be a draft PR and let @Vasilije1990 take the lead.

PR #291, opened 2023-11-16
Support for Vectara's hallucination evaluation model

agokrani:


penguine-ip:
@agokrani A few things here: 1) rename HallucinationEvaluationMetric to HallucinationMetric, 2) file name should match metric name (ie hallucination_metric.py), 3) @Anindyadeep is leading the effort to decouple scoring logic used in `def measure()` to a `Scorer` class in `./scoring.py`, please move the scoring logic (`VectaraHallucinationEvaluationModel`), and @Anindyadeep where do you think think the hallucination scorer should be at? Thanks!

deepeval/metrics/vectara_hallucination_evaluation.py
```diff
@@ -0,0 +1,64 @@
...
+
+        max_score = 0
+        for c in context_list:
+            score = self.model.predict(c, test_case.actual_output)
+            if score > max_score:
+                max_score = score
+
+        self.success = max_score > self.minimum_score
+        self.score = max_score
+        return max_score
```
Anindyadeep: Hi @agokrani,

Thanks for the awesome contribution. Just one thing, as @penguine-ip mentioned, can you move this part of the logic inside `metrics/scoring.py`. There you can see lot of other similar implementations present. 

The reason, we are trying to decouple scoring functions to `metric` is, scoring function should be granular enough, so that user can easily access those to make their own metric (which can be combination of their logic with some of our scoring logics) without getting coupled with deepeval's BaseMetric class functionalities. 

Let, me know if you need more help.

tests/test_hallucination_evaluation.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+def test_hallucination_evaluation_metric():
+    metric = HallucinationEvaluationMetric(minimum_score=0.5)
+    test_case = LLMTestCase(
+        input="placeholder",
+        actual_output="A blond drinking water in public.",
+        context=["A man with blond-hair, and a brown shirt drinking out of a public water fountain."],
+    )
+    assert_test(test_case, [metric])
```
Anindyadeep: Two things here:

1. Other than this test, please do provide some tests inside the scorer too `tests/test_scoring.py`.
2. And also, put some couple more tests, (like output type checks, output correctness etc)

deepeval/metrics/vectara_hallucination_evaluation.py
```diff
@@ -0,0 +1,64 @@
...
+    ):
+        # We use a smple cross encoder model
+        os.environ["TOKENIZERS_PARALLELISM"] = "false"
+        self.model = CrossEncoder(model_name)
+
+    def predict(self, text_a: str, text_b: str):
+        scores = self.model.predict([[text_a, text_b]])
+        # https://huggingface.co/vectara/hallucination_evaluation_model
+
+        return scores[0]
```
Anindyadeep: Same, here, try to move this scoring logic inside the `metrics/scoring.py`

Anindyadeep (changes requested):

deepeval/metrics/vectara_hallucination_evaluation.py
agokrani: Hi @Anindyadeep

Makes sense. However, I am still wondering where all the model related code would go. Should I just download the model there?

Anindyadeep: Yeah, I have similar thoughts on that. We might need to discuss more on this. For the current PR, we can simply have it inside the scorer engine. In the next set of iteration, we can come up with better structuring.

agokrani:
Hi @Anindyadeep, 

I have pushed some changes as we discussed on the call yesterday. Just check what it feels like to you. I guess everything should be working now. 

@penguine-ip  Please check also this commit I have moved some scoring model files here and there.

deepeval/metrics/hallucination_evaluation.py
```diff
@@ -0,0 +1,64 @@
...
+        minimum_score: float = 0.5,
+        model_name: str = "vectara/hallucination_evaluation_model",
+    ):
+        # For Crossencoder model, move to singleton to avoid re-instantiating
+
+        with progress_context(
+            "Downloading HallucinationEvaluationModel (may take up to 2 minutes if running for the first time)..."
+        ):
+            self.model = VectaraHallucinationEvaluationModel(model_name)
+        self.minimum_score = minimum_score
```
Anindyadeep: Thanks a lot for the changes, that we discussed beforehand. One small thing I would like to get changed, i.e., Since, you already put Hallucination Evaluation model inside `models/` folder, so can you import it here, instead of defining it here?

agokrani: Actually you are right, this shouldn't be here. We can now call it via scoring so why do I need this right? We have some duplication of score and metric. We need to take a deeper look at how to organize this. If we are doing score then do scores and not metrics. Or do we want to do both? 

@penguine-ip  your thoughts on this?

Anindyadeep: Right now we want to keep both. The notion now is the following:

1. decouple the scoring logic
2. decouple the models as you mentioned. 
3. Metrics should call the function from the metrics/scoring.py to do all the computation.

deepeval/models/hallucination_model.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class HallucinationModel(metaclass=Singleton):
+    def __init__(self, model_name: str = "vectara-hallucination"):
+        # We use a smple cross encoder model
+        os.environ["TOKENIZERS_PARALLELISM"] = "false"
+        model_name = name_to_card(model_name)
+        self.model = CrossEncoder(model_name)
+
+    def score(self, source: str, prediction: str):
```
penguine-ip: No need for this, scoring logic should be in scoring.py and the model module can be reserved for handling the initialization of a model instead.

penguine-ip (commented):
We can leave it in two separate PRs in future reviews. Also, I have seen users define their own custom hallucination metric and they expect the lower the score, the better. We can have both minimum_score and introduce a new maximum_score, only allow users to supply one, and handle the success logic in measure()

deepeval/metrics/hallucination_evaluation.py
penguine-ip: Yes so metrics will have `measure`, which imports a scorer from `scoring`. The main purpose of `measure` is to define what constitutes as a passing criteria for a given score (is it >=/=/<= minimum score, maximum score, etc), while the purpose of the scorer is to actually calculate the score itself. And of course if that requires a model we import it from `models`

deepeval/metrics/hallucination_metric.py
```diff
@@ -0,0 +1,43 @@
...
+        self,
+        minimum_score: float = 0.5,
+    ):
+        self.minimum_score = minimum_score
+
+    def measure(self, test_case: LLMTestCase):
+        if test_case.actual_output is None or test_case.context is None:
+            raise ValueError("Output or context cannot be None")
+
+        context_list = []
```
penguine-ip, line 22: No need for this, I'm guessing this was copied from another metric file, but those are outdated. Right now LLMTestCase.context is always a list of strings.

deepeval/metrics/scoring.py
```diff
@@ -201,6 +201,39 @@ def faithfulness_score(
...
+        """
+        try:
+            from deepeval.models.hallucination_model import (
+                HallucinationModel,
+            )
+        except ImportError as e:
+            print(e)
+        model = "vectara-hallucination" if model is None else model
+
+        # TODO: add this progress context in the correct place
```
penguine-ip: Can remove

deepeval/models/hallucination_model.py
penguine-ip, line 1: `progress_context` should be in this file, again I know for other metrics it's in another place but those are outdated and are to be refactored soon.

penguine-ip:
@agokrani Looking solid, left a few more comments, but otherwise good to merge!

deepeval/metrics/hallucination_metric.py
```diff
@@ -0,0 +1,43 @@
+import os
+from deepeval.singleton import Singleton
+from deepeval.test_case import LLMTestCase
+from deepeval.utils import chunk_text, softmax
+from deepeval.metrics.base_metric import BaseMetric
+from deepeval.evaluator import assert_test
+from deepeval.progress_context import progress_context
+from sentence_transformers import CrossEncoder
```
Anindyadeep: I guess, we do not need the `CrossEncoder` class import any more. So, I believe that can be removed.

Anindyadeep (approved):
LGTM. 

From the future repos, It would be great, if PRs like this can be broken into multiple small PRs. For example in this case, we can break this into two PRs as followed:

- PR1: Support for Vectera's Hallucination Evaluation Model.
- PR2: Adding `deepeval.models` module to register all the models for model based metrics. 

Something similar like the above would help us also to review easily and keep things more organized for future 😊.

deepeval/metrics/hallucination_metric.py
agokrani: I think we still need this because LLMTestCase.context can be None

deepeval/models/hallucination_model.py
agokrani: I have moved this as a comment inside the model class as TODO just to be reminded.

deepeval/metrics/scoring.py
agokrani: Removing this is causing problem. Let it be there for now.

deepeval/metrics/hallucination_metric.py
penguine-ip: But you don't have to check for list of strings

deepeval/models/hallucination_model.py
```diff
@@ -0,0 +1,18 @@
...
+from sentence_transformers import CrossEncoder
+from deepeval.models.model_map import model_map, name_to_card
+
+
+class HallucinationModel(metaclass=Singleton):
+    def __init__(self, model_name: str = "vectara-hallucination"):
+        # We use a smple cross encoder model
+        os.environ["TOKENIZERS_PARALLELISM"] = "false"
+
+        # TODO: add this progress context in the correct place
```
penguine-ip, line 13: quick question, how long does it take the model to load? Is there a reason why this is left commented out?

PR #305, opened 2023-11-23
Refactor DeepEval Scoring. 

Anindyadeep:
The DeepEval scoring is a bit cluttered, since we are not able to come up with a proper maintainable structure of different scoring logics (model dependent and model independent). So, this PR tries to refactor those and come up with a stable version of the Scoring engine. The structure of the Scoring engine is as followed:

1. A scoring engine `deepeval/scorer/` contains different scoring mechanism. (Right now it contains scoring.py, but more will be introduced soon). The scoring mechanisms are the most granular function which can be imported and used in any `Metric` implementation. 

2. Models `deepeval/models` contains a list of models, which will be used in scoring logics which requires a model to score. We are keeping the models separate, because the model implementation can change and also the models can also be called separately if required by the user.


penguine-ip:
Makes sense, agree with the abstraction.

penguine-ip:
@Anindyadeep Super well done, code is nice and clean and super easy to read, thank you!

Anindyadeep:
> @Anindyadeep Super well done, code is nice and clean and super easy to read, thank you!

Ayy, Thanks @penguine-ip, upnext documentation :)

PR #368, opened 2023-12-17
FEAT: Adding callbacks functionality

Pratyush-exe:
Introducing **DeepEvalCallback** for 🤗**huggingface** Trainer! ✨

1. Created DeepEvalCallback
2. Created RichManager for proper display of things (currently only for callbacks, can be added to rest deepeval code)
3. Added test script under tests folder.
4. Added retrieval context to Golden
5. Added optional param of List[Golden] to EvaluationDataset.

Issue: #383


tests/test_callbacks.py
```diff
@@ -0,0 +1,153 @@
...
+            "input_ids": x["input_ids"][0],
+            "labels": x["labels"][0],
+            "attention_mask": x["attention_mask"][0],
+        }
+    )
+    return dataset, merged_dataset, tokenized_dataset
+
+
+def create_deepeval_dataset(dataset, sample_size):
+    eval_dataset = [dataset[row] for row in range(5, 10)]
```
penguine-ip: Why the range (did you mean to use sample size instead?)

tests/test_callbacks.py
```diff
@@ -0,0 +1,153 @@
...
+    # Create Trainer instance
+    trainer = Seq2SeqTrainer(
+        model=model,
+        tokenizer=tokenizer,
+        args=training_args,
+        data_collator=data_collator,
+        train_dataset=tokenized_dataset["train"],
+    )
+
+    eval_dataset = create_deepeval_dataset(dataset["train"], sample_size=5)
```
penguine-ip, line 143: Is there a type definition for dataset["train"] for me to dig into? Or is it different depending on the dataset

penguine-ip (commented):
Just two nit questions, thanks again :))))

tests/test_callbacks.py
Pratyush-exe: `dataset` that I am passing here is the fine-tuning dataset itself. And the `sample_size` was meant to randomly pick that many data points from `dataset` for deepeval evaluation. That part lags, Will fix it today.

tests/test_callbacks.py
Pratyush-exe: Yeah it depends on the dataset, I am using 🤗's `datasets` library for loading and stuff (https://github.com/huggingface/datasets) which uses this type of partitioning.

tests/test_callbacks.py
penguine-ip: I see, thanks!

Pratyush-exe: Fixed this one. Please check it out!

penguine-ip:
@Pratyush-exe  LGTM, thanks!

PR #434, opened 2024-01-25
delete redundant "Toxicity"

nicholasburka:
as much as i like the song by System of a Down


penguine-ip:
Haha thanks for the correction!

PR #457, opened 2024-02-06
Removed macOS-specific .DS_Store file and updated .gitignore

kiselitza:
Your repo main root directory contains a macOS-specific file, which should be ignored.
This PR removes that file from repo codebase and updates .gitignore so future mac contributors don't introduce it again.


penguine-ip:
Thanks for the catch!

PR #480, opened 2024-02-15
Telemetry opt out

bderenzi:
Added an environment variable and documentation to optionally opt-out of the telemetry data collection.


penguine-ip:
thanks!

Issue #539, opened 2024-02-27
Multithreading error handling


AndresPrez:
**Describe the bug**
Some metrics compute a list of verdicts, statements or other things, in multithreading mode. When an error is raised from one of these child threads it is not being raised up to the parent thread, which makes the execution flow of the calling parent thread continue "normally" as it nothing happened. The main problem with this is that may go on with empty lists of verdicts, statements or whatever, thus resulting in an inaccurate evaluation of the test case.

**To Reproduce**
One common scenario I've been running into is when using a custom LLM or a GPT model instantiated with a low number of output max tokens as a parameter, thus increasing the probability of a json decode error due to the generation of an incomplete JSON output 

I know this can be avoided by using better LLMs or with longer output tokens, but I still consider that this should be better handled by properly raising the error up to the calling thread.

**Expected behavior**
Raise any errors originated from child thread to the calling thread.

<details>
<summary>Code Example</summary>

This is how multithreading is used right now to generate the list of verdicts for the ContextualRelevancyMetric
```python
    def _generate_verdicts_list(
        self, text: str, retrieval_context: List[str]
    ) -> List[List[ContextualRelevancyVerdict]]:
        verdicts_list: List[List[ContextualRelevancyVerdict]] = []
        threads = []
        lock = Lock()
        for context in retrieval_context:
            thread = Thread(
                target=self._generate_verdicts,
                args=(text, context, verdicts_list, lock),
            )
            threads.append(thread)
            thread.start()

        for thread in threads:
            thread.join()

        return verdicts_list
```

One way to catch any errors raised by one or many of these child threads could be done like the following:
```python
    def _generate_verdicts_list(
        self, text: str, retrieval_context: List[str]
    ) -> List[List[ContextualRelevancyVerdict]]:
        verdicts_list: List[List[ContextualRelevancyVerdict]] = []
        threads = []
        thread_exceptions = []  # List to store exceptions from threads
        lock = Lock()
        def thread_wrapper(func, args, exceptions_list, lock):
            try:
                func(*args)
            except Exception as e:
                # Lock is not strictly necessary here if you're only appending to a list,
                # but is good practice if you modify the shared data structure in other ways
                with lock:
                    exceptions_list.append(e)
        for context in retrieval_context:
            thread = Thread(
                # target=self._generate_verdicts,
                target=thread_wrapper,
                args=(self._generate_verdicts, (text, context, verdicts_list, lock), thread_exceptions, lock),
            )
            threads.append(thread)
[...]

Issue #538, opened 2024-02-27
Make multithreading optional


AndresPrez:
**Is your feature request related to a problem? Please describe.**
There's some multithreading logic inside some of deepeval's metrics. Some accept the multithreading flag to make that optional.

**Describe the solution you'd like**
Similar to what's already implemented for metrics `HallucinationMetric` and `FaithfulnessMetric`, it would nice to also have the ability to turn multithreading off for the rest of the metrics such as `ContextualRelevancyMetric`.

**Additional context**
The ask is simply because Im computing different metrics on different test cases that Im sending to multiple threads, but metrics such as `ContextualRelevancyMetric` is creating multiple threads itself from with a secondary thread.

penguine-ip:
^taking longer than expected, hopgin to have it out tomorrow

penguine-ip:
@AndresPrez just added this feature (v0.20.85). Switched over from threading to async execution. If you're using a custom model, you'll have to make tiny modifications to implement an async method to generate model outputs: https://docs.confident-ai.com/docs/metrics-introduction#using-a-custom-llm

Apart from this, you should see a lot of speed ups without having to make any changes.

penguine-ip:
@AndresPrez anticipating some bugs though, so please let me know if you run into any!

AndresPrez:
@penguine-ip I've looked at the code changes, looks good! I haven't had time to test it out yet but I will soon!

penguine-ip:
@AndresPrez thanks! Let me know if you run into any errors, one thing that some users are encountering is now rate limits since they run too concurrently...

EssamWisam:
> @AndresPrez thanks! Let me know if you run into any errors, one thing that some users are encountering is now rate limits since they run too concurrently...

Facing this issue with synthesizer when giving it an input of many documents to embed. The solution I did was synthesize subset by subset but it would be good for this to be handled internally.

penguine-ip:
@EssamWisam are you using your own model? Also if you come to our discord we will reply a lot faster.

PR #542, opened 2024-02-27
Implements function wrapper for thread error handling

AndresPrez:
tackles: https://github.com/confident-ai/deepeval/issues/539 and https://github.com/confident-ai/deepeval/issues/538


AndresPrez:
@penguine-ip here's my proposal, I'm not sure what those errors in the failed check are though

penguine-ip:
@AndresPrez Thanks! I have a much better idea how to implement it now. I'll merge it later today, then tag you on another PR that does the minor refactor. The failing tests is fine, it uses github secrets which only I have access to, once i merge it the workflow will pass :)

PR #549, opened 2024-03-01
add truthfulqa benchmark dataset

rohinish404:


penguine-ip:
^ Looking good so far @rohinish404 !

kritinv:
Hey @rohinish404 , good job on the work so far. I just have a couple of pointers that might help you further implement TruthfulQA. I saw that you defined the TruthfulQATask to consist of generation and multiple_choice. This means that you also need to define different metrics and templates for these 2 tasks in the predict method and generate_output method.  

### Suggestions on Metrics 

For multiple_choice, I think it makes a lot of sense to use exact_match_score (as you have currently done). 

For the generation task, however, the original paper used human evaluation, but also trained GPT-judge as a low-cost alternative (...‘We introduce a new metric for this purpose, which we call “GPT-judge”. GPT-judge is a GPT-3-6.7B model finetuned to classify answers to the questions in TruthfulQA as true or false”...). More details here https://arxiv.org/pdf/2109.07958.pdf) if you are curious.

In addition to GPT-judge that they mentioned in the paper, they defined a couple of other LLM-based metrics for generation https://github.com/sylinrl/TruthfulQA/tree/main. This includes GPT-sim, as well as BLEURT, ROUGE, and BLEU metrics, which are used to compare the model's answer to each of the true and false reference answers. The score is then given by [max similarity to a true reference answer] - [max similarity to a false reference answer]. 

You can read their implementation for all these metrics here [metrics.py](https://github.com/sylinrl/TruthfulQA/blob/main/truthfulqa/metrics.py), which they used here  [evaluate.py](https://github.com/sylinrl/TruthfulQA/blob/main/truthfulqa/evaluate.py/) for evaluation.

You may want to think about how to define the metrics in the most "accessible" way for users (i.e. using only GPT-3 instead of 5 models would mean only needing to supply 1 API_key), as well as how to abstract the code into DeepEval’s ecosystem (e.g. using DeepEval metrics to evaluate each prediction). 

### **Suggestions on Templating (template.py)**
For templating (prompting) both generation and multiple_choice, [utilities.py](https://github.com/sylinrl/TruthfulQA/blob/main/truthfulqa/utilities.py) and [presets.py](https://github.com/sylinrl/TruthfulQA/blob/main/truthfulqa/presets.py) defined what prompts they used and how they formatted it.  You may also want to look at https://github.com/confident-ai/deepeval/pull/574 for multiple choice templating, but I recommend you follow the original implementation as closely as possible.

_You picked a difficult benchmark, so keep up the good work! :)_

PR #593, opened 2024-03-13
Caching

kritinv:
### Key Contributions
- Implemented TestRunCacheManager() class to manage cached test_runs
- Integrated TestRunCacheManager to execute_test_run methods inside evaluate.py
- Edited indicator to show caching when necessary
Screenshot 2024-03-12 194342

### Future Considerations

- Caching affects the time metric! Currently, the time starts recording at the beginning of each test_run. As a result, cached test_runs will be much faster and won't reflect the actual time it takes to run all the metrics. I don't know how important this is, maybe it isn't at all, but this is solvable by simply recording the individual metrics' time, and then also cache that.
- Again, I'm not sure if this functionality was intentional, but I noticed that when we specify deepeval "-n 4", the circular indicators don't show like when we don't specify -n
- I implemented TestRunCacheManager() by saving api_test_cases instead of a dictionary of metrics_metadata. My reasoning is because we can easily access variables in the future if needed. However, this makes the code slightly chunkier, so idk if it was a good design decision.
- I need to further modularize the code! It's very chunky right now
- I only tested assert_test, haven't tested entire evaluations yet


deepeval/evaluate.py
```diff
@@ -103,43 +165,92 @@ def execute_test_cases(
...
-                score=metric.score,
-                threshold=metric.threshold,
-                reason=metric.reason,
-                success=metric.is_successful(),
-                evaluationModel=metric.evaluation_model,
-            )
-            api_test_case.metrics_metadata.append(metric_metadata)
+            metric_metadata = None
+            if metric.__name__ in cached_metrics_metadata:
+                metric_metadata = cached_metrics_metadata[metric.__name__]
```
penguine-ip: Check if evaluation model, threshold is the same as well

if include_reason: check if reason is available, if not compute metric again
if not include_reason: don't copy it

deepeval/evaluate.py
```diff
@@ -69,31 +71,96 @@ def create_api_test_case(
...
+        already_in_cache = (key in look_up_map)
+        hyperparameter_same = cached_test_run.hyperparameters == test_run.hyperparameters
+        actual_output_exists = False
+        metrics_metadata_exists = False
+        cached_api_test_case = None
+
+        if already_in_cache:
+            cached_api_test_case = look_up_map[key]
+            actual_output_exists = cached_api_test_case.actual_output != None
+            if actual_output_exists:
```
penguine-ip: Don't need this if statement

deepeval/evaluate.py
```diff
@@ -69,31 +71,96 @@ def create_api_test_case(
...
+        #########################################
+        # Get cached data if available
+        ##########################################
+        already_in_cache = (key in look_up_map)
+        hyperparameter_same = cached_test_run.hyperparameters == test_run.hyperparameters
+        actual_output_exists = False
+        metrics_metadata_exists = False
+        cached_api_test_case = None
+
+        if already_in_cache:
```
penguine-ip: Don't need this variable, use cached_api_test_case instead

deepeval/evaluate.py
```diff
@@ -69,31 +71,96 @@ def create_api_test_case(
...
+            cached_api_test_case = look_up_map[key]
```
penguine-ip: Do this on line 119, but use `lookup_map.get(key, None)` instead

deepeval/evaluate.py
```diff
@@ -69,31 +71,96 @@ def create_api_test_case(
...
+        look_up_map = cached_test_run.test_cases_lookup_map
+
+        #########################################
+        # Get cached data if available
+        ##########################################
+        already_in_cache = (key in look_up_map)
+        hyperparameter_same = cached_test_run.hyperparameters == test_run.hyperparameters
+        actual_output_exists = False
+        metrics_metadata_exists = False
+        cached_api_test_case = None
```
penguine-ip: cached_api_test_case = lookup_map.get(key, None)

deepeval/evaluate.py
```diff
@@ -69,31 +71,96 @@ def create_api_test_case(
...
+
+        test_run = test_run_manager.get_test_run()
+        cached_test_run = test_run_cache_manager.get_cached_test_run()
+        key = generate_cache_key(test_case=test_case)
+        look_up_map = cached_test_run.test_cases_lookup_map
+
+        #########################################
+        # Get cached data if available
+        ##########################################
+        already_in_cache = (key in look_up_map)
```
penguine-ip: Don't need this

deepeval/evaluate.py
```diff
@@ -69,31 +71,96 @@ def create_api_test_case(
...
+
+        if already_in_cache:
+            cached_api_test_case = look_up_map[key]
+            actual_output_exists = cached_api_test_case.actual_output != None
+            if actual_output_exists:
+                metrics_metadata_exists = cached_api_test_case.metrics_metadata != None
+        can_use_cache = (already_in_cache and hyperparameter_same and actual_output_exists and metrics_metadata_exists)
+
+        cached_metrics_metadata = {} 
+        if can_use_cache:
```
penguine-ip: use `lookup_map.get(key, None)`

deepeval/evaluate.py
```diff
@@ -69,31 +71,96 @@ def create_api_test_case(
...
+            metrics_metadata = look_up_map[key].metrics_metadata
+            cached_metrics_metadata = {metadata.metric: metadata for metadata in metrics_metadata}
```
penguine-ip: naming a bit confusing, the metrics metadata from lookup map is already the cached metrics metadata (make it cached metrics metadata map maybe?)

deepeval/evaluate.py
```diff
@@ -69,31 +71,96 @@ def create_api_test_case(
...
-            metric_metadata = MetricsMetadata(
-                metric=metric.__name__,
-                score=metric.score,
-                threshold=metric.threshold,
-                reason=metric.reason,
-                success=metric.is_successful(),
-                evaluationModel=metric.evaluation_model,
-            )
+            metric_metadata = None
+            if metric.__name__ in cached_metrics_metadata:
```
penguine-ip: Same here, (use `get`)

deepeval/evaluate.py
```diff
@@ -69,31 +71,96 @@ def create_api_test_case(
...
+                metric_metadata = cached_metrics_metadata[metric.__name__]
+            
+            # If same metric_metadata does not exist in cache 
+            # or threshold is not in same as in cached
+            # recomputer metric
+            new_threshold = metric.threshold
```
penguine-ip: No need another variable

deepeval/evaluate.py
```diff
@@ -69,31 +71,96 @@ def create_api_test_case(
...
+                metric.measure(test_case)
+                metric_metadata = MetricsMetadata(
+                    metric=metric.__name__,
+                    score=metric.score,
+                    threshold=metric.threshold,
+                    reason=metric.reason,
+                    success=metric.is_successful(),
+                    evaluationModel=metric.evaluation_model,
+                )
+                if cached_api_test_case:
```
penguine-ip: Clear cached_api_test_case.metrics_metadata at the very beginning right before this loop (not sure if you have already) so you 1) get rid of metrics metadata that are not part of the current test run, 2) don't duplicate metrics metadata.

deepeval/evaluate.py
```diff
@@ -69,31 +71,96 @@ def create_api_test_case(
...
-            )
+            metric_metadata = None
+            if metric.__name__ in cached_metrics_metadata:
+                metric_metadata = cached_metrics_metadata[metric.__name__]
+            
+            # If same metric_metadata does not exist in cache 
+            # or threshold is not in same as in cached
+            # recomputer metric
+            new_threshold = metric.threshold
+            if not metric_metadata or new_threshold != metric_metadata.threshold:
```
penguine-ip: Can abstract this logic in a helper function, since you also have to check for things like include_reason, same evaluation model, and strict mode

deepeval/evaluate.py
kritinv: Ah I didn't clear it because I assumed we wanted to cache as much info as possible for same metrics and different thresholds. I'll only make sure we cache the most recent test_run.

deepeval/test_run/api.py
```diff
@@ -1,14 +1,36 @@
...
+    strict_mode: bool
     metric: str
     score: float
     threshold: float
     success: bool
+    evaluation_model: Optional[str] = None
     reason: Optional[str] = None
-    evaluation_model: Optional[str] = Field(None, alias="evaluationModel")
+
+    # Optional attributes (Not every BaseMetric has these attributes)
```
penguine-ip: I wouldn't add metrics configurations such as `criteria`, `n`, `include_reason` here because `MetricsMetadata` maps 1:1 to the schema on Confident.

deepeval/cli/test.py
```diff
@@ -49,16 +49,22 @@ def run(
         "-n",
         help="Number of processes to use with pytest",
     ),
+    repeat: int = typer.Option(
```
penguine-ip: I updated this file for the new flag, i think the easiest for conflicts right now would just be to copy and paste the new file entirely then add the remaining logic for caching.

deepeval/evaluate.py
```diff
@@ -28,6 +31,37 @@ class TestResult:
...
+    STRICT_MODE = 'strict_mode'
+    THRESHOLD = 'threshold'
+    EVALUATION_MODEL = 'evaluation_model'
+
+    @classmethod
+    def values(cls):
+        return [member.value for member in cls]
+    
+# Not every BaseMetric has these attributes
+class MetricsOptionalParams(Enum):
```
penguine-ip: I think we wanted to merge the optional and required params

deepeval/evaluate.py
```diff
@@ -68,32 +102,108 @@ def create_api_test_case(
...
+        }
+
+    # Start with required attributes, assuming they must exist in metric
+    for attr in MetricsRequiredParams.values():
+        metadata_kwargs[attr] = getattr(metric, attr)
+    # Then create the optional attrevaluationModelibutes
+    for attr in MetricsOptionalParams.values():
+        if hasattr(metric, attr):
+            metadata_kwargs[attr] = getattr(metric, attr)
+    return MetricsMetadata(**metadata_kwargs)
```
penguine-ip: Left comment for MetricsMetadata below, might need additional logic to handle the configuration attributes that are not going to Confident

deepeval/test_run/cache.py
```diff
@@ -0,0 +1,149 @@
...
+from deepeval.test_run.api import APITestCase
+from deepeval.utils import delete_file_if_exists
+
+CACHE_FILE_NAME = ".deepeval-cache.json"
+TEMP_CACHE_FILE_NAME = ".temp-deepeval-cache.json"
+
+class CachedTestRun(BaseModel):
+    # metrics lookup map will lookup whether the existing metrics already exist according to
+    # the dictionary key str. The key str should be a recursively serialized llm test case params
+    test_cases_lookup_map: Optional[dict[str, APITestCase]] = Field(None, alias="testCasesLookupMap")
```
penguine-ip:
```python
class MetricConfiguration(BaseModel):
      pass

class CachedMetricsMetadata(BaseModel):
     metric_metadata: MetricsMetadata
     metric_configuration: MetricConfiguration

class CachedAPITestCase(BaseModel):
    cached_metrics_metadata: List[MetricsMetadata] = Field(
        ..., alias="metricsMetadata"
    )
```

PR #600, opened 2024-03-15
Fix fallback to dict if model_dump is not available

kubre:
I was not able to find tests around this to run but you can let me know if something is wrong or missing


penguine-ip:
@kubre Thanks looks good!

kubre:
Any idea in which version this fix will land?

PR #612, opened 2024-03-18
Fix typo in G-Eval

Peilun-Li:
I hope it's not a typo on purpose


penguine-ip:
@Peilun-Li definitely not, thanks!

PR #613, opened 2024-03-19
fix--fix several miss spells in the prompt templates

Kelp710:
### overview
Fix several missed spells in prompt templates to make the prompts generate more accurate outputs.


penguine-ip:
@Kelp710 Thanks!

PR #619, opened 2024-03-20
Support logprob based G-Eval

Peilun-Li:
As discussed in discord https://discordapp.com/channels/1118193057620316202/1217240554845245440 , to support logprob-based G-Eval scoring.

Notes:
1. I made it a strict opt-in feature to maintain backward compatibility.
2. It's fail-safe -- if logprob doesn't work it will fall back to normal one-shot score.
3. Also relaxed our Python version requirements a bit (as we are still using Python 3.9 -- no need to make it strict IMO as a generic SDK)
4. Have tested this on OpenAI but not on Azure OpenAI.


penguine-ip:
Thanks @Peilun-Li ! Take a few more days before getting to this PR, appreciate it :)

PR #623, opened 2024-03-23
Features/cleanupsynthesizer

penguine-ip:


deepeval/synthesizer/synthesizer.py
```diff
@@ -155,49 +154,68 @@ def generate_goldens(
...
+        if self.context_generator is None:
+            print("context generator")
+            self.context_generator = ContextGenerator(
+                docuemnt_paths,
+                chunk_size,
+                chunk_overlap,
+                multithreading=self.multithreading,
+            )
+
+        max_goldens_per_context = 2
```
penguine-ip: The default max goldens per context for doc goldens generation will be 2 for now. 

I first check if max goldens per doc < max goldens per context, which if it is, i should only generate 1 goldens per context.

Afterwards, I also divide max goldens per doc with max goldens per context to make sure num_context doesn't exceed max goldens per doc

deepeval/synthesizer/context_generator.py
```diff
@@ -2,120 +2,151 @@
...
         num_chunks = len(self.combined_chunks)
-        if num_context > num_chunks:
-            raise ValueError("Not enough chunks (" + str(num_chunks) + 
-                             ") to return the requested number of contexts ("  + str(num_context) + 
-                             "). Please decrease chunk_size or increase document length.")
-        clusters = self._get_n_random_clusters(n = num_context, cluster_size = max_context_size)
-        contexts = []
+        print(num_chunks, "Num chUNKS")
+        print(num_context, "Num CONTEXT")
+        num_context = min(num_context, num_chunks)
```
penguine-ip, line 39: Confine it instead of raising error, potentially here is causing the bug

deepeval/synthesizer/context_generator.py
```diff
@@ -2,120 +2,151 @@
...
+                self._get_n_other_similar_chunks(chunk, n=cluster_size - 1)
+            )
             chunk_cluster.append(cluster)
+        print(chunk_cluster)
         return chunk_cluster
 
     def _get_n_random_chunks(self, n: int) -> List[str]:
-        if not self.combined_chunks or len(self.combined_chunks) < n:
-            raise ValueError("Not enough chunks to return the requested number of random nodes.")
+        n = min(len(self.combined_chunks), n)
```
penguine-ip, line 100: Same here, confining it instead of raising a bug. I don't think here is where it is causing the extremely short context lengths.

penguine-ip (commented):
There are some context lengths where it is extremely short as shown in deepeval/results json files. Maybe it is due to the remaining text chunk being used as context, not 100% sure.

PR #650, opened 2024-04-01
Fix logprob-based G-Eval

Peilun-Li:
Fix issues of logprob-based G-Eval:
1. The `return_raw_response` parameter is no longer needed now that we introduced `generate_raw_response`
2. Fix the score value (`int(token_logprob["token"])` instead of `int(logprob)`). Also add a decimal check before the conversion.
3. Move the `math.log(0.01)` up before the for loop to reduce unnecessary re-computation

### DIFF
--- a/deepeval/metrics/g_eval/g_eval.py
+++ b/deepeval/metrics/g_eval/g_eval.py
@@ -295,10 +295,12 @@ class GEval(BaseMetric):
                 # Filter out non-decimal token to prevent errors in later int(token) conversion
                 if not token_logprob["token"].isdecimal():
                     continue
-        
+
                 # Calculate the linear probability
                 linear_prob = math.exp(logprob)
-                token_linear_probability[int(token_logprob["token"])] = linear_prob
+                token_linear_probability[int(token_logprob["token"])] = (
+                    linear_prob
+                )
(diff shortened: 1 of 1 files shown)


deepeval/metrics/g_eval/g_eval.py
```diff
@@ -284,16 +284,23 @@ def generate_weighted_summed_score(
...
             for token_logprob in score_logprobs["top_logprobs"]:
-                # Filter out tokens with <1% linear probability, i.e., logprobs < math.log(0.01)
                 logprob = token_logprob["logprob"]
 
-                if logprob < math.log(0.01):
+                # Filter out low probability tokens
+                if logprob < min_logprob:
+                    continue
+                # Filter out non-decimal token to prevent errors in later int(token) conversion
+                if not token_logprob["token"].isdecimal():
```
penguine-ip, line 296: Why not just use the `logprob` variable?

Peilun-Li: Oh basically the `token_logprob` dict looks like `{'token': '9', 'bytes': [57], 'logprob': -0.18066935},` `logprob` is the log-probability for the given `token`, so we'll need to retain both for calculating `token` * probabillity.

PR #663, opened 2024-04-03
ChromaDB Integration

kritinv:
Integrated chromadb to synthesizer (context_generator) to save embeddings for documents that have already been supplied to hidden folder.

Future Consideration

- [ ] Use search from chromaDB directly maybe faster
- [ ] Make chromaDB package optional
- [ ] Add a function to delete cached embeddings (but would this be necessary?)


penguine-ip:
hey @kritinv, what's the new algorithm with the Chroma implementation right now? (and also what was the old one before chroma)

I'm asking since the Chunk class that previously held embeddings might not be needed anymore, but not 100% sure.

PR #699, opened 2024-04-18
Support custom GPTModel arguments

Peilun-Li:
# Why 
In many cases during our evaluation, we want the evaluation results to be deterministic and reproducible, e.g., through passing `temperature=0` and a fixed `seed` to the model. This is not currently easily achievable using the vanilla `GPTModel` interface -- cutting out a custom model class could help unblock, but in that way 1) it's an overhead anyways, 2) also we'll lose all the benefits using a native model (cost reporting, logprob based g-eval, etc.).

# What
1. We actually already have `*args, **kwargs` placeholders in GPTModel's init function but they are unused currently. This PR just enables us to utilize them.
2. Also updating the native model checking logic, by honoring any GPTModel as a native model
3. Deduplicate the model initialization and native model checking logic by introducing a shared util function, for easier maintenance.

### DIFF
--- a/deepeval/metrics/answer_relevancy/answer_relevancy.py
+++ b/deepeval/metrics/answer_relevancy/answer_relevancy.py
@@ -6,6 +6,7 @@ from deepeval.metrics.utils import (
     validate_conversational_test_case,
     trimAndLoadJson,
     check_llm_test_case_params,
+    initialize_model,
 )
 from deepeval.test_case import (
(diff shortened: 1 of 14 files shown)


penguine-ip:
@Peilun-Li Looking good! Gonna test it out tomorrow and merge it, thank you so much!

penguine-ip:
All good, thanks!

PR #712, opened 2024-04-23
G_Eval cite params

kritinv:


penguine-ip:
@kritinv The g_eval.py file should keep initialize_model

PR #729, opened 2024-04-29
sentence -> reason to work with templates

lbux:
Simple change but precision and recall call for "reason" (and so does the actual chat template text for relevancy). The original example should work fine on its own but if you define a chat template to provide structed json to work for all 3 metrics, it breaks on relevancy due to example asking for "sentence" instead of "reason" thus confusing the model with conflicting instructions.


penguine-ip:
@lbux Thanks for catching this, much appreciated!

PR #735, opened 2024-05-01
Fixing JSON structure with LLM itself

sc-govsin:
Testing this approach to fix the JSON with another LLM call. This approach can be better but I'm just testing this as of now.


sc-govsin:
made pr on main repo by mistake.

penguine-ip:
@sc-govsin did this work?

sc-govsin:
@penguine-ip  for some LLMs it does (Mixtral), for some it doesn't (Command R).
Command R was generating garbage JSON.
Still there is small need of modification, I'll create a new PR once I complete the testing with few more LLMs.

PR #736, opened 2024-05-02
Remove "input" argument requirement for Faithfulness metric

philipchung:
The definition of the Faithfulness metric does not depend on Input.  I have checked the implementation code and prompts; the prompts do not appear to use input either.  Removing Input allows this metric to be used in contexts where I just want to check the claims from an output against a context (e.g. compare claims in actual output against claims from retrieval context).


penguine-ip:
Hey @philipchung, faithfulness isn't the only metric with input and actual output mandatory, and the reason these two parameters are mandatory is because we want to make this an LLM specific evaluation framework.

If you don't want input, you can provide an empty string "" instead, hope this helps!

PR #747, opened 2024-05-07
docs: added documentation to use Google Vertex AI Model with Deepeval

Adi8885:
Added documentation to use Google Vertex AI Model with Deepeval


penguine-ip:
@Adi8885 Thanks! We don't have vertex so really appreciate this.

PR #751, opened 2024-05-07
EvaluationDataset.save_as QoL Change

jakelucasnyc:
making EvaluationDataset.save_as return the file save path so it can be used later in the code


penguine-ip:
@jakelucasnyc tysm! Will be included in tonight's release

PR #778, opened 2024-05-21
Config Loading from YAML & README Update

deval-shah:
- Implemented YAML-based configuration loading for evaluation settings.
- Added files 
    - `deepeval/metrics/registry.py` to map metric names to class objects
    -  `deepeval/metrics/loader.py` to load metrics from yaml and initalize
- Updated README with instructions for using config.yaml.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -1,0 +1,29 @@
+metrics:
+  answer_relevancy:
+    threshold: 0.7
+    model: gpt-4-turbo
+    include_reason: true
+  faithfulness:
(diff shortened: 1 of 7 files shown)


config.yaml
```diff
@@ -0,0 +1,29 @@
+metrics:
+  answer_relevancy:
+    threshold: 0.7
+    model: gpt-4-turbo
```
nicoeiris11, line 4: Isn't `deepeval` using gpt4o by default?

PR #806, opened 2024-06-04
Stateless Metrics

kritinv:
### Stateless Support for Metrics

Stateless support was implemented for all metrics except RAGAS and knowledge retention. This involves adding context variables and changing the measure and a_measure accordingly. The measure method is more complex, because there is also support for async_mode, which involved implementing an additional function to overcome the loss of information due to using context variables in coroutines. In addition, the indicator was adapted to support a_measure.

### DIFF
--- a/deepeval/metrics/answer_relevancy/answer_relevancy.py
+++ b/deepeval/metrics/answer_relevancy/answer_relevancy.py
@@ -95,7 +95,7 @@ class AnswerRelevancyMetric(BaseMetric):
                 self.reason = self._generate_reason(test_case.input)
                 self.success = self.score >= self.threshold
                 if verbose:
-                    print(f"statements: {self.statements}\nverdicts: {self.verdicts}")                
+                    print(f"statements: {self.statements}\nverdicts: {self.verdicts}\n")                
                 return self.score
     
     async def _measure_async(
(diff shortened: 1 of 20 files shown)


deepeval/metrics/answer_relevancy/answer_relevancy.py
```diff
@@ -30,6 +31,13 @@ class AnswerRelvancyVerdict(BaseModel):
 
 
 class AnswerRelevancyMetric(BaseMetric):
+
+    _statements: ContextVar[List[str]] = ContextVar('statements', default=[])
```
penguine-ip: Is it possible for context variables to default to None? Since right now we default to none: https://github.com/confident-ai/deepeval/blob/main/deepeval/metrics/base_metric.py#L7

Also add `error` as well.

deepeval/metrics/bias/bias.py
```diff
@@ -203,11 +268,56 @@ def is_successful(self) -> bool:
             self.success = False
         else:
             try:
-                self.success = self.score <= self.threshold
+                self._success.set(self.score >= self.threshold)
```
penguine-ip: Careful copy n pasting (should be self.score <= self.threshold instead)

Issue #743, opened 2024-05-06
Cant use HuggingFace Model for evaluation


Kraebs:
When i follow the example on this page:
https://docs.confident-ai.com/docs/metrics-introduction 

and try to use Mistral-7B as evaluation-model, i always get this error when running the exact code in the tutorial. 
**It seems there is a mistake in the code when using HugigngFace models for evaluation instead of ChatGPT.**

Error:
---------------------------------------------------------------------------
JSONDecodeError                           Traceback (most recent call last)
File [~/.conda/envs/evaluation/lib/python3.12/site-packages/deepeval/metrics/utils.py:58](https://file+.vscode-resource.vscode-cdn.net/mnt/data/home/deest/Desktop/Code/Python/GitLab/pdf/~/.conda/envs/evaluation/lib/python3.12/site-packages/deepeval/metrics/utils.py:58), in trimAndLoadJson(input_string, metric)
     [57](https://file+.vscode-resource.vscode-cdn.net/mnt/data/home/deest/Desktop/Code/Python/GitLab/pdf/~/.conda/envs/evaluation/lib/python3.12/site-packages/deepeval/metrics/utils.py:57) try:
---> [58](https://file+.vscode-resource.vscode-cdn.net/mnt/data/home/deest/Desktop/Code/Python/GitLab/pdf/~/.conda/envs/evaluation/lib/python3.12/site-packages/deepeval/metrics/utils.py:58)     return json.loads(jsonStr)
     [59](https://file+.vscode-resource.vscode-cdn.net/mnt/data/home/deest/Desktop/Code/Python/GitLab/pdf/~/.conda/envs/evaluation/lib/python3.12/site-packages/deepeval/metrics/utils.py:59) except json.JSONDecodeError:

File [~/.conda/envs/evaluation/lib/python3.12/json/__init__.py:346](https://file+.vscode-resource.vscode-cdn.net/mnt/data/home/deest/Desktop/Code/Python/GitLab/pdf/~/.conda/envs/evaluation/lib/python3.12/json/__init__.py:346), in loads(s, cls, object_hook, parse_float, parse_int, parse_constant, object_pairs_hook, **kw)
    [343](https://file+.vscode-resource.vscode-cdn.net/mnt/data/home/deest/Desktop/Code/Python/GitLab/pdf/~/.conda/envs/evaluation/lib/python3.12/json/__init__.py:343) if (cls is None and object_hook is None and
[...]

PR #814, opened 2024-06-06
Improve loading of json if end of json is not found for metrics

Yleisnero:
Some LLMs struggle to generate a complete and correct JSON (e.g. mistral-7b-instruct). Also see issue #743.
However, often, only the closing bracket of the JSON is missing. This can be fixed easily by adding the bracket at the end of the input string.

### DIFF
--- a/deepeval/metrics/utils.py
+++ b/deepeval/metrics/utils.py
@@ -54,6 +54,10 @@ def trimAndLoadJson(
     end = input_string.rfind("}") + 1
     jsonStr = input_string[start:end] if start != -1 and end != 0 else ""
 
+    if start != -1 and end == 0:
+        input_string = input_string + "}"
+        end = len(input_string)
+
     try:
         return json.loads(jsonStr)
     except json.JSONDecodeError:


FaizaQamar:
Hey, I couldn't understand your solution. The same lines of code are in the utils.py file related to missing paranthesis. But the problem persists. Can you please explain your solution? @Yleisnero

Yleisnero:
@FaizaQamar I encountered the problem that Mistral was generating a correct json but was leaving out the closing curly bracket. So this solution only fixes this specific case.

MINJIK01:
I attached bracket at the end of the string, but it still doesn't work and reply same error.

I wrote like below.

test_case = LLMTestCase(
    input="Write me a joke}",
    actual_output="Why did the chicken go to the doctor? Because it had a fowl cough! I hope that made you cluck with laughter!",
)

Yleisnero:
This is not a fix that the user should apply. As @FaizaQamar mentioned, this code is in the utils.py because it is a merged pull request. This fixes some problems, but not everything. LLMs are not deterministic and unreliable. If you still encounter the issue, this here is not the solution.

PR #816, opened 2024-06-06
Improve loading of json if end of json is not found for metrics

Yleisnero:
Some LLMs struggle to generate valid jsons (e.g. mistral-7b-instruct). See also #743.
However, often, the problem is only that the closing bracket is missing. Adding the bracket helps using custom LLMs.


penguine-ip:
@Yleisnero Didn't know this was the case, thanks!

PR #857, opened 2024-06-28
Synthesize RED-TEAM

kritinv:


deepeval/dataset/dataset.py
```diff
@@ -404,58 +405,179 @@ def pull(self, alias: str, auto_convert_goldens_to_test_cases: bool = True):
...
+                _show_indicator=_show_indicator,
+                responses=responses,
+                use_case=use_case,
+            )
+        )
 
     def generate_goldens(
         self,
         contexts: List[List[str]],
-        synthesizer: BaseSynthesizer = None,
```
penguine-ip, line 419: Was this deleted because of circular import? Need to let users define their own synthesizer, can make it untyped `synthesizer = None` but check for type in the method using `is_instance(synthesizer, BaseSynthesizer)`

deepeval/dataset/dataset.py
```diff
@@ -404,58 +405,179 @@ def pull(self, alias: str, auto_convert_goldens_to_test_cases: bool = True):
...
                 _show_indicator=_show_indicator,
+                evolution_types=evolution_types,
+                use_case=use_case
             )
         )
 
     def generate_goldens_from_docs(
         self,
         document_paths: List[str],
-        synthesizer: BaseSynthesizer = None,
```
penguine-ip, line 445: Same here

deepeval/dataset/dataset.py
```diff
@@ -404,58 +405,179 @@ def pull(self, alias: str, auto_convert_goldens_to_test_cases: bool = True):
...
+                include_expected_output=include_expected_output,
                 max_goldens_per_document=max_goldens_per_document,
                 chunk_size=chunk_size,
                 chunk_overlap=chunk_overlap,
                 num_evolutions=num_evolutions,
                 enable_breadth_evolve=enable_breadth_evolve,
+                evolution_types=evolution_types
             )
         )
+        print(self.goldens)
```
penguine-ip: Rmb to delete

PR #871, opened 2024-07-08
get_model_name shouldn't be an async function

AndresPrez:
the class' DeepEvalBaseModel get_model_name function shouldn't be an async function since its called by the __init__ function of BaseMetric instances.


penguine-ip:
@AndresPrez tysm!

Issue #882, opened 2024-07-12
"deepeval login --api-key" throws error “TypeError: Option() missing 1 required positional argument: 'default’"


jerrydboonstra:
**Describe the bug**
`deepeval login --api-key VALIDKEY` fails with error:

```
Traceback (most recent call last):
  File "/opt/conda/bin/deepeval", line 5, in <module>
    from deepeval.cli.main import app
  File "/opt/conda/lib/python3.10/site-packages/deepeval/cli/main.py", line 19, in <module>
    typer.Option(
TypeError: Option() missing 1 required positional argument: 'default'
```

**To Reproduce**
Create and launch a docker image on a host with an Ampere GPU (A10g, ...) and `--platform linux/amd64`

Dockerfile
==
```
FROM ghcr.io/predibase/lorax:latest 
SHELL ["conda", "run", "-n", "base", "/bin/bash", "-c"]
WORKDIR /app
COPY requirements.txt .
RUN pip3 install --no-cache-dir --upgrade pip && pip3 install --no-cache-dir -r requirements.txt
COPY entrypoint.sh /deepeval-entrypoint.sh
ENV PYTHONUNBUFFERED 1
ENTRYPOINT [ "/deepeval-entrypoint.sh" ]
```

entrypoint.sh
==
```
#!/bin/bash
set -eEuo pipefail
deepeval login --api-key ${DEEPEVAL_KEY}
```

requirements.txt
==
```
deepeval==0.21.65
```

**Expected behavior**
Login success.

**Host:**
> uname -a
Linux b0c7edc50b19 5.15.0-1049-aws #54~20.04.1-Ubuntu SMP Fri Oct 6 22:04:33 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

> python -V
Python 3.10.14

> env
```
NVIDIA_VISIBLE_DEVICES=all
PYTHONUNBUFFERED=1
HOSTNAME=309a231f9f0a
[...]

PR #883, opened 2024-07-12
Patch for issue 882 "deepeval --login" fails

jerrydboonstra:
https://github.com/confident-ai/deepeval/issues/882

This is a misuse of the Annotations class according to GPT-4o, and i can't disagree since this patch fixes the issue in my non-working environment (docker/ubuntu), and doesn't regress in my working environment (mac m3).  It also passes `deepeval/cli/test.py`.  I did not run other tests.

### DIFF
--- a/deepeval/cli/main.py
+++ b/deepeval/cli/main.py
@@ -8,11 +8,12 @@ import webbrowser
 app = typer.Typer(name="deepeval")
 app.add_typer(test_app, name="test")
 
+
 @app.command()
 def login(
     api_key: str = typer.Option(
-        "", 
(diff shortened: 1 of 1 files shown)


jerrydboonstra:
Looking at the failing tests, 1. I didn't reformat w/ black and should have, 2. the annotation is used by the testing framework which is now failing because its missing? I'm surprised that there is application code that is only for testing purposes, unless i'm reading this wrong.

jerrydboonstra:
fixed the formatting issue.  the remaining test failures appear to be related to the test harness environment?

penguine-ip:
@jerrydboonstra  tysm! DW about the tests, permissions thing

PR #886, opened 2024-07-14
Another round of spelling and grammar updates in docs

callmephilip:


penguine-ip:
@callmephilip thanks again :)

PR #889, opened 2024-07-16
add pip install requirement to documentation for rouge_score

oftenfrequent:
I got caught starting a new project using `rouge_score`. The method assumes the user has installed the `rouge-score` package. If it is not included in the list of dependencies for this project, I think it should at least be included in the description of the method.


penguine-ip:
@oftenfrequent tysm!

PR #915, opened 2024-07-25
fix the bug that if one case occur error in one metric, all the rest …

wanghuanjing:
When run evaluation with function: deepeval.evaluate([testcases], [metrics]), if one case failed on one metric with some error, all the rest of cases will fail on that metric, debug and found the metric error is not reset when run every testcase, both the rest testcases get that error, my commit fixed that bug through reset the metric.error in evaluate function


penguine-ip:
@wanghuanjing thank you so much!

Issue #914, opened 2024-07-24
deepeval v0.21.71 - v 0.21.73 MMLU prediction results are wrong and the task scores and overall score are always 0 for Mistral7B and Gemma2B models


dequanchen:
**❗BEFORE YOU BEGIN❗**
Are you on discord? 🤗 We'd love to have you asking questions on discord instead: https://discord.com/invite/a3K9c8GRGt

**Describe the bug**
I have followed the page of "https://docs.confident-ai.com/docs/benchmarks-introduction" to do MMLU benchmark testing for Mistral-7B (mistralai/Mistral-7B-v0.1) (tasks=[MMLUTask.HIGH_SCHOOL_COMPUTER_SCIENCE, MMLUTask.ASTRONOMY]) and a full MMLU benchmark testing for Gemma-2B (google/gemma-2b) (all 57 tasks).  The Overall Score and task-specific scores for the two models are always 0 as follows:
(1) Mistral-7B model two task testing:
print("Overall Score:", benchmark.overall_score)
print("Task-specific Scores: ", benchmark.task_scores)
Overall Score: 0.0
Task-specific Scores:                             Task  Score
0  high_school_computer_science    0.0
1                     astronomy    0.0

(2) Gemma-2B model  57 tasks testing:
Overall MMLU Accuracy: 0.0
Overall Score:  0.0
Task-specific Scores:                                     Task  Score
0          high_school_european_history    0.0
1                       business_ethics    0.0
2                    clinical_knowledge    0.0
3                      medical_genetics    0.0
4                high_school_us_history    0.0
5                   high_school_physics    0.0
6             high_school_world_history    0.0
7                              virology    0.0
8            high_school_microeconomics    0.0
9                          econometrics    0.0
10             college_computer_science    0.0
11                  high_school_biology    0.0
12                     abstract_algebra    0.0
13              professional_accounting    0.0
14                           philosophy    0.0
15                professional_medicine    0.0
16                            nutrition    0.0
17                         global_facts    0.0
18                     machine_learning    0.0
19                     security_studies    0.0
20                     public_relations    0.0
21              professional_psychology    0.0
22                           prehistory    0.0
23                              anatomy    0.0
24                      human_sexuality    0.0
25                     college_medicine    0.0
26  high_school_government_and_politics    0.0
27                    college_chemistry    0.0
28                    logical_fallacies    0.0
29                high_school_geography    0.0
30               elementary_mathematics    0.0
31                          human_aging    0.0
[...]

Issue #819, opened 2024-06-08
accuracy always comes 0, might be a bug in my code i am unable to find.


AbhinavJangra29:
**❗BEFORE YOU BEGIN❗**
Are you on discord? 🤗 We'd love to have you asking questions on discord instead: https://discord.com/invite/a3K9c8GRGt

**Describe the bug**
i want to perform a mmlu benchmark of a opensource model:    https://huggingface.co/Telugu-LLM-Labs/Indic-gemma-2b-finetuned-sft-Navarasa-2.0
input format:
## Response: {response}

**To Reproduce**
here is the notebook: https://github.com/AbhinavJangra29/Abhinav-Jangra/blob/main/MMLU_navarasa_issue.ipynb

**Expected behavior**

**Screenshots**

![image](https://github.com/confident-ai/deepeval/assets/107471490/8ae12f81-b8c3-4683-a348-21d85028bc22)

![image](https://github.com/confident-ai/deepeval/assets/107471490/f1542173-f6c2-4f5b-a5b3-974eb0ba438f)

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

hubstrauss:
Same issue here. I looked a bit into it, what I am seeing so far in deepeval:
1. MMLU class uses an exact string match between the target and the prediction. However, seems like the prediction consists of the "whole" string, not only the predicted answer whereas the target is only the golden answer (and not the whole string)
See here: https://github.com/confident-ai/deepeval/blob/e2075deb431ba1f5a0f818b8773fa21a46ce49ed/deepeval/scorer/scorer.py#L95
2. The construction of prompt before generation seems weird - "confiment" is directly concatenated after the Response keyword
See here: https://github.com/confident-ai/deepeval/blob/e2075deb431ba1f5a0f818b8773fa21a46ce49ed/deepeval/benchmarks/mmlu/template.py#L21C1-L22C1

Side note: In your code @AbhinavJangra29, I think you could set max_new_token to a lower value, as MMLU is a multiple-choice question answering

PR #918, opened 2024-07-30
remove the weird prompt before generations

wjfu99:
Here is an example:
```
<s> The following are multiple choice questions (with answers) about high school computer science.

Which of the following is an example of the use of a device on the Internet of Things (IoT) ?
A. A car alerts a driver that it is about to hit an object.
B. A hiker uses a G P S watch to keep track of her position.
C. A refrigerator orders milk from an online delivery service when the milk in the refrigerator is almost gone.
D. A runner uses a watch with optical sensors to monitor his heart rate.
Answer: C

Many Web browsers allow users to open anonymous windows. During a browsing session in an anonymous window, the browser does not record a browsing history or a list of downloaded files. When the anonymous window is exited, cookies created during the session are deleted. Which of the following statements about browsing sessions in an anonymous window is true?
A. The activities of a user browsing in an anonymous window will not be visible to people who monitor the user's network, such as the system administrator.
B. Items placed in a Web store's shopping cart for future purchase during the anonymous browsing session will not be saved on the user's computer.
C. A user will not be able to log in to e-mail or social media accounts during the anonymous browsing session.
D. A user browsing in an anonymous window will be protected from viruses launched from any web sites visited or files downloaded.
Answer: B

What is the output of "abc"[::-1] in Python 3?
A. Error
B. abc
C. cba
D. c
Answer: C

Let x = 1. What is x << 3 in Python 3?
A. 1
B. 3
C. 8
D. 16
Answer:Output 'A', 'B', 'C', or 'D'. Full answer not needed.
```

The output prompt `Output 'A', 'B', 'C', or 'D'. Full answer not needed.` is redundent. LLM should start answer with `Answer:`.

### DIFF
--- a/deepeval/benchmarks/mmlu/template.py
+++ b/deepeval/benchmarks/mmlu/template.py
@@ -16,9 +16,6 @@ class MMLUTemplate:
         for i in range(n_shots):
             prompt += MMLUTemplate.format_question(train_set[i])
         prompt += input
-
-        # define ouptut confinement
-        prompt += "Output 'A', 'B', 'C', or 'D'. Full answer not needed."
         return prompt
 
     @staticmethod


wjfu99:
Some related issues
#819 
#914

penguine-ip:
@wjfu99 thanks!

PR #932, opened 2024-08-05
Fixed bug in ContextGenerator where it wouldn't generate the correct number of contexts + fix tests pipeline

fschuh:
`ContextGenerator.generate_contexts()` was not always generating the expected number of contexts due to the `num_contexts` parameter being overwritten, which affected subsequent generations in the for loop.  
The issue manifested itself when `num_chunks` was smaller than `num_context` (which happened very often on small documents).

Also added new unit test for the bugfix, and fixed and re-enabled all commented-out tests in `test_synthesizer.py`.

I also noticed the tests pipeline has been broken for a while, so I added some missing test dependencies to `pyproject.toml`, and made some changes on the following tests:

- `tests/test_hybrid_tracing.py`: refactored into a pytest test, and disabled due to missing data files
- `tests/test_json_metrics.py`: added `main` guard to prevent the test logic from running at import time


fschuh:
I fixed the tests, but it seems the pipeline is not setting the OPENAI_API_KEY environment variable which causes the tests to fail. Is this on purpose? If so, then I guess the tests cannot be run.

penguine-ip:
Hey @fschuh , thanks for the PR. Do you think you can revert the changes to the test files? Since those are things we use internally. Thanks!

penguine-ip:
@fschuh The test json metrics is fine but can you also revert synthesizer? We don't test synthesizer in our pipelines but just run it as python scripts to save token cost although it in the test directory. 

For the missing files, we have it locally which is why it is not available on GitHub. Thanks!

fschuh:
I've reverted `test_synthesizer.py`, and just added my new test to it. It's no longer using `pytest` now. Let me know if you need any further changes.

On a side note, for these tests that don't use `pytest`, I'd suggest moving them out of the tests folder or at least keep them separate from `pytest` tests, as some of those tests cause straight failures due to them not being written with `pytest` in mind. 

Also, when running `pytest`, it will attempt to open each of those files, and some of them have test logic that runs at import time. That means LLM calls being made just when importing anything from those files. 
Adding a `main` guard to `test_synthesizer.py` and `test_hybrid_tracing.py` would help with this issue.

Also, are there plans to get the CI pipeline green again? That would help with contributions to this repo tremendously.

penguine-ip:
@fschuh Thanks! Right now not a high priority for contributions. The failing test cases are normal due to permission issues.

PR #959, opened 2024-08-16
Bugfix (utils): Assign error string before raising

gCaglia:
Before, the check_llm_test_case_params function would raise an UnboundLocalError instead of the probably intended and better manageable ValueError.


penguine-ip:
@gCaglia thanks for catching it! :)

PR #1002, opened 2024-09-12
fix: fix concation issue using f string

Jinankhan:
What does it implement/fix?
The pull request serves as a fix for the str and int concat issue arising in the red_team python file ,thereby replacing the concatination with f string

Any other comments?
None


penguine-ip:
@Jinankhan fixed it + some migrations hence why didn't merge this PR, but it is out in the latest version v1.1.8!

PR #1035, opened 2024-09-25
Fix copied_turns in ConversationalTestCase

jaime-cespedes-sisniega:
There is a bug at the time of using `ConversationalTestCase` with more than one turn. Currently only the last turn is added to `copied_turns` list.

Moving this line inside the for loop (adding each turn to `copied_turns`) solves the bug.


penguine-ip:
@jaime-cespedes-sisniega thanks!

PR #1045, opened 2024-10-01
Added limit count parameter for faithfulness and summarization

lesar64:
Hi, small proposed parameter we customized in our implementation of deepeval, since especially faithfulness balloons to massive token sizes in large RAG-Systems and isn't able to finish the JSON because it tries to return too many claims and truths even with maximum max_token_limit, and GPT-4o completion incentivization. 

1.  Limit for claims:

        def generate_claims(text, limit_count: int = 0):

          if limit_count == 1:
  
              limit = " the single most important FACTUAL claim"
  
          elif limit_count > 0:
  
              limit = f" the {limit_count} most important FACTUAL claims"
  
          else: limit = " FACTUAL claims"
  
          return f"""Based on the given text, please generate a comprehensive list of{limit} that...

3.  Limit for truths:
    
        def generate_truths(text, limit_count: int = 0):
            
            if limit_count == 1:
    
                limit = " the single most important FACTUAL, undisputed truth"
    
            elif limit_count > 0:
    
                limit = f" the {limit_count} most important FACTUAL, undisputed truths per document"
    
            else:
    
                limit = " FACTUAL, undisputed truths"
    
            return f"""Based on the given text, please generate a comprehensive list of{limit}, that can...

4. Small typo fix in ContextualRelevancyTemplate:

> "When what was some of Einstein's achievements?"

to

> "What were some of Einstein's achievements?"


penguine-ip:
Thanks! @lesar64

PR #1074, opened 2024-10-11
Pass back test_run info with evaluate

john-lemmon-lime:
I'm automating a flow with DeepEval but need to publish a Slack message with the test run link. This passes back the test_run info when I used the evaluate function.

But this changes the interface of the evaluate function.


penguine-ip:
@john-lemmon-lime addressed and out in 1.3.6 :)

PR #1104, opened 2024-10-21
fix one more incorrect multistep_arithmetic_two key in schema.py

NikyParfenov:
Fixed bug with ONE MORE key in name 'multistep_arithmetic_two' in schema.py that causes an error when BigBenchHard.predict() falls into exception


penguine-ip:
@NikyParfenov thanks! Really appreicate it.

PR #1129, opened 2024-10-28
GEval docs: `strict` -> `strict_mode`

chkimes:
The docs as written lead to:

```
TypeError: GEval.__init__() got an unexpected keyword argument 'strict'
```


penguine-ip:
@chkimes  tysm!

PR #1152, opened 2024-11-13
pass test name to test result

AugmentMo:
Passing test name to test result.

Allows working with test names in the test results when using evaluate() function.


penguine-ip:
@AugmentMo  thx merged :)

PR #1180, opened 2024-11-22
fix: corrected unaligned attacks enum to match promptfoo

nabeel-chhatri:
- promptfoo generateHarmful  changed vulnerability type.
- swapped to new categories for unaligned attacks


penguine-ip:
@kritinv Is this safe to merge?

PR #1182, opened 2024-11-23
Fixed g-eval doc.

zyuanlim:
I was following the G-Eval guide for `Correctness` metric and could not get the evaluation to work correctly as it kept ignoring the `expected_output` in my LLMTestCase, despite the criteria or evaluation_steps explicitly stating to evaluate against the expected output (following exactly the example code). I realised that `LLMTestCaseParams.EXPECTED_OUTPUT` should be in the `evaluation_params` list, and after adding that, it worked correctly.


penguine-ip:
@zyuanlim Thanks so much! Didn't catch it previousluy :)

PR #1184, opened 2024-11-24
Update guides-red-teaming.mdx : Fix object name of load_model()

MANISH007700:
Changed 
```
model = self.load_model()
```
to
```
client = self.load_model()
```
as that's the object var name used to call the completion API of OpenAI. 

cc: @kritinv


MANISH007700:
@kritinv can I merge this ?

Issue #1171, opened 2024-11-21
Bug: dataset.py - All 4 Open statements in this file are lacking encoding="utf-8" argument


CAW-nz:
Originally I had identified that the two Open statements in EvaluationDataset.save_as() are lacking encoding="utf-8" argument - refer to the below information.  However further testing has shown that **ALL 4 file Open() statements need this encoding argument**.

The equivalent two Open statements in Synthesizer.save_as() have this argument.
The lack of **encoding="utf-8"** argument is causing issues in the output of the json and csv files when calling the dataset version.

For example **synthesizer.save_as**(file_type='json',directory=out_dir) produced the following output snippet for context:
> "context": [
            " for achieving \nambitious climate change mitigation goals and climate resilient \ndevelopment ( high conﬁdence ). Climate resilient development is \nenabled by increased international cooperation including mobilising \nand enhancing access to ﬁnance, particularly for developing countries, \nvulnerable regions, sectors and groups and aligning ﬁnance ﬂows \nfor climate action to be consistent with ambition levels and funding \nneeds ( high conﬁdence )."

whereas when converted to a dataset via

`from deepeval.dataset import EvaluationDataset`
`dataset = EvaluationDataset(goldens=synthesizer.synthetic_goldens)`
`dataset.save_as(file_type='json',directory=out_dir)`

for the same data, **dataset.save_as**() gave...

>"context": [
            " for achieving \nambitious climate change mitigation goals and climate resilient \ndevelopment ( high con\ufb01dence ). Climate resilient development is \nenabled by increased international cooperation including mobilising \nand enhancing access to \ufb01nance, particularly for developing countries, \nvulnerable regions, sectors and groups and aligning \ufb01nance \ufb02ows \nfor climate action to be consistent with ambition levels and funding \nneeds ( high con\ufb01dence )."

All instances of words starting with "fi" (which was actually stored as a single character "ﬁ") became "\ufb01" and "fl" (actually single character "ﬂ") became "\ufb02".  This happened throughout the entire json file.
Another character that isn't being handled is "1.5°C" which appears as "1.5\u00b0C" in the Dataset version but properly output in Synthesizer version.

The call with **dataset.save_as**(file_type='**csv**') is even failing for my data probably due to this issue (whereas **synthesizer.save_as**(file_type='**csv**') is successful).
Bottom of trace dump=
[...]

PR #1196, opened 2024-11-28
Fix issue #1171

CAW-nz:
This is the PR to fix Issue #1171 

Only changes in save_as() method are needed as the two open("r") versions now have the required encoding argument (with "utf-8" as the default string).

I've not added encoding_type string argument like has been done for the two open("r") versions in this file.
Can be added if you wish, but then ensure_ascii maybe should be too.
Plus if we add encoding_type (and ensure_ascii) here then we probably should do the same in synthesizer save_as method too.
Will let you decide how much is needed.

At this point all I've done is **exactly mirror the code** that exists in synthesizer.save_as().

### DIFF
--- a/deepeval/dataset/dataset.py
+++ b/deepeval/dataset/dataset.py
@@ -778,7 +778,7 @@ class EvaluationDataset:
         full_file_path = os.path.join(directory, new_filename)
 
         if file_type == "json":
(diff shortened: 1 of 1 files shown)


penguine-ip:
Hey @CAW-nz I think this is good enough for now, thanks!

PR #1208, opened 2024-12-04
Fixes #1207

### LINKED ISSUE #1207: ModuleNotFoundError: No module named 'nest_asyncio'
**Describe the bug**
I encountered a `ModuleNotFoundError: No module named 'nest_asyncio'` error after installing the `deepeval` library. It seems that `nest_asyncio`, which is required by the library, is not automatically installed as a dependency. 

**To Reproduce**
Create a new virtual environment:
```bash
python3 -m venv .venv  
source .venv/bin/activate 
```
Install uv & deepeval
```bash
curl -LsSf https://astral.sh/uv/install.sh | sh
uv init
uv add deepeval
```
Run a python file with a deepeval import (`python3 main.py`)
```python
import deepeval
```

**Potential Cause:**
I'm no python expert, but the `nest_asyncio` package is not included in the `install_requires` section of `setup.py`. Might this be the root cause of the issue, as `nest_asyncio` is required by `deepeval` but not automatically installed unless explicitly listed in `install_requires`.

**Additional context**
I'm using UV as a package manager and python 3.12.3.

kbarendrecht:
I encountered a `ModuleNotFoundError: No module named 'nest_asyncio'` error after installing the `deepeval` library. It seems that `nest_asyncio`, which is required by the library, is not automatically installed as a dependency.

see #1207

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -157,7 +157,7 @@ cython_debug/
 #  be found at https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore
 #  and can be added to the global gitignore or merged into this file.  For a more nuclear
 #  option (not recommended) you can uncomment the following to ignore the entire idea folder.
-#.idea/
+.idea/
 docs/.docusaurus
 node_modules
 .deepeval
--- a/setup.py
+++ b/setup.py
@@ -36,6 +36,7 @@ setup(
         "opentelemetry-sdk>=1.24.0,<2.0.0",
         "opentelemetry-exporter-otlp-proto-grpc>=1.24.0,<2.0.0",
         "grpcio==1.60.1",
+        "nest-asyncio",
     ],
     extras_require={
         "dev": ["black"],


kbarendrecht:
@kritinv @penguine-ip would you mind taking a look? it's a tiny change.

penguine-ip:
@kbarendrecht hey thanks for the bump I missed it - one quick question what's the ./idea folder in gitignore?

kbarendrecht:
Hi @penguine-ip, that's the configuration folder for Jetbrains IDE's. it creates it on opening a folder automaticly. I didn't want to force my settings onto anyone. It's recommened to exclude the entire thing or parts of it.

@see https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore#L4

penguine-ip:
@kbarendrecht oh no not at all was just curious about the extra directory, thanks for the PR

PR #1218, opened 2024-12-09
Feature/synthesis cost

vjsliogeris:
I was wondering how much it cost for me to synthesize goldens. I found that the costs were generated, but not used, so I created a variable and put those costs there. Found it weird that it wasn't in the main repo so creating a PR.
Open to improve this.
Open to hearing out if this doesn't belong here.

Copied this from the metric costs (in this case from [g_eval](https://github.com/confident-ai/deepeval/blob/main/deepeval/metrics/g_eval/g_eval.py)) , where the costs variable (originally `evaluation_cost`, here `synthesis_cost`) gets instantiated/created on `measure` and `a_measure`, and the `cost` from all model calls just gets added.


penguine-ip:
Hey @vjsliogeris looks amazing, thanks again!

pazevedo-hyland:
@vjsliogeris , this change makes it so that you can't support custom models

For example:
```
from litellm import acompletion, completion
import instructor


class LiteLLM(DeepEvalBaseLLM):
    def __init__(
        self,
        model="anthropic.claude-3-haiku-20240307-v1:0",
    ):
        self.model = model
        self.client = instructor.from_litellm(completion)

    def load_model(self):
        return self.model

    def generate(self, prompt: str, schema: BaseModel) -> BaseModel:
        messages = [{"content": prompt, "role": "user"}]
        response = self.client.chat.completions.create(
            model=self.model, messages=messages, response_model=schema
        )
        return response

    async def a_generate(self, prompt: str, schema: BaseModel) -> BaseModel:
        return self.generate(prompt, schema)

    def get_model_name(self):
        return "LiteLLM Model"
```

image

vjsliogeris:
@pazevedo-hyland Thanks for reporting this bug. I'll look into it when I have some time.

PR #1222, opened 2024-12-12
Add doc count to the ContextualPrecision template (Fixes #1205)

### LINKED ISSUE #1205: ContextualPrecision endless generation
**Describe the bug**
For longer/more complex contexts, my evaluator stalls in the `ContextualPrecision._generate_verdicts()` in an endless generation loop. I can reproduce this in an LLM playground but unfortunately I can't share the exact contexts here.

I'm 'only' using llama 3.1 70B as evaluator, so using a more capable model would probably help, but I was actually able to resolve this by simply adding the number of nodes to the prompt *(this has helped me before with similar issues)*. 

I'm happy to submit a PR containing the change, then you can decide whether to integrate or not. Are you interested?

enrico-stauss:
# Resolve Endless Generation Issue (ContextualPrecision Verdicts)

As described in issue #1205, for longer lists of complex contexts, (less capable) LLMs could loose track of the document count and therefore enter an endless generation loop. This can often be resolved by simply passing the explicit document count as part of the prompt.

### DIFF
--- a/deepeval/metrics/contextual_precision/template.py
+++ b/deepeval/metrics/contextual_precision/template.py
@@ -1,6 +1,9 @@
 class ContextualPrecisionTemplate:
     @staticmethod
-    def generate_verdicts(input, expected_output, retrieval_context):
+    def generate_verdicts(input, expected_output, retrieval_context: str | list):
+        document_count_str = ""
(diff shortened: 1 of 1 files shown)


penguine-ip:
@enrico-stauss thanks!

Issue #1169, opened 2024-11-19
Use Poetry build instead of setup.py


FrancoisMasson1990:
**❗BEFORE YOU BEGIN❗**
Are you on discord? 🤗 We'd love to have you asking questions on discord instead: https://discord.com/invite/a3K9c8GRGt

**Describe the bug**

There is a discrepancy between the version of some of the libraries you use in setup.py vs the ones in pyproject.toml
Solution;
Use poetry build when generatin new release of the lib

Also can you accept Python3.12 version ?

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

penguine-ip:
@FrancoisMasson1990 python 3.12 ok in latest release :)

PR #1223, opened 2024-12-12
GH-1169: sync dependencies between setup py and pyproject toml, remove setup py

ChristianMurphy:
setup files will automatically be taken care of by running
```
poetry build
poetry publish
```

or both at once with

```
poetry publish --build
```

source: https://python-poetry.org/docs/cli#publish
resolves: GH-1169


penguine-ip:
Hey @ChristianMurphy thanks for the PR but we don't want to move it for now since there might be breaking changes and we don't have time to test everything right now. Closing this PR for now since the main branch is deviating too much and leaving this issue for next month.

Issue #99, opened 2023-09-11
Improve typing (Mypy) support


ColabDog:

jyyogi:
Is this implemented? I don't see a `py.typed` file nor does my mypy linter recognize `deepeval`:

```shell
error: Skipping analyzing "deepeval": module is installed, but missing library stubs or py.typed marker  [import-untyped]
```

PR #1224, opened 2024-12-12
GH-99: include types and check types

ChristianMurphy:
builds on: GH-1223
resolves: GH-99 
source: https://safjan.com/the-importance-of-adding-py-typed-file-to-your-typed-package/

note:
The types aren't currently checks so it is expected that there are some errors in the type check task.
Adding this as a check is a step towards making the types valid, and they can be progressively updated until the check does pass.


penguine-ip:
@ChristianMurphy tbh i''m not too sure what this is so i'm afriad of merging it, will close this for now

ChristianMurphy:
@penguine-ip the goal is to support https://docs.python.org/3/library/typing.html so that consumers of this library using get typeahead competition when using the deepeval API in an IDE, and can validate they are using the types correctly with a checker like https://www.mypy-lang.org/, https://github.com/microsoft/pyright, or https://github.com/facebook/pyre-check

To do that this project needs to:
1. Pick a packaging tool (there are several in the code right now ans its unclear which actually publishes), I'd recommend poetry https://github.com/confident-ai/deepeval/pull/1223 or UV, but am open to maintainer suggestions .
2. For that package manager include a py.typed file https://safjan.com/the-importance-of-adding-py-typed-file-to-your-typed-package/ so downstream packages know to look for types
3. (Optional) check the internal code and tests with a type checker to ensure the types shared are accurate

alechoey:
Seconding that it would be great to have better type checking.

PR #1229, opened 2024-12-17
fix: MMLUTemplate.format_subject should be a static method

Jerry-Terrasse:
This appears to be a minor oversight and can be resolved by simply adding the `@staticmethod` decorator.


penguine-ip:
@Jerry-Terrasse tysm :)

Issue #1227, opened 2024-12-13
Avoid unnecessary opentelemetry imports


volodymyr-lotos:
**Is your feature request related to a problem? Please describe.**
Framework imports `opentelemetry` dependencies even when `telemetry_opt_out` is evaluated to `True`. It causes some conflicts with other libraries used on my project.

**Describe the solution you'd like**
Import `opentelemetry` only after validation that telemetry should be sent.

**Describe alternatives you've considered**
N/A

**Additional context**
Problems are related to protobuf versions that are used in opentelemetry.
```TypeError: Couldn't build proto file into descriptor pool: duplicate file name opentelemetry/proto/common/v1/common.proto```

PR #1231, opened 2024-12-17
Ensure opentelemetry only loaded if telemetry is not opted outed

chaliy:
Fixes https://github.com/confident-ai/deepeval/issues/1227

 Details
 
 - opentelemetry is loaded during deepeval import, because opentelemetry loads protobufs it conflicts with other libraries
 - PR makes sure that opentelemetry is only loaded when Telemetry is enabled


penguine-ip:
@chaliy thanks!

PR #1247, opened 2025-01-03
guardrails

kritinv:


penguine-ip (commented):
Is there a need to include input, response, etc. in the guard score?

PR #1256, opened 2025-01-09
Fix import path of sensitive information disclosure attacks

mrazizi:
Fix import path of PIILeakageType, PromptLeakageType and IntellectualPropertyType


penguine-ip:
@mrazizi thanks!

Issue #843, opened 2024-06-21
Invalid value for 'top_logprobs': must be less than or equal to 5 for AzureOpenAi Models


Neeraj319:
**❗BEFORE YOU BEGIN❗**
Are you on discord? 🤗 We'd love to have you asking questions on discord instead: https://discord.com/invite/a3K9c8GRGt

**Describe the bug**
The value range for `top_logprobs` with azureopenai is 0-5, but we are sending `20` as the default hardcoded value in `metrices/g_eval/g_eval.py` 
This raises the following exception. 
```
Traceback (most recent call last):
  File "/home/human_object/programming/dobby-be/src/deepeval_test.py", line 44, in <module>
    metric.measure(test_case)
  File "/home/human_object/.pyenv/versions/3.11.9/lib/python3.11/site-packages/deepeval/metrics/g_eval/g_eval.py", line 132, in measure
    ) = loop.run_until_complete(self._measure_async(test_case))
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/human_object/.pyenv/versions/3.11.9/lib/python3.11/asyncio/base_events.py", line 654, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/home/human_object/.pyenv/versions/3.11.9/lib/python3.11/site-packages/deepeval/metrics/g_eval/g_eval.py", line 196, in _measure_async
    await self.a_measure(test_case, _show_indicator=False)
  File "/home/human_object/.pyenv/versions/3.11.9/lib/python3.11/site-packages/deepeval/metrics/g_eval/g_eval.py", line 174, in a_measure
    g_score, reason = await self._a_evaluate(test_case)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/human_object/.pyenv/versions/3.11.9/lib/python3.11/site-packages/deepeval/metrics/g_eval/g_eval.py", line 255, in _a_evaluate
    res, cost = await self.model.a_generate_raw_response(
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/human_object/.pyenv/versions/3.11.9/lib/python3.11/site-packages/tenacity/_asyncio.py", line 88, in async_wrapped
    return await fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1261, opened 2025-01-10
top_logprobs configurable

derickson:
Related to https://github.com/confident-ai/deepeval/issues/843

|Azure deployments have a version.
not all version of gpt-4 and gpt-4 on azure support logprobs and top_logprobs at all
when they do, the top_logprobs setting can only go to 5 instead of 20 on OpenAI service

for Azure OpenAI gtp-40 version 2024-11-20, I am seeing the behavior that azure is taking commands to 20 and then internally changing it to 5 (based on the test results), so this might not be necessary, but there are other Azure versions out there I can't test against and this might be useful to others who can't upgrade.

tested against your pytest with LLM skipped
tested against OpenAI service ... no problems
tested against Azure gtp-40 version 2024-11-20 ... no problems


penguine-ip:
hey @derickson thanks really appreciate it, will cut a release for it asap :)

Issue #1262, opened 2025-01-10
metric.evaluate() raise  TypeError: unsupported operand type(s) for +=: 'NoneType' and 'float'


rawheel:
**Describe the bug**
TypeError in metric.evaluate() when called directly.

When the evaluate() method of a metric (derived from BaseMetric) is called directly (without first calling measure()), it throws a TypeError: unsupported operand type(s) for +=: 'NoneType' and 'int/float'. This is because the evaluation_cost attribute in the BaseMetric class is initialized to None.

During the evaluation process, the code attempts to increment evaluation_cost (e.g., evaluation_cost += cost). However, since evaluation_cost is None by default, this operation results in the TypeError.

Expected Behavior: evaluation_cost should be initialized to 0 to allow numerical additions.

**To Reproduce**
1. Instantiate a metric (e.g., metric = GEval()).
2. Call metric.evaluate(test_case) directly.

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**
 - OS: iOS
 - Browser: Chrome

**Additional context**

jbbqqf:
Hi! Triaging older issues — I think this can be closed: the `NoneType` `+=` crash from `evaluation_cost` is now guarded centrally.

Evidence:
- `BaseMetric._accrue_cost` ([`deepeval/metrics/base_metric.py:61-65`](https://github.com/confident-ai/deepeval/blob/ec8ea66/deepeval/metrics/base_metric.py#L61-L65)) now guards both sides of the addition:
  ```python
  def _accrue_cost(self, cost: float) -> None:
      if self.evaluation_cost is not None and cost is not None:
          self.evaluation_cost += cost
      else:
          self.evaluation_cost = None
  ```
  So calling a metric's evaluate path before `measure()` initialises `evaluation_cost = 0`, or with a custom LLM that returns `cost is None`, no longer raises `TypeError: unsupported operand type(s) for +=: 'NoneType' and 'float'`.
[...]

PR #1263, opened 2025-01-10
Set default value of evaluation_cost to 0 in BaseMetric class

rawheel:
Addresses #1262 The evaluate() method of GEval derived from BaseMetric was raising a TypeError: unsupported operand type(s) for +=: 'NoneType' and 'int/float' when called directly (without a prior call to measure()). This was due to the evaluation_cost attribute being initialized to None in the BaseMetric class.

```self.evaluation_cost += cost``` This throws an error as evaluation_cost is None by default in BaseMetric class and cost is a float.

This PR initializes evaluation_cost to 0 in the BaseMetric attributes, preventing the TypeError and allowing evaluate() to be called directly without issues.

### DIFF
--- a/deepeval/metrics/base_metric.py
+++ b/deepeval/metrics/base_metric.py
@@ -16,7 +16,7 @@ class BaseMetric:
     verbose_mode: bool = True
     include_reason: bool = False
     error: Optional[str] = None
-    evaluation_cost: Optional[float] = None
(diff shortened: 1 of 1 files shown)


penguine-ip:
Hey @rawheel i'm going to go ahead and close this one since we do want to keep a None and 0 distinction, thanks!

PR #1274, opened 2025-01-16
fix batched evaluate() behavior 

BjarniHaukur:
In the batched .evaluate() case, the metrics gathered do not match the columns of the pd.DataFrame created


penguine-ip:
@BjarniHaukur thanks!

PR #1287, opened 2025-01-19
OpenAI model updates

luarss:
Building on my previous PR, just making sure both `gpt_model*.py` files are updated to latest values.

- make sure gpt_model and gpt_model_schematic share same set of models
- update valid multimodal models
- update various costs


penguine-ip:
@luarss thanks so much, we should work more together :)

PR #1300, opened 2025-01-22
Exclude tests folder during installation

fj11:
After installed deepeval, I found that the tests folder also be installed under site-packages folder. Which cause the conflict if the original project already has a tests folder, so in this PR, I exclude the tests folder.


penguine-ip:
Hey @fj11 the test folder as in the unit tests in our repo right?

fj11:
> Hey @fj11 the test folder as in the unit tests in our repo right?

Yes, the unit tests folder

PR #1313, opened 2025-01-25
Update benchmarks-MMLU.mdx with MMLUTask

AMindToThink:
This fixes one of the problems with the [introduction to MMLU](https://docs.confident-ai.com/docs/benchmarks-mmlu).

There remain others:

The steps described result in the model generating long outputs, not single answers.
Also, batching doesn't work because MMLU has prompts of different lengths.


penguine-ip:
Hey @AMindToThink thanks appreicate it

PR #1316, opened 2025-01-25
Context precision docfix

### LINKED ISSUE #1212: Documentation for ContextualPrecision
The [documentation ](https://docs.confident-ai.com/docs/metrics-contextual-precision) states that contextual precision uses the `actual_output` which is not true as it can be seen [here](https://github.com/confident-ai/deepeval/blob/dd11c22b93d008ce0e386c88c86c96a38ad7c4f4/deepeval/metrics/contextual_precision/contextual_precision.py#L66-L74).

luarss:
Fixes #1212


penguine-ip:
Hey @luarss not going to merge this because all metrics require input and actual output regardless if its used for evaluation or not

Issue #954, opened 2024-08-14
Custom LLM Example do not Work


Otavio-Parraga:
**Describe the bug**
After setting the code, I always get an error with the following description: 

**AttributeError: 'list' object has no attribute 'find'**

I tried to run the code on my local server, with some modifications, and thus on Google Colab, just copying and pasting the website example (https://docs.confident-ai.com/docs/guides-using-custom-llms#creating-a-custom-llm)

Running it on server leads to the same error, and with other metrics (such as SummarizationMetric) I get the same error 
`AttributeError: 'list' object has no attribute 'find'` but replacing _find_ by _claim_

**To Reproduce**
Google Colab with code: https://colab.research.google.com/drive/1JbzWggqaxMKakQSzuVSYXN1J-96ttjOk?usp=sharing

**Expected behavior**
The lib should evaluate the metric.

**Screenshots**
![Captura de Tela 2024-08-14 às 16 58 46](https://github.com/user-attachments/assets/8ef4b0c2-4336-4ac7-849a-ddf845bd1bfa)

**Additional context**
The full error stack trace:
```

Event loop is already running. Applying nest_asyncio patch to allow async execution...
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
[/usr/local/lib/python3.10/dist-packages/deepeval/metrics/answer_relevancy/answer_relevancy.py](https://localhost:8080/#) in _a_generate_statements(self, actual_output)
    231             try:
--> 232                 res: Statements = await self.model.a_generate(
    233                     prompt, schema=Statements

TypeError: CustomLlama3_8B.a_generate() got an unexpected keyword argument 'schema'

During handling of the above exception, another exception occurred:

AttributeError                            Traceback (most recent call last)
7 frames
[<ipython-input-7-453899681174>](https://localhost:8080/#) in <cell line: 8>()
      6 
      7 metric = AnswerRelevancyMetric(model=custom_llm)
----> 8 metric.measure(test_case)

[/usr/local/lib/python3.10/dist-packages/deepeval/metrics/answer_relevancy/answer_relevancy.py](https://localhost:8080/#) in measure(self, test_case)
     50             if self.async_mode:
     51                 loop = get_or_create_event_loop()
---> 52                 loop.run_until_complete(
     53                     self.a_measure(test_case, _show_indicator=False)
     54                 )

[...]

PR #1347, opened 2025-02-06
fix: "Use a Custom LLM" guide

kaikaidai:
_Original issue flagged by @Otavio-Parraga [Custom LLM Example do not Work](https://github.com/confident-ai/deepeval/issues/954)_

Hi! When running the current example code for using Custom LLMs, I also got `AttributeError: 'list' object has no attribute 'find'`. 

Managed to find an implementation that would work & wanted to update the docs so others could get started too. 

- I found the JSON Confinement code to be necessary rather than optional. Therefore the relevant libraries for this are installed from the beginning. I also tried adjusting the language & structure of the rest of the guide to reflect this so it isn't confusing for others - but pls feel free to change anything back!

- `model_json_schema()` replaces `schema()` as the latter has been deprecated

- Added the [Selene Mini](https://huggingface.co/AtlaAI/Selene-1-Mini-Llama-3.1-8B) model in as an open source example 🤗

### DIFF
--- a/docs/guides/guides-using-custom-llms.mdx
+++ b/docs/guides/guides-using-custom-llms.mdx
@@ -4,19 +4,48 @@ title: Using Custom LLMs for Evaluation
 sidebar_label: Using Custom LLMs for Evaluation
 ---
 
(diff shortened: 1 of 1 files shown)


penguine-ip:
Thanks!

PR #1359, opened 2025-02-12
Wouldn't it be possible to allow the use of a custom template class?

giovcandido:
This way, we could adapt the instruction as we want. We could even provide translated versions of the instructions.


penguine-ip:
hey @giovcandido this is a pretty big shake up to the framework, can you show me what you're trying to achieve (like the test case you're trying to evaluate) so i can make a better judgement on whether this is the best way forward for deepeval?

PR #1379, opened 2025-02-23
Relax grpcio version constraint "1.67.1" -> "^1.67.1"

reasonmethis:
Changed grpcio dependency from an exact version (`1.67.1`) to a caret constraint (`^1.67.1`) in `pyproject.toml`. This prevents errors such as the following, when using `pip install -U deepeval` in a project that already uses grpcio:

```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
grpcio-status 1.68.1 requires grpcio>=1.68.1, but you have grpcio 1.67.1 which is incompatible.
```


reasonmethis:
Note: The tests are failing with the error:

```
pyproject.toml changed significantly since poetry.lock was last generated. Run `poetry lock` to fix the lock file.
```

I assume that the core maintainers would like to have full visibility and control over the diffs and would therefore prefer to update the poetry lock file themselves rather than have the change be included by an external contributor as part of the PR.

praveenkannan (approved):
Looks good to me!

penguine-ip:
hey @reasonmethis not at all this looks great, will merge it now, test locally tmr, and cut a new release for it! thanks.

PR #1406, opened 2025-03-02
Enable the API cost tracking in Synthesizer

chuqingG:
Add a new parameter (False by default) to configure and enable complete cost tracking. An use case is shown as following:
```python
synthesizer = Synthesizer(model="gpt-4o-mini", cost_tracking=True)
goldens = synthesizer.generate_goldens_from_docs(document_paths=doc_paths)
dataset = EvaluationDataset(goldens=goldens)
```
The output looks like:
image


penguine-ip:
@chuqingG this looks great, and you reformatted too! will go over it in the next day and merge it asap, thanks!

penguine-ip:
@chuqingG one question what is the reset cost for? i think the cumulative cost will scare a lot of people

chuqingG:
> @chuqingG one question what is the reset cost for? i think the cumulative cost will scare a lot of people

@penguine-ip I added reset_cost because technically generate_goldens_from_context() can be called either directly from outside or by generate_goldens_from_docs(). My intention is that the API cost only shows the overhead of one end-to-end synthesis call.

In other words, the API cost is accumulated when the generate_goldens_from_context is called from generate_goldens_from_docs, but it won't print out inside of generate_goldens_from_context. At the end of generate_goldens_from_docs, it will print the accumulated cost (generate_contexts, generate_goldens_from_context, and quality check).

Thanks!

penguine-ip:
@chuqingG thanks for the detailed explanation, definitely ok and love the money bag emoji! Thanks :)

penguine-ip:
@chuqingG for this "I also modified the quality check part to let small cheap models work smoothly in my personal branch."

What was it you had to modify? Any chance it can be used for deepeval?

PR #1414, opened 2025-03-05
Update big_bench_hard.py

skirdey-inflection:
res.answer might not exist if model doesn't return it correctly structured output, AttributeError could catch it


penguine-ip:
@skirdey-inflection will catching it help with the benchmark execution?

skirdey-inflection:
> @skirdey-inflection will catching it help with the benchmark execution?

it should catch scenarios when model answers {} without 'answer' key. 

If model times out, etc this exception handling will need to add these specific exectpions of the requests

penguine-ip:
@skirdey-inflection thanks!

PR #1423, opened 2025-03-09
fix(squad): use StringSchema instead of MultipleChoiceSchemaLower

### LINKED ISSUE #1422: Fix SQuAD Extractive Answer Prediction by Using StringSchema Instead of MultipleChoiceSchemaLower
### Bug Description
When running the SQuAD benchmark, the model’s prediction is consistently returning `"a"` instead of the expected extractive answer (e.g., `"solar"`, `"Rankine"`, etc.). This happens because the benchmark uses the `MultipleChoiceSchemaLower`, which is meant for multiple-choice questions—instead of a schema that expects a full text answer directly extracted from the context.

### To Reproduce
1. Configure the SQuAD benchmark with a task (e.g., Steam Engine) and run it:
```python
benchmark = SQuAD(
    tasks=[SQuADTask.STEAM_ENGINE],
    n_shots=5,
    evaluation_model="o3-mini",
    n_problems_per_task=5,
)
benchmark.evaluate(model=custom_llm)
```

2. Observe that for every example, the prediction returned is `"a"` and the score is 0.

### Proposed Solution  
Replace every usage of `MultipleChoiceSchemaLower` with `StringSchema` in the SQuAD benchmark.

1. **Update the Import Statement:**  
   Change the import from:
   ```python
   from deepeval.benchmarks.schema import MultipleChoiceSchemaLower
   ```
   to:
   ```python
   from deepeval.benchmarks.schema import StringSchema
   ```

2. **Update All Usages:**  
   Ensure that every part of the SQuAD benchmark class that currently uses `MultipleChoiceSchemaLower` is modified to use `StringSchema`. This change ensures that the model's output is processed as a full text extractive answer, rather than as a multiple-choice option.

carvalho28:
Switch to StringSchema in the SQuAD benchmark to correctly process extractive answers.

Fixes #1422 .

### DIFF
--- a/deepeval/benchmarks/squad/squad.py
+++ b/deepeval/benchmarks/squad/squad.py
@@ -8,7 +8,7 @@ from deepeval.benchmarks.base_benchmark import DeepEvalBaseBenchmark
 from deepeval.models import DeepEvalBaseLLM
 from deepeval.benchmarks.squad.task import SQuADTask
 from deepeval.benchmarks.squad.template import SQuADTemplate
-from deepeval.benchmarks.schema import MultipleChoiceSchemaLower
+from deepeval.benchmarks.schema import StringSchema
 from deepeval.telemetry import capture_benchmark_run
 from deepeval.metrics.utils import initialize_model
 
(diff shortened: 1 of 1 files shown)


penguine-ip:
@carvalho28  thanks!

PR #1424, opened 2025-03-10
Fix azure embedding command

AmaliMatharaarachchi:
Fix azure embedding command.
Error:
No such option: --embedding_deployment-name Did you mean --embedding-deployment-name?


penguine-ip:
@AmaliMatharaarachchi thanks!

PR #1425, opened 2025-03-10
update g_eval.py

kritinv:


penguine-ip:
@kritinv please make into helper function

PR #1434, opened 2025-03-15
fix: DAG NonBinary template

shrimpnoodles:
fixed DAG NonBinaryJudgement template for generating verdict--JSON schema was incomplete. Minor modifications node templates for format unison.


penguine-ip:
@shrimpnoodles thanks!

Issue #1436, opened 2025-03-15
prettify_list() Converts Non-ASCII Characters to Unicode Escape Sequences


milk333445:
**❗BEFORE YOU BEGIN❗**  
Are you on discord? 🤗 We'd love to have you asking questions on discord instead: https://discord.com/invite/a3K9c8GRGt  

### **Describe the bug**  
When calling `prettify_list()` on a list containing Pydantic models with non-ASCII characters (such as Chinese), the function returns a JSON string with Unicode escape sequences instead of properly displaying the characters.  

This happens because `json.dumps(jsonObj, indent=4)` defaults to `ensure_ascii=True`, which converts non-ASCII characters into Unicode escape sequences (e.g., `"你好"` becomes `"\u4f60\u597d"`).  

### **To Reproduce**  
1. Create a Pydantic model with a field containing non-ASCII characters (e.g., Chinese).  
2. Call `prettify_list()` on a list containing an instance of the model.  
3. The output will display Unicode escape sequences instead of the actual characters.  

Example code to reproduce the issue:  
```python
from pydantic import BaseModel
from deepeval.utils import prettify_list

class ExampleModel(BaseModel):
    text: str

example_instance = ExampleModel(text="你好")
print(prettify_list([example_instance]))
```

### **Expected behavior**  
The function should output properly formatted JSON without Unicode escape sequences when dealing with non-ASCII characters.  

For example, instead of:  
```json
[
    {
        "text": "\u4f60\u597d"
    }
]
```
It should output:
```json
[
    {
        "text": "你好"
    }
]
```

### **Suggested Fix**  
Modify the function in `deepeval/utils.py` by adding `ensure_ascii=False` to `json.dumps`:
```python
formatted_elements.append(
    json.dumps(jsonObj, indent=4, ensure_ascii=False).replace("\n", "\n    ")
)
```
This ensures that non-ASCII characters are displayed correctly in the formatted JSON output.

### **Desktop:**  
- OS: Windows
- Python Version: 3.10.14
- deepeval: 2.5.5

penguine-ip:
@milk333445 i believe it is resolved :)

drinktee:
I think the ensure_ascii should be False

PR #1437, opened 2025-03-16
Added the ensure_ascii attribute to prettify_list() function in utils.py

A-Vamshi:
Added the ensure_ascii attribute to prettify_list() function in utils.py to solve the issue #1436


penguine-ip:
@A-Vamshi thank you!

Issue #1363, opened 2025-02-14
Verbose printing goes into infinite loop in notebook


penguine-ip:
Looking for some help, a long running tricky bug:

In colab notebooks, when you turn `verbose_mode` = `True` what happens is you get a recursion error. This isn't reproducible when running the methods to print verbose logs in isolation.

See the code here: https://colab.research.google.com/drive/10zfwulUlTOXpJibVAjUGy-cdybi2JsLx#scrollTo=PvIxWTuU1aI4

fetz236:
Looking into this. Will keep you updated.

jayzalani:
@penguine-ip this needs to be closed Ig

penguine-ip:
@jayzalani yup!

PR #1444, opened 2025-03-18
Fix: Fix the verbose printing going into an infinite loop in notebook

fetz236:
Seems to work over here:
https://colab.research.google.com/drive/1Hjt29I4VOAbtdR9I_SxZ8ySm9V3czYZY?usp=sharing

I have managed to remove the recursion error.

https://github.com/confident-ai/deepeval/issues/1363


fetz236:
I am kind of confused on the output required. I have checked and the underlying problem seems to be that Rich's console is having problems with IPython's display system. If you use sys instead `sys.stdout.write()` vs `print()` it should work as well but the output needs to be rewritten.

fetz236:
https://colab.research.google.com/drive/1LEvIzJdHKfSCjHpgRh0zsFDsWCJBk-0y#scrollTo=aALUlF9g0XIm


Think this is better looks like it works

PR #1459, opened 2025-03-20
Makes synthesizer work with Azure Open Ai

ntgussoni:
Hello guys, 

This more than a Pull Request is a space of discussion, since I want to be sure the changes don't just make it work for me but break for the other cases.

- The first fix is that `self.model.generate(prompt)` without `schema=Response` returns `str` and not `{ response:... }`
- The second one, and this perhaps more controversial.

that `res` is the following because the return of AzureOpenAi is a tuple of `Tuple[res, cost]`
``` python
(SyntheticDataList(data=[SyntheticData(input='.....')]), 0.0001128)
```

Therefore on line 727 it breaks. Since `AttributeError: 'tuple' object has no attribute 'data'`
```python
    synthetic_data_items = res.data
```

I'm not using any custom model, just Azure Open AI set in the config.
I don't understand why it's classified as `not_gpt_model`, but even if that's right then we need a special case for handling the output.

Is there anything im missing?

### DIFF
--- a/deepeval/synthesizer/synthesizer.py
+++ b/deepeval/synthesizer/synthesizer.py
@@ -915,7 +915,7 @@ class Synthesizer:
         else:
             try:
                 res = model.generate(prompt, schema=schema)
-                return res
+                return res[0]
             except TypeError:
                 res = model.generate(prompt)
                 data = trimAndLoadJson(res, self)
@@ -963,7 +963,7 @@ class Synthesizer:
             else:
                 res, cost = self.model.generate(prompt)
                 self.synthesis_cost += cost
-                return res.response
+                return res
         else:
             try:
                 res: Response = self.model.generate(prompt, schema=Response)
@@ -981,7 +981,7 @@ class Synthesizer:
             else:
                 res, cost = await self.model.a_generate(prompt)
                 self.synthesis_cost += cost
-                return res.response
+                return res
         else:
             try:
                 res: Response = await self.model.a_generate(
--- a/deepeval/synthesizer/synthesizer.py
+++ b/deepeval/synthesizer/synthesizer.py
@@ -13,7 +13,7 @@ import tqdm
 import csv
 import os
 
-from deepeval.models import GPTModel
+from deepeval.models import GPTModel, AzureOpenAIModel
 from deepeval.utils import get_or_create_event_loop, is_confident
 from deepeval.synthesizer.chunking.context_generator import ContextGenerator
 from deepeval.metrics.utils import (
--- a/deepeval/synthesizer/synthesizer.py
+++ b/deepeval/synthesizer/synthesizer.py
@@ -915,7 +915,7 @@ class Synthesizer:
         else:
             try:
                 res = model.generate(prompt, schema=schema)
-                if isinstance(model, AzureOpenAIModel)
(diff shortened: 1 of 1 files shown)


ntgussoni:
Everything Ive changed is because it wasn't working and I have in fact tested it.

I can check the documentation as well, however I'm not using a custom model, just azure open Ai, which is defined internally. Maybe I'm not following what you meant?

penguine-ip:
I think that's fine @ntgussoni , just to confirm returning `res` instead of `res.response` doesn't break anything right? Because i don't see an issue with the rest of the PR expect for this one minor detail that might cause breaking changes

ntgussoni:
Yeah, there's no response if there is no schema passed.

srkirkland:
I get the same error running the synthesizer w/ Ollama, I suspect for a similar reason.  Hopefully a fix can be found that works for all models.

Reproduction:

`deepeval set-ollama deepseek-r1:70b`

and then running the example code from https://docs.confident-ai.com/docs/synthesizer-generate-from-contexts#generate-your-goldens:

```
synthesizer = Synthesizer()
synthesizer.generate_goldens_from_contexts(
    # Provide a list of context for synthetic data generation
    contexts=[
        ["The Earth revolves around the Sun.",
            "Planets are celestial bodies."],
        ["Water freezes at 0 degrees Celsius.",
            "The chemical formula for water is H2O."],
    ]
)
```

results in `AttributeError: 'tuple' object has no attribute 'data'`.

Full error stack 
```
  File "<stdin>", line 1, in <module>
  File "/home/postit/.conda/envs/policywonk/lib/python3.12/site-packages/deepeval/synthesizer/synthesizer.py", line 275, in generate_goldens_from_contexts
    loop.run_until_complete(
  File "/home/postit/.conda/envs/policywonk/lib/python3.12/asyncio/base_events.py", line 686, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/home/postit/.conda/envs/policywonk/lib/python3.12/site-packages/deepeval/synthesizer/synthesizer.py", line 421, in a_generate_goldens_from_contexts
    await asyncio.gather(*tasks)
  File "/home/postit/.conda/envs/policywonk/lib/python3.12/site-packages/deepeval/synthesizer/synthesizer.py", line 1001, in task_wrapper
    return await func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/postit/.conda/envs/policywonk/lib/python3.12/site-packages/deepeval/synthesizer/synthesizer.py", line 446, in _a_generate_from_context
    synthetic_inputs: List[SyntheticData] = await self._a_generate_inputs(
                                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/postit/.conda/envs/policywonk/lib/python3.12/site-packages/deepeval/synthesizer/synthesizer.py", line 723, in _a_generate_inputs
    synthetic_data_items = res.data
                           ^^^^^^^^
AttributeError: 'tuple' object has no attribute 'data'
```

Versions:
`deepeval == 2.6.3`
`ollama version is 0.6.2`

penguine-ip:
Hey  @srkirkland can you open a new issue for this if this does not fix anything?

PR #1460, opened 2025-03-21
Fix: ollama base_url setting

paul91:
When running `deepeval set-ollama <model-name> --base-url=<host>` it saves the base url as `LOCAL_MODEL_BASE_URL` in your `.deepeval` config. It appears that the ollama model handler in Deepeval was mistakenly looking at `LOCAL_EMBEDDING_BASE_URL` instead so it would still attempt to call the default localhost base url.


penguine-ip:
@paul91 tysm! :)

PR #1464, opened 2025-03-24
Fix async return logic in measure method

RomaanMkv:
**Bug Description**  
Async execution path in Metrcs `measure` method fails to return the score due to misplaced return statement while using `async_mode=True`.

**To Reproduce**  
1. Create test case using `LLMTestCase` class:
```py
case = LLMTestCase(
        input=query,
        actual_output=actual_output,
        expected_output=expected_output,
        context=[expected_output],
        retrieval_context=retrieval_context,
)
```
2. Create Metric, for example `ContextualPrecisionMetric` and use `async_mode=True`:
```py
metric = ContextualPrecisionMetric(
            threshold=threshold,
            model=model,
            async_mode=True,
)
```
3. Run metric calculation:
```py
result = metric.measure(case)
```
4. Observe result. Its value will be 'none', since incorrect placement of return statement in 'measure'

**Expected behavior**  
Both async and sync execution paths should return non `none` value -  `self.score`.

### DIFF
--- a/deepeval/metrics/answer_relevancy/answer_relevancy.py
+++ b/deepeval/metrics/answer_relevancy/answer_relevancy.py
@@ -81,7 +81,7 @@ class AnswerRelevancyMetric(BaseMetric):
                     ],
                 )
 
-                return self.score
+            return self.score
 
     async def a_measure(
         self,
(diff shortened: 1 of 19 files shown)


penguine-ip:
@RomaanMkv thanks so much for changing it for all metrics!

Issue #1473, opened 2025-03-27
OllamaEmbeddingModel returns 404 when attempting to embed texts


paul91:
It appears that the `OllamaEmbeddingModel` is using the `OpenAI` client to communicate with an ollama instance. This will end up throwing 404s as the OpenAI URI structure is different than ollama's.

When `a_embed_text` is called, a POST gets sent to `<base_url>/embeddings`, while ollama expects these at `<base_url>/api/embed`.

https://github.com/confident-ai/deepeval/blob/1f47e88d6afeda5340dde6b0c385e42b428d2a51/deepeval/models/embedding_models/ollama_embedding_model.py#L24

Since deepeval is already using some langchain_x packages, we could probably just add `langchain_ollama` and use `OllamaEmbeddings` in load_model.

paul91:
It appears that Ollama does support the OpenAI request format, you just have to set your `base_url` to include `/v1` at the end so requests go to `<ollama-host>/v1/embedding` instead. This is a bit confusing as `OllamaModel` uses the `ollama` base library which works with a bare base_url.

paul91:
Resolved by #1474

PR #1474, opened 2025-03-27
Make OllamaEmbeddingModel consistent with OllamaModel

paul91:
This is in reference to #1473, where I am attempting to ensure that `OllamaEmbeddingModel` is consistent with `OllamaModel` in that they both use the same underlying `ollama` module.

This addresses having to set a different base_url's between the two when using the same ollama host. OpenAI seems to need `<ollama_host>/v1` so its calls to `/embeddings` lands on `/v1/embeddings` where the OpenAI compatible endpoint actually is https://github.com/ollama/ollama/blob/main/docs/openai.md#openai-python-library. Since OllamaModel uses `ollama` it expects the base_url to not include `/v1` and it adds it automatically.

Using the same module let's us use the same `base_url` and makes this a bit less confusing.

It's fine if you'd rather handle this some other way, I figured I would at least open this as a result of my own debugging.

### DIFF
--- a/deepeval/cli/main.py
+++ b/deepeval/cli/main.py
@@ -236,7 +236,7 @@ def set_ollama_embeddings_env(
         ..., help="Name of the Ollama embedding model"
     ),
     base_url: str = typer.Option(
-        "http://localhost:11434/v1/",
+        "http://localhost:11434",
         "-b",
         "--base-url",
         help="Base URL for the Ollama embedding model API",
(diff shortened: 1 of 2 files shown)


kritinv:
Hey @paul91. This makes sense will review and merge asap.

kritinv:
@penguine-ip I think this makes sense to merge now. The CLI has been changed accordingly (no need to update docs since URL is handled internally).

penguine-ip:
@paul91 @kritinv thanks!

PR #1481, opened 2025-03-30
Features/convo simulator

penguine-ip:


penguine-ip:
@tsenart that's great! i'm sure our implementation needs a lot of improvement, would love to see how you built it if possible ^ :)

tsenart:
Can share a redacted version without the actual prompts, here it goes: https://gist.github.com/tsenart/25f4e1c6a88ad6661da50f339b2ac2da

Two things that I hadn't solved yet:

1. I couldn't figure out how to make the voice agent have the first message in the history. That's how our actual agent works.
2. The agents just keep blabbering with each other saying good bye until max turns is reached. Haven't researched the best way to make the conversation end more naturally. Would tool calling be the best approach? Give the agents a end conversation tool that they can call when they feel the conversation has reached an end (or otherwise!)

PR #1484, opened 2025-04-01
draft structure

penguine-ip:


penguine-ip:
@realei here is the PR, will talk more on discord

PR #1485, opened 2025-04-01
Fix Synthetic Dataset Generation Encoding Issues (UnicodeDecodeError)

AahilShaikh:
Ran into a UnicodeDecodeError when generating synthetic datasets from documents, specifically this:
```
UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d in position 232: character maps to <undefined>
```
Turns out TextLoader was defaulting to Windows encoding instead of UTF-8. Found the issue in doc_chunker.py  https://github.com/confident-ai/deepeval/blob/852363ef5dde6e4e27c48e234ba6527a6479a15b/deepeval/synthesizer/chunking/doc_chunker.py#L246

Implemented a fix to:

 - Default to auto-detecting encoding, which should handle most cases.
 - Allow manual encoding configuration for those weird edge cases.

Let me know if you guys have any thoughts or prefer a different approach!


kritinv:
Hey @AahilShaikh. Thanks for making the PR. I'm wondering if this is finished, and whether `from humanfriendly.terminal import auto_encode` is necessary?

AahilShaikh:
Ah good catch @kritinv, Pycharm's autocomplete is a little too eager sometimes 😅. Other than that, this PR should be good to go, unless there's something I'm missing, in which case, feel free to let me know.

penguine-ip:
@AahilShaikh @kritinv  looks great, thanks!

PR #1490, opened 2025-04-02
fix: Correction of type hints

konerzajakub:
The **context_quality_threshold** and **context_similarity_threshold** fields were originally typed as integers, but their default values were floats.


penguine-ip:
@konerzajakub thanks!

PR #1493, opened 2025-04-03
Gemini Native Model Support

kritinv:


deepeval/models/mlllms/gemini_model.py
```diff
@@ -0,0 +1,288 @@
...
+from pydantic import BaseModel
+from google.genai import types
+from google import genai
+
+from deepeval.key_handler import KeyValues, KEY_FILE_HANDLER
+from deepeval.models.base_model import DeepEvalBaseMLLM
+from deepeval.test_case import MLLMImage
+
+
+valid_multimodal_gemini_models = [
```
meteatamel, line 12: I know this is already merged but isn't having a list of valid models a little brittle? There are newer models already and now we can't use them in evaluations due to this hard coded list

rarsan: I agree. I originally added it for consistency and to prevent a user from incorrectly setting text-only model for a multimodal LLM eval use case. However, with single-modality Gemini 1.0 pro being discontinued this month, and all current Gemini 2.0+ being multimodal...I think this allow list is no longer necessary.

meteatamel: I opened a PR to remove these checks: https://github.com/confident-ai/deepeval/pull/1503

Issue #1502, opened 2025-04-10
Gemini model classes should not hardcode valid models


meteatamel:
Currently, the `gemini_model.py` in `llms` and `mlllms` folders hardcode a list of valid model names and use that list for error checking. This is good for error checking but it's very brittle as new models come up almost every month and old models get deprecated. The list is already out of date as of today.

My suggestion is to not hard code this list at all. If a user enters the wrong or out-of-date model, the request will fail anyway. It might be harder to debug for the user why the request failed but keeping this list up to date and relying on a new release of DeepEval for every model update is not going to work either.

PR #1503, opened 2025-04-10
Fixes: Gemini model classes should not hardcode valid models #1502

meteatamel:


penguine-ip:
@meteatamel why?

meteatamel:
@penguine-ip I explained in the issue #1502

penguine-ip:
@meteatamel i see that makes a lot of sense, i've merged it, thanks!

PR #1516, opened 2025-04-17
Remove async from `get_model_name` of `DeepEvalBaseEmbeddingModel`

ramipellumbi:
As defined, the model name prints as a coroutine object when generating goldens with a custom embedding model. I didn't see a reason for this to be an `async` function in the code


penguine-ip:
@ramipellumbi thanks!

PR #1531, opened 2025-04-23
Improve string representation `LLMTestCase` in `g_eval.py`

joaopmatias:
* Correct different `LLMTestCase` string representations in `evaluate` and `_a_evaluate` methods.
* Extract the corresponding transformation to a new method.


penguine-ip:
@joaopmatias great thanks!

PR #1543, opened 2025-04-28
Add custom file handlers

ramipellumbi:
As discussed in [discord](https://discord.com/channels/1118193057620316202/1198017217690877953/threads/1362849910625538269)
* Add ability to add custom file loaders when constructing from documents
* Organizes imports in `synthesizer` library and removes all unused imports


deepeval/synthesizer/file_handler.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def __init__(
+        self, handler: FileHandler, path: str, encoding: Optional[str] = None
+    ):
+        self.handler = handler
+        self.path = path
+        self.encoding = encoding
+
+    def lazy_load(self) -> Iterator[LCDocument]:
+        return self.handler.load(self.path, self.encoding)
```
ramipellumbi, line 37: this is the only method recommended to be overridden in `BaseLoader`

penguine-ip:
@ramipellumbi hey thanks for the PR! just had time to look at this. It looks pretty nice generally, the only thing is we're trying to avoid making any langchain stuff a required dependency so would like to make a late import for it, also would it be possible to have a few lines of docs for this new feature in this page? https://www.deepeval.com/docs/synthesizer-generate-from-contexts

thanks!

PR #1551, opened 2025-05-02
llm attributes improved

spike-spiegel-21:
LLM attribute updated


spike-spiegel-21:
closing this for now, coming up with some better approach

PR #1563, opened 2025-05-07
fix: fix error when use custom llm adapter

ahmad-einieh:


penguine-ip:
Hey @ahmad-einieh can you explain what was the error?

ahmad-einieh:
Hey @penguine-ip,

I'm running into an issue when trying to create a custom model based on `DeepEvalBaseLLM`. The problem stems from this line in the base class:

```python
self.evaluation_cost = 0 if self.using_native_model else None
```

Since `DeepEvalBaseLLM` doesn't provide a way to configure this behavior, it prevents me from using a custom model the way I need to for my project. This customization is important to my workflow.

Here's the traceback I get when running my code:

```
  File ".../llm_as_judge.py", line 581, in <module>
    batch_results_obj = multi_judge.evaluate_batch(test_cases_obj)
  File ".../llm_as_judge.py", line 361, in evaluate_batch
    result = self._evaluate_test_case(test_case, test_id, metadata)
  File ".../llm_as_judge.py", line 235, in _evaluate_test_case
    metric.measure(test_case)
  File ".../site-packages/deepeval/metrics/g_eval/g_eval.py", line 131, in measure
    loop.run_until_complete(
  File ".../lib/python3.12/asyncio/base_events.py", line 686, in run_until_complete
    return future.result()
  File ".../site-packages/deepeval/metrics/g_eval/g_eval.py", line 175, in a_measure
    g_score, reason = await self._a_evaluate(test_case)
  File ".../site-packages/deepeval/metrics/g_eval/g_eval.py", line 280, in _a_evaluate
    self.evaluation_cost += cost
TypeError: unsupported operand type(s) for +=: 'NoneType' and 'float'
```

PR #1601, opened 2025-05-15
add write utils

daehuikim:
When I use deepeval with massive data without web application, It is difficult to track every results from the printed logs.
To address this, I hope to add file writing functions on the utils to get log  for evaluation.
This function even helps to track missing data by connection errors.

### DIFF
--- a/deepeval/evaluate/configs.py
+++ b/deepeval/evaluate/configs.py
@@ -23,6 +23,7 @@ class DisplayConfig:
     print_results: bool = True
     verbose_mode: Optional[bool] = None
     display_option: Optional[TestRunResultDisplay] = TestRunResultDisplay.ALL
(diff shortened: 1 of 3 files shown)


daehuikim:
@penguine-ip Yes please. :)

penguine-ip:
Hey @daehuikim thanks for changing it, i was hoping to have everything in utils. An optional `file_output_dir` would be great for the display config, what do you think?

daehuikim:
@penguine-ip Can you confirm the change from the last commit please? I hope I am on the right way. 
I just made ```file_output_dir``` in the ```DisplayConfig``` and everything is happen in the utils.py

penguine-ip:
@daehuikim this is perfect, thanks!

Issue #1586, opened 2025-05-12
`deepeval --help` gives error, typer incompatibility with click v8.2.0 + Typer 0.15.3


vipulgupta2048:
**Describe the bug**
I was following the Getting Started. Running `deepeval --help` after a fresh installation in a virtual environment results in this error: 

```
(confident) ➜  deepeval git:(main) ✗ deepeval --help
 Usage: deepeval [OPTIONS] COMMAND [ARGS]...                                              
╭────────────────────────── Traceback (most recent call last) ───────────────────────────╮
│ /Users/vipulgupta2048/.virtualenvs/confident/bin/deepeval:8 in <module>                │
│                                                                                        │
│   5 from deepeval.cli.main import app                                                  │
│   6 if __name__ == '__main__':                                                         │
│   7 │   sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])               │
│ ❱ 8 │   sys.exit(app())                                                                │
│   9                                                                                    │
│                                                                                        │
│ ╭────────────────────────────────────── locals ──────────────────────────────────────╮ │
│ │ app = <typer.main.Typer object at 0x114482e90>                                     │ │
│ │  re = <module 're' from                                                            │ │
│ │       '/usr/local/Cellar/python@3.13/3.13.2/Frameworks/Python.framework/Versions/… │ │
│ │ sys = <module 'sys' (built-in)>                                                    │ │
│ 
TypeError: Parameter.make_metavar() missing 1 required positional argument: 'ctx'
```

From looking around and LLMs, it seems (TypeError: Parameter.make_metavar() missing 1 required positional argument: 'ctx') is a known issue that can occur with certain combinations of Typer, Click, and Python versions, especially with very new Python releases.

I tried downgrading click to 8.1.7 which is what Typer supports as the latest stable with `pip install 'click==8.1.7'`

After this the command works as expected

```
(confident) ➜  deepeval git:(main) ✗ deepeval --help
                                                                                          
 Usage: deepeval [OPTIONS] COMMAND [ARGS]...                                              
                                                                                          
╭─ Options 
```

**Desktop (please complete the following information):**
 - OS: MacOS 14.7.4 (23H420) running Python 3.13.2

Would highly recommend checking in if recommending `pip install -U deepeval` as recommended in the docs is a good option. While following the guide, I thought it could lead to instability as dependencies would auto-update causing incompatibilities.

ig-raiderler:
[...]

PR #1607, opened 2025-05-19
Restricted click and typer for compatibility

A-Vamshi:
Fixed the compatibility issue for typer and click from issue #1586


penguine-ip:
@A-Vamshi thanks!

PR #1609, opened 2025-05-20
Added @Sam120204's fix for issue 1526. Tested only on Windows 10.

Propet40:
Fixes the ChromaDB issue which prevents usage of the Synthesizer on Windows machines.


Propet40:
I only tested it with a Windows machine as that is what I have access to.

penguine-ip:
@Propet40 hey thanks so much!

PR #1619, opened 2025-05-23
Fix ValueError for TruthfulQA

mdsalnikov:
ValueError: 5 columns passed, passed data had 4 columns in line 107


penguine-ip:
@mdsalnikov thanks!

Issue #1612, opened 2025-05-21
GSM8K: `AttributeError: 'tuple' object has no attribute 'answer'`


jamesbraza:
**Describe the bug**

It seems impossible to avoid `AttributeError: 'tuple' object has no attribute 'answer'` with GSM8K.

**To Reproduce**

Running the code from https://github.com/confident-ai/deepeval/blob/v2.9.1/docs/docs/benchmarks-gsm8k.mdx#usage

```python
from deepeval.benchmarks import GSM8K
from deepeval.models import GPTModel

benchmark = GSM8K(n_problems=10, n_shots=3, enable_cot=True)

benchmark.evaluate(model=GPTModel())
print(benchmark.overall_score)
```

I get:

```none
...
  File "/path/to/repo/scripts/deepeval_play.py", line 11, in <module>
    benchmark.evaluate(model=GPTModel())
  File "/path/to/repo/.venv/lib/python3.12/site-packages/deepeval/benchmarks/gsm8k/gsm8k.py", line 53, in evaluate
    prediction, score = self.predict(model, golden).values()
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/repo/.venv/lib/python3.12/site-packages/deepeval/benchmarks/gsm8k/gsm8k.py", line 99, in predict
    prediction = str(res.answer)
                     ^^^^^^^^^^
AttributeError: 'tuple' object has no attribute 'answer'
```

**Expected behavior**

No `AttributeError` crash when running this code.

**Additional context**

A similar same error was hit in https://github.com/confident-ai/deepeval/issues/1535.

DeepEval should use `mypy`.

A-Vamshi:
Is this issue fixed?

Issue #1535, opened 2025-04-27
AttributeError: 'tuple' object has no attribute 'answer'


kunlun122:
**Describe the bug**
When I tried to evaluate the Ollama model using the MMLU (Massive Multitask Language Understanding) from deepeval, the following error occurred:**AttributeError: 'tuple' object has no attribute 'answer'**

`from deepeval.models import OllamaModel
from deepeval.metrics import AnswerRelevancyMetric
from deepeval.benchmarks import MMLU
from deepeval.benchmarks.mmlu.task import MMLUTask

model = OllamaModel(
    model="qwen2.5:7b",
    base_url="http://localhost:11434"
)
answer_relevancy = AnswerRelevancyMetric(model=model)
print(model.generate("你好"))

benchmark = MMLU(
    tasks=[MMLUTask.HIGH_SCHOOL_COMPUTER_SCIENCE, MMLUTask.ASTRONOMY],
    n_shots=3
)

benchmark.evaluate(model=model)
print(benchmark.overall_score)`
**Screenshots**

![Image](https://github.com/user-attachments/assets/5f286536-e1ce-4c8e-b886-42f5e16d2d6e)

![Image](https://github.com/user-attachments/assets/27f98c70-60d3-44ef-bb37-1c4dd7869907)

**Desktop (please complete the following information):**
 - OS: ubuntu 24.04
 - deepeval 2.7.6, install by pip

krishna0125:
Hi @kunlun122 , Please find the fix in the PR. you can clone the repo and edit the code. This will help you to run the code.

PR #1628, opened 2025-05-27
Fix/gsm8k tuple response attribute error

SYED-M-HUSSAIN:
### 📍 Summary

This PR resolves [(https://github.com/confident-ai/deepeval/issues/1612)](https://github.com/confident-ai/deepeval/issues/1612) by adding robust handling for non-standard model responses (e.g., tuples) that previously caused an `AttributeError` in the `GSM8K` benchmark.

### 🐞 Problem

The original implementation assumed that model outputs would always return an object with an `.answer` attribute (e.g., `NumberSchema`). However, in some cases, the model returns a tuple, causing the following error:

```
AttributeError: 'tuple' object has no attribute 'answer'
```

This occurred specifically at:

```python
prediction = str(res.answer)
```

and earlier:

```python
prediction, score = self.predict(model, golden).values()
```

### ✅ Solution

This PR includes a targeted fix that preserves backward compatibility:

* **Replaced unsafe `.values()` unpacking** with direct dictionary key access for safer and clearer code.
* **Introduced a helper method** `_extract_prediction_from_response()` to robustly extract the prediction from various formats:

  * `NumberSchema` objects
  * Tuples
  * Raw strings, dictionaries, or objects with `.text` or `.content` attributes
* **Improved exception handling** by catching both `AttributeError` and `TypeError` in case of unexpected structures.

#### In `gsm8k.py`:

* Replaced:

  ```python
  prediction, score = self.predict(model, golden).values()
  ```

  With:

  ```python
  result = self.predict(model, golden)
  prediction = result["prediction"]
  score = result["score"]
  ```

* Modified:

  ```python
  prediction = str(res.answer)
  ```

  To:

  ```python
  prediction = self._extract_prediction_from_response(res)
  ```

* Added a new method `_extract_prediction_from_response()` for robust prediction extraction.

### 🛡️ Impact

* ✅ Fixes the reported crash in GSM8K benchmark (#1612)
* ✅ Makes prediction extraction robust to unexpected model return formats
* ✅ Preserves existing functionality and backward compatibility

---

### 🧾 Related

* Issue: #1612
* Related discussion: #1535

---

### DIFF
--- a/deepeval/benchmarks/gsm8k/gsm8k.py
+++ b/deepeval/benchmarks/gsm8k/gsm8k.py
@@ -1,4 +1,4 @@
-from typing import List, Optional, Dict
+from typing import List, Optional, Dict, Union
 from tqdm import tqdm
 
 from deepeval.dataset import Golden
@@ -52,7 +52,10 @@ class GSM8K(DeepEvalBaseBenchmark):
             for idx, golden in enumerate(
                 tqdm(goldens, desc=f"Processing {self.n_problems} problems")
             ):
-                prediction, score = self.predict(model, golden).values()
+                result = self.predict(model, golden)
+                prediction = result["prediction"]
+                score = result["score"]
+                
                 if score:
                     overall_correct_predictions += 1
                 predictions_row.append(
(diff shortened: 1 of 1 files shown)


penguine-ip:
@SYED-M-HUSSAIN yes please lets simplify it a bit! we can always fix it further if someone raises an issue again :)

SYED-M-HUSSAIN:
Thanks @penguine-ip, I’ve refactored the method to simplify the if-else chain

penguine-ip:
@SYED-M-HUSSAIN thanks!

Issue #1629, opened 2025-05-27
TypeError: unsupported operand type(s) for /: 'list' and 'int'


deepeshAg:
A TypeError occurs when running the code, specifically:
TypeError: unsupported operand type(s) for /: 'list' and 'int'
This happens in image_coherence.py at the line score = score / 10. The variable score is a list, but the code tries to divide it by an integer, which is not allowed in Python.

Run the script (e.g., python3 main.py) that uses the ImageCoherenceMetric from the deepeval package.

Expected behavior
The code should perform a numeric division on a single number (int or float), not on a list. If score is a list of numbers, the code should aggregate it (e.g., take the average) before dividing.

**Screenshots**

![Image](https://github.com/user-attachments/assets/b3599cf1-0335-4300-afaf-2843fea07bcb)

OS: macOS (darwin 24.5.0)
Browser: N/A
Version: Python 3.10.10

penguine-ip:
hey @deepeshAg and @A-Vamshi i believe its fixed in 3.0.1 (https://github.com/confident-ai/deepeval/commit/1b52189a36f55753259830f7cc7905040d9adc18#diff-cac10f0ffab564dfa6232eaf44643f86e3054ef8ba85187eea12261b2ddd07d6), thanks!

PR #1637, opened 2025-05-28
Conditional type safety for score 

A-Vamshi:
Fixed the TypeError from the issue #1629
Aggregated the score if it is an instance of list to average of the scores


penguine-ip:
@A-Vamshi actually i fixed it in here already: https://github.com/confident-ai/deepeval/commit/1b52189a36f55753259830f7cc7905040d9adc18#diff-cac10f0ffab564dfa6232eaf44643f86e3054ef8ba85187eea12261b2ddd07d6

the problem was the schema for some should be `score: float` instead of `score: List[float]` but there was a copy and paste error when building it initially :)

A-Vamshi:
@penguine-ip ah gotcha, sorry I didn't check the schema, thanks!

PR #1642, opened 2025-05-30
Improvements to AnswerRelevancy templates

Spectavi:
Ensures all examples are demarcated with "Example:" and "==== END OF EXAMPLE ====".
Adds clarifying language on splitting statements and ensuring valid JSON is output.


Spectavi:
Ping, these basic adjustments will increase the consistency of the Answer Relevancy metric, particularly when using smaller models like Llama 3.3 70b.  Without it I would sporadically get back invalid JSON causing noise in my eval results.

penguine-ip:
@Spectavi thanks!

PR #1654, opened 2025-06-04
Fix: Support read only file systems

### LINKED ISSUE #1577: Error: [Errno 30] Read-only file system: '.deepeval'
facing this error while trying to use the library in aws lambda.

lkacenja:
When we import telemetry.py, we do some file system preparation. Some folks run their evaluation code in environments that don't allow write access to the file system like Lambda's or their Google Cloud or Azure equivalents. This causes evaluations to fail when DeepEval is imported. DeepEval provides an environmental variable to opt out of telemetry. It would be great if this flag was checked, before fiddling with the file system.

This PR wraps the file system in a check for the environmental variable, "DEEPEVAL_TELEMETRY_OPT_OUT".

We tested this branch in our AWS Lambdas and locally and it seemed to work in our use cases. 

closes #1577

### DIFF
--- a/deepeval/telemetry.py
+++ b/deepeval/telemetry.py
@@ -27,30 +27,7 @@ TELEMETRY_DATA_FILE = ".deepeval_telemetry.txt"
 TELEMETRY_PATH = os.path.join(HIDDEN_DIR, TELEMETRY_DATA_FILE)
 
 #########################################################
-### Move Folders ########################################
-#########################################################
-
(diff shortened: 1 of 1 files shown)


penguine-ip:
Hey @lkacenja thanks for the PR, taking a look right now

penguine-ip:
hey @lkacenja do you think you can introduce another env variable likw `"DEEPEVAL_FILE_SYSTEM"` and make it default to READ_AND_WRITE but also READ_ONLY?

PR #1690, opened 2025-06-13
Update guides-llm-observability.mdx

YashDhirajOza:
Udapted major wrong contents and improved it by a lot


penguine-ip:
Hey @YashDhirajOza are you sure this is an improvement?

Issue #1697, opened 2025-06-16
evaluate function raise TypeError: 'TestResult' object is not iterable


ELvisZxc:
**❗BEFORE YOU BEGIN❗**
Are you on discord? 🤗 We'd love to have you asking questions on discord instead: https://discord.com/invite/a3K9c8GRGt

**Describe the bug**
`evaluate` function use `DisplayConfig` with `file_output_dir`, will raise TypeError: 'TestResult' object is not iterable.

the `aggregate_metric_pass_rates_to_file` function is called with a single TestResult object instead of a list of TestResult objects. on line 466 of in utils.py

**To Reproduce**
```
        evaluate(
            test_cases=test_cases,
            metrics=[
                AnswerRelevancyMetric(model=model),
                FaithfulnessMetric(model=model),
                ContextualPrecisionMetric(model=model),
                ContextualRecallMetric(model=model),
                ContextualRelevancyMetric(model=model)
            ],
            display_config=DisplayConfig(file_output_dir=f"{self.path}/benchmark_response"),
        )
```

**Expected behavior**
The evaluate function should correctly handle the test_cases parameter as a list of TestResult objects, ensuring that aggregate_metric_pass_rates_to_file can process the input correctly.

**Screenshots**

![Image](https://github.com/user-attachments/assets/6349bcd8-0805-4126-9afe-055e97db1182)

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

PR #1705, opened 2025-06-18
Fixing evaluate function raise TypeError: 'TestResult' object is not iterable

adityabharadwaj198:
This fixes [#1697](https://github.com/confident-ai/deepeval/issues/1697)


penguine-ip:
@adityabharadwaj198 thanks so much!

Issue #1707, opened 2025-06-18
implement "don't rewrite to file if it is read only" using DEEPEVAL_FILE_SYSTEM


adityabharadwaj198:
**❗BEFORE YOU BEGIN❗**
Are you on discord? 🤗 We'd love to have you asking questions on discord instead: https://discord.com/invite/a3K9c8GRGt

**Is your feature request related to a problem? Please describe.**
Including a DEEPEVAL_FILE_SYSTEM env variable, which a user can set to "READ_ONLY" or "READ_WRITE"
the idea is that we won't rewrite to file if it is read only

**Describe the solution you'd like**
DEEPEVAL_FILE_SYSTEM env check added in __init__.py https://github.com/adityabharadwaj198/deepeval/blob/aditya/fix-type-error/deepeval/__init__.py#L73
but it looks like it isn’t used yet across file-writing logic.

1. scan the codebase for all file write operations (open(..., "w"), etc.)
2. Wrap them with a check for is_read_only_env() so they no-op in READ_ONLY mode

**Describe alternatives you've considered**

**Additional context**

PR #1713, opened 2025-06-21
Implement don't re-write to a file if it's read only (addresses https://github.com/confident-ai/deepeval/issues/1707) 

adityabharadwaj198:
1. Moved the is_read_only_env() method from __init__.py to utils.py to avoid a circular dependency while importing 
2. Added is_read_only_env() check in front of critical code paths where we write to a file 
3. added some tests


penguine-ip:
@adityabharadwaj198 i think there's a lot more writing across deepeval, especially in test-run where there is caching

PR #1715, opened 2025-06-21
chore: add claude-opus-4 and claude-sonnet-4 models support

AbhishekRP2002:
fix: add error handling for unknown model in anthropic client  in calculate_cost

### Summary

This PR updates the `calculate_cost` method in the `AnthropicModel` class to raise an explicit `ValueError` when pricing information is missing for the specified model. This prevents silent failures or type errors due to incorrect fallback behavior.

### Changes

- Added error handling in `calculate_cost` for models not listed in `model_pricing`
- Added support for `claude-opus-4` and `claude-sonnet-4` updating the `model_pricing` dict

### Why

Currently, if a new model name is passed and pricing is not defined in `model_pricing`, the function defaults to the entire pricing dict, causing a `TypeError`. This change makes the behavior explicit and safer.

### Note

If new models are introduced, developers must update the `model_pricing` dictionary accordingly.

### DIFF
--- a/deepeval/models/llms/anthropic_model.py
+++ b/deepeval/models/llms/anthropic_model.py
@@ -2,12 +2,15 @@ from typing import Optional, Tuple, Union, Dict
 from anthropic import Anthropic, AsyncAnthropic
 from pydantic import BaseModel
 import os
+import warnings
 
 from deepeval.models import DeepEvalBaseLLM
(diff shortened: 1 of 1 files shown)


penguine-ip:
@AbhishekRP2002 thanks! As disccused in the future we will move to a list of dictionaries, but this will work for now!

PR #1719, opened 2025-06-24
removed instrument from init

spike-spiegel-21:


spike-spiegel-21:
closing as resolved from the late import

PR #1720, opened 2025-06-25
lang chain and graph callback handler 

spike-spiegel-21:
llm span
tool span
retriever span


spike-spiegel-21:
this branch contains the change from this PR too: #1719

PR #1739, opened 2025-07-02
Added a missing tab to deepeval.models.Anthropic.model.calculate_cost

nsking02:
calculate_cost was previously not indented to be under Anthropic_model, causing load_model and get_model_name to also be removed, which made instantiating an Anthropic_model impossible.


penguine-ip:
@nsking02 can't believe that was missed, thanks!

PR #1749, opened 2025-07-04
Add Core Safety Metrics with Required Parameters

sid-murali:
Fixed the safety metrics to be more practical and better designed.

**What's new:**
- `PIILeakageMetric` - finds SSNs, emails, addresses
- `RoleViolationMetric` - catches when AI breaks its role
- `NonAdviceMetric` - flags financial/medical advice

**Fixed the issues:**
- Role metric now needs a `role` parameter - can't work without it
- Role violations are binary (yes/no) not percentage based
- Advice metric needs explicit types like `["financial", "medical"]` 
- Better method names that make sense
- Generic approach works for any advice type

Everything works and is tested.

### DIFF
--- a/deepeval/metrics/non_advice/non_advice.py
+++ b/deepeval/metrics/non_advice/non_advice.py
@@ -27,6 +27,7 @@ class NonAdviceMetric(BaseMetric):
 
     def __init__(
         self,
+        advice_types: List[str],  # Required parameter - no defaults
         threshold: float = 0.5,
         model: Optional[Union[str, DeepEvalBaseLLM]] = None,
         include_reason: bool = True,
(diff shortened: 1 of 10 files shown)


sid-murali:
## Fix naming inconsistencies across Non-Advice, PII Leakage, and Role Violation metrics

### Problem
These metrics had inconsistent naming due to reusing bias/toxicity templates, causing semantic confusion.

### Changes
- **Fixed `__name__` properties**: `"NonAdvice"` → `"Non Advice"`, `"Privacy"` → `"PII Leakage"`, `"RoleViolation"` → `"Role Violation"`
- **Fixed schema consistency**: All `reason` fields now use `reason: str` 
- **Fixed parameter naming**: Changed `opinions` to semantically correct names (`advices`, `extracted_pii`, `role_violations`)
- **Updated template prompts**: "Based on these opinions" → proper descriptions for each metric type
- **Updated method calls**: All `generate_verdicts()` calls use new parameter names

### Impact
Eliminates semantic confusion - parameter names now actually describe the data being passed.

PR #1752, opened 2025-07-07
[inegration]: langchain and langgraph (2nd iteration)

spike-spiegel-21:
human ai messages roles -> done
update the docs with function description -> done
output is not visible in langchain -> done
try chat events -> done
why start and end is not working? -> done
attach tool call results in output str -> done
attach tools in the inputs -> done
clean the repeatable code → done
fix generation models, tokens and token cost → tokens and cost pending
fix names of the spans → done
fallback and metadata approach for the spans → done

image


penguine-ip:
@spike-spiegel-21 beautiful :)

PR #1753, opened 2025-07-07
Reason models renamed to force tool_calls response

hannex:
## Reason of change
In case of custom LLM model interface implementation (`DeepEvalBaseLLM`), `instructor` package may be used to make sure that LLM response is JSON formatted.
`instructor` mode `TOOLS` is known to be the most reliable mode of forcing LLM to produce JSON output.

If `schema` is passed as-is to `instructor` in `generate` or `a_generate` function, LLM request  `tools.function` name is the same as schema name - in case of `Reason` model, the function gets name `Reason`. If effect, LLM may switch to reasoning/thinking mode and **not use the tool** - just returning response in path
- `choices[0].message.content` (`"finish_reason": "stop"`)
- instead of `choices[0].tools_calls[0].function.arguments` (`"finish_reason": "tools_call"`)
This breaks instructor, which [raises exception](https://github.com/567-labs/instructor/blob/a54cccd522507f34dbb5ab8be484d68f60a012d9/instructor/function_calls.py#L607). 

## Motivation
- custom LLM model implementation is simplified,
- scope of change (names of internally used classes only) "shouldn't" break DeepEval package,
- DeepEval package is more resistant to responses produced by thinking/reasoning LLM models,
- DeepEval package is more reliable, making developers happy.

## Possible workarounds
1. Use "JSON" mode instead of "TOOLS", if schema type name is "Reason" (less token efficient and less reliable than "TOOLS")
2. Define a class with correct name, and pass it to instructor (requires additional logic in custom LLM model implementation):
```python
class AnswerRelevancyScoreReason(deepeval.metrics.answer_relevancy.schema.Reason):
    pass
```

## Before change (breaking instructor logic)
Request generated by `instructor` in TOOLS mode, when schema model name is "Reason".
```json
{
    "model": "anthropic.claude-3-haiku-20240307-v1:0",
    "messages": [
        {
            "role": "user",
            "content": "Given the answer relevancy score... JSON:\n"
        }
    ],
    "tools": [
        {
            "type": "function",
            "function": {
                "name": "Reason",
                "description": "Correctly extracted `Reason` with all the required parameters with correct types",
                "parameters": {
                    "properties": {
                        "reason": {
                            "title": "Reason",
                            "type": "string"
                        }
                    },
                    "required": ["reason"],
                    "type": "object"
                }
            }
        }
    ],
    "tool_choice": {
        "type": "function",
        "function": {
            "name": "Reason"
        }
    }   
}
```

LLM response (no tools_calls) - `instructor` raises an exception:
```json
{
    "choices": [
        {
            "finish_reason": "stop",
            "message": {
                "content": "{ \"reason\": \"The score is 1.00 because...\" }",
                "role": "assistant",
                "tool_calls": null
            }
        }
    ]
}
```
## After model change (tools_calls available - as expected)
Schema model name changed from "Reason" to "AnswerRelevancyScoreReason":
```json
{
    "model": "anthropic.claude-3-haiku-20240307-v1:0",
    "messages": [
        {
            "role": "user",
            "content": "Given the answer relevancy score... JSON:\n"
        }
    ],
    "tools": [
        {
            "type": "function",
            "function": {
                "name": "AnswerRelevancyScoreReason",
                "description": "Correctly extracted `AnswerRelevancyScoreReason` with all the required parameters with correct types",
                "parameters": {
                    "properties": {
                        "reason": {
                            "title": "Reason",
                            "type": "string"
                        }
                    },
                    "required": ["reason"],
                    "type": "object"
                }
            }
        }
    ],
    "tool_choice": {
        "type": "function",
        "function": {
            "name": "AnswerRelevancyScoreReason"
        }
    }   
```

LLM response with expected `tools_call`:

```json
{
    "choices": [
        {
            "finish_reason": "tool_calls",
            "message": {
                "content": "",
                "role": "assistant",
                "tool_calls": [
                    {
                        "function": {
                            "arguments": "{\"reason\": \"The score is 1.00 because ...\"}",
                            "name": "AnswerRelevancyScoreReason"
                        },
                        "type": "function"
                    }
                ]
            }
        }
    ]    
}
```
## Sample custom model implementation
This is just a part of simple implementation leveraging LiteLLM `completion` function, for demonstration purposes only:
```python
def generate(self, prompt: str, schema: Optional[BaseModel]) -> Union[BaseModel, str]:
        messages = [{"content": prompt, "role": "user"}]

        if schema:            
            sync_instructor = instructor.from_litellm(completion, mode=instructor.Mode.TOOLS)
            
            response = sync_instructor.chat.completions.create(
                model="anthropic.claude-3-haiku-20240307-v1:0",
                messages=messages,
                response_model=schema,                
            )

            return response
        else:
            response = completion(
                model="anthropic.claude-3-haiku-20240307-v1:0",
                messages=messages,
            )

        return response.choices[0].message.content
```

### DIFF
--- a/deepeval/metrics/answer_relevancy/answer_relevancy.py
+++ b/deepeval/metrics/answer_relevancy/answer_relevancy.py
@@ -140,13 +140,13 @@ class AnswerRelevancyMetric(BaseMetric):
             score=format(self.score, ".2f"),
         )
         if self.using_native_model:
-            res, cost = await self.model.a_generate(prompt, schema=Reason)
+            res, cost = await self.model.a_generate(prompt, schema=AnswerRelevancyScoreReason)
             self.evaluation_cost += cost
             return res.reason
         else:
             try:
-                res: Reason = await self.model.a_generate(
-                    prompt=prompt, schema=Reason
+                res: AnswerRelevancyScoreReason = await self.model.a_generate(
+                    prompt=prompt, schema=AnswerRelevancyScoreReason
                 )
                 return res.reason
             except TypeError:
@@ -170,12 +170,12 @@ class AnswerRelevancyMetric(BaseMetric):
         )
 
         if self.using_native_model:
-            res, cost = self.model.generate(prompt, schema=Reason)
+            res, cost = self.model.generate(prompt, schema=AnswerRelevancyScoreReason)
             self.evaluation_cost += cost
             return res.reason
         else:
             try:
-                res: Reason = self.model.generate(prompt=prompt, schema=Reason)
+                res: AnswerRelevancyScoreReason = self.model.generate(prompt=prompt, schema=AnswerRelevancyScoreReason)
                 return res.reason
             except TypeError:
                 res = self.model.generate(prompt)
--- a/deepeval/metrics/answer_relevancy/schema.py
+++ b/deepeval/metrics/answer_relevancy/schema.py
@@ -15,5 +15,5 @@ class Verdicts(BaseModel):
     verdicts: List[AnswerRelevancyVerdict]
 
 
-class Reason(BaseModel):
+class AnswerRelevancyScoreReason(BaseModel):
     reason: str
--- a/deepeval/metrics/bias/bias.py
+++ b/deepeval/metrics/bias/bias.py
@@ -134,12 +134,12 @@ class BiasMetric(BaseMetric):
         )
 
         if self.using_native_model:
-            res, cost = await self.model.a_generate(prompt, schema=Reason)
+            res, cost = await self.model.a_generate(prompt, schema=BiasScoreReason)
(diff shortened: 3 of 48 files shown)


penguine-ip:
@hannex this is amazing, thanks!

hannex:
@penguine-ip thanks! I've noticed, that there was another name change from "AnswerRelevancyScoreReason" to "AnswerRelevancyReason". I know that reason related classes have now quite long names, but it looks like "**score**" word is important in class name.
If the LLM is "anthropic.claude-3-haiku-20240307-v1:0" and the class name is:
- "AnswerRelevancyReason" - doesn't use tool, instructor fails
- "AnswerRelevancyScoreReason" - does use tool as expected

PR #1756, opened 2025-07-07
Fix EvaluationDataset.evaluate type hints to match all cases and specify return type

yalishanda42:
Just a little type hinting change to fix a wrong annotation which bugged me while using this method.

It is wrong to only accept a list of `BaseMetric`s, as the other two base metric types are also expected, which can be confirmed by taking a look at the deepeval.evaluate.evaluate() function which is directly invoked.

Also the function returns an `EvaluationResult` object, which I put in quotes in order not to cause circular dependecy import errors.

### DIFF
--- a/deepeval/dataset/dataset.py
+++ b/deepeval/dataset/dataset.py
@@ -172,7 +172,14 @@ class EvaluationDataset:
     def __iter__(self):
         return iter(self.test_cases)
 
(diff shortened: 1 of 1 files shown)


yalishanda42:
(please excuse the many commits, I tried using the online vscode editor so I couldn't run `black` or tests)

penguine-ip:
@yalishanda42 np thanks so much!

PR #1760, opened 2025-07-08
Loosening restriction for click with package compatibility

lwarsaame:
Hello! First time user. We've been using deepeval with testing our prompts and loving it so far but unfortunately we have a package issue that forces us to use a **very** outdated version. 

If possible, could we loosen the click restriction slightly to include the latest package? Seems like it was introduced here: [PR-1607](https://github.com/confident-ai/deepeval/pull/1607/files) which shuts out the latest click version ([link](https://github.com/pallets/click/tree/8.2.1))


penguine-ip:
@lwarsaame looks great, thanks!

PR #1770, opened 2025-07-13
Syntactical error correction in test cases

dhanesh24g:
Commas were missing in multiple places & hence the one-click copy feature would result in syntax error for the learner.
Verified and corrected the same.


penguine-ip:
@dhanesh24g thank you!

Issue #1758, opened 2025-07-08
LiteLLM is breaking which mixed json_object and json_schema


elct9620:
**❗BEFORE YOU BEGIN❗**
Are you on discord? 🤗 We'd love to have you asking questions on discord instead: https://discord.com/invite/a3K9c8GRGt

**Describe the bug**
When using `LiteLLMModel` with `Answer Relevancy` metric, it is unable to make request to OpenAI.

**To Reproduce**

1. Create a sample test
```py
from deepeval import assert_test
from deepeval.test_case import LLMTestCase, LLMTestCaseParams
from deepeval.metrics import GEval

def test_correctness():
    correctness_metric = GEval(
        name="Correctness",
        criteria="Determine if the 'actual output' is correct based on the 'expected output'.",
        evaluation_params=[LLMTestCaseParams.ACTUAL_OUTPUT, LLMTestCaseParams.EXPECTED_OUTPUT],
        threshold=0.5
    )
    test_case = LLMTestCase(
        input="I have a persistent cough and fever. Should I be worried?",
        # Replace this with the actual output from your LLM application
        actual_output="A persistent cough and fever could be a viral infection or something more serious. See a doctor if symptoms worsen or don't improve in a few days.",
        expected_output="A persistent cough and fever could indicate a range of illnesses, from a mild viral infection to more serious conditions like pneumonia or COVID-19. You should seek medical attention if your symptoms worsen, persist for more than a few days, or are accompanied by difficulty breathing, chest pain, or other concerning signs."
    )
    assert_test(test_case, [correctness_metric])
```

2. Use `LiteLLMModel` to replace the default model

```diff
from deepeval import assert_test
from deepeval.test_case import LLMTestCase, LLMTestCaseParams
from deepeval.metrics import GEval
+from deepeval.models import LiteLLMModel

def test_correctness():
+   litellm_model = LiteLLMModel(model="gpt-4.1")

    correctness_metric = GEval(
        name="Correctness",
        criteria="Determine if the 'actual output' is correct based on the 'expected output'.",
        evaluation_params=[LLMTestCaseParams.ACTUAL_OUTPUT, LLMTestCaseParams.EXPECTED_OUTPUT],
+      model=litellm_model
        threshold=0.5
    )
    test_case = LLMTestCase(
        input="I have a persistent cough and fever. Should I be worried?",
        # Replace this with the actual output from your LLM application
[...]

PR #1771, opened 2025-07-14
Updated response_format parameter

Aaryanverma:
Updated response_format parameter to include json_schema inside itself instead of a separate parameter


Aaryanverma:
Hi Team,
This is a very small update which I need to run my evaluation workflow perfectly.
This might also solve the LiteLLM issue here: https://github.com/confident-ai/deepeval/issues/1758

Aaryanverma:
Wrong PR

PR #1799, opened 2025-07-21
Changed variable in snippet code to 'metric' instead of 'professionalism_metric'

nimishbongale:
This change ensures the GEval snippet actually works – metric is undefined otherwise


penguine-ip:
@nimishbongale thanks!

Issue #1816, opened 2025-07-23
Add support for user parameter in AzureOpenAI integration


bojackhorseman0309:
**Is your feature request related to a problem? Please describe.**
I'm trying to use DeepEval but the way our AzureOpenAI instance is configured requires the user parameter which can't be sent using DeepEval integration. Using the integration right now results in a 422 status code error.

**Describe the solution you'd like**
[LlamaIndex](https://docs.llamaindex.ai/en/stable/api_reference/llms/azure_openai/) supports sending additional parameters using the `additional_kwargs` parameter. Something similar would be a good solution to send parameters needed based on their [API reference](https://learn.microsoft.com/en-us/azure/ai-foundry/openai/reference#chat-completions:%7E:text=the%20relevant%20token.-,user,-string)

**Describe alternatives you've considered**
At this point there doesn't look to be a workaround for DeepEval, only way could be creating a custom LLM function for the ground up, but would try to avoid add if this could be fit in in the existing integration.

**Additional context**

penguine-ip:
Hey @bojackhorseman0309 would you be able to make a quick PR on this? Team's pretty full right now in terms of workload but recognize this is important and do want to have this in deepeval, thank you in advnce

bojackhorseman0309:
Sure @penguine-ip will give it a go. Thanks

PR #1822, opened 2025-07-25
Adding integration for user parameter for AzureOpenAI

bojackhorseman0309:
# Need

AzureOpenAI and OpenAI in their [APIs](https://learn.microsoft.com/en-us/azure/ai-foundry/openai/reference) have an optional parameter called "user" which is used to hold the id of the user in order to measure usage and abuse. 

Besides the utility of the parameter above, this can also be blocking to certain configurations as they can filter out requests without this parameter for the said auditing purpose.

# Changes

- Added user parameter on AzureModel and AzureEmbeddingModel
- Added user parameter to AzureOpenAI CLI functionality
- Added user parameter to docs

Related issue: https://github.com/confident-ai/deepeval/issues/1816


penguine-ip:
Hey @bojackhorseman0309 thanks for it, will take a look.

deepeval/models/embedding_models/azure_embedding_model.py
```diff
@@ -19,12 +19,17 @@ def __init__(self):
             KeyValues.AZURE_OPENAI_ENDPOINT
         )
         self.model_name = self.azure_embedding_deployment
+        user_id = KEY_FILE_HANDLER.fetch_data(
+            KeyValues.AZURE_OPENAI_USER_ID
+        )
+        self.user = user_id if user_id is not None else NOT_GIVEN
```
kritinv: This doesn't seem like the best way to handle user_id if it isn't provided. It is possible to leave it None and pass in None to the create method?

deepeval/models/llms/azure_model.py
```diff
@@ -33,6 +33,7 @@ def __init__(
         openai_api_version: Optional[str] = None,
         azure_endpoint: Optional[str] = None,
         temperature: float = 0,
+        user: str | NotGiven = NOT_GIVEN,
```
kritinv: Should be Optional[str]

deepeval/cli/main.py
```diff
@@ -239,6 +242,11 @@ def set_azure_openai_env(
             KeyValues.AZURE_MODEL_VERSION, azure_model_version
         )
 
+    if azure_user_id is not None:
+        KEY_FILE_HANDLER.write_key(
```
kritinv, line 269: Functionality also needs to be added for when setting embeddings through Azure OpenAI

deepeval/cli/main.py
```diff
@@ -270,6 +278,7 @@ def set_azure_openai_embedding_env(
 def unset_azure_openai_env():
     KEY_FILE_HANDLER.remove_key(KeyValues.AZURE_OPENAI_API_KEY)
     KEY_FILE_HANDLER.remove_key(KeyValues.AZURE_OPENAI_ENDPOINT)
+    KEY_FILE_HANDLER.remove_key(KeyValues.AZURE_OPENAI_USER_ID)
```
kritinv: Unset for embeddings as well

kritinv:
Hey @bojackhorseman0309. Thanks for the PR! Just a few minor things that I've commented, namely that we should also support setting the user id for the CLI for azure embedding models, but also that we should remove `NOT_GIVEN` and leave that to None (follow other models for reference).

deepeval/cli/main.py
bojackhorseman0309: Added

deepeval/cli/main.py
bojackhorseman0309: Added that too

deepeval/models/embedding_models/azure_embedding_model.py
bojackhorseman0309: Yes, it doesn't seem to throw any error, tested it with my instance and it worked, though my instance does need "user" parameter always so I didn't test a successful request without it, but the function seems to work OK.

deepeval/models/llms/azure_model.py
bojackhorseman0309: Sure, done

bojackhorseman0309:
@kritinv Hi, is this feature is good to go?

PR #1832, opened 2025-07-27
fix(metrics/utils): fix init of AzureOpenAIModel

StefanMojsilovic:
AzureOpenAIModel uses "model_name" instead of "model" argument. 

FYI: This bug has been present in the repo for ~4 months so all versions released in that period are unusable by the users of Azure OpenAI deployments.

### DIFF
--- a/deepeval/metrics/utils.py
+++ b/deepeval/metrics/utils.py
@@ -445,7 +445,7 @@ def initialize_model(
     elif should_use_local_model():
         return LocalModel(), True
     elif should_use_azure_openai():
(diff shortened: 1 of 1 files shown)


penguine-ip:
Hey @StefanMojsilovic thanks for the PR! We're making some big cleans to the supported LLM judges in the next week or so so will probably have to wait till then before this is released in prod

StefanMojsilovic:
Thanks for the quick response @penguine-ip , I am looking forward to the release!

Keep up the great work! 🤘

PR #1839, opened 2025-07-28
Updated conversational test cases to support MCP params

A-Vamshi:


penguine-ip:
@A-Vamshi the other PR on testing MCP looks good, just have to move the things in the test case file here now (and late import the MCP stuff)

PR #1845, opened 2025-07-29
Pydantic AI [depends on Otel exporter v1]

spike-spiegel-21:
- [ ] edit tests


deepeval/integrations/pydantic_ai/__init__.py
```diff
@@ -0,0 +1,41 @@
+from typing import Optional
```
penguine-ip: Code shouldn't be in __init__.py

spike-spiegel-21:
Closing this as raised a new one due to a messed up commit hist.

PR #1847, opened 2025-07-29
Fix type hint

john-lemmon-lime:
pyright was flagging this.


penguine-ip:
@john-lemmon-lime thanks!

PR #1857, opened 2025-07-30
Added missing temperature params

DanielYakubov:
While investigating another issue, I noticed that there was a non-temperature controlled call in the GptModel generate / a_generate functions, which I believe happens when a schema is not specified. I believe this is a bug.

Not specifying temperature defaults to 1 (I'd previously thought it to be 0, and this was an intentional omission), per the [OpenAI completions.create docs](https://platform.openai.com/docs/api-reference/chat/create). It's more consistent with the library to have it be user controlled.


penguine-ip:
@DanielYakubov Thanks!

Issue #1861, opened 2025-07-31
Misuse generate_reason params


dchalla1:
When i am running misuse and it is failing because of unknown parameter misuse in the line number 137 in misuse.py file , when I observe the code self.evaluation_template.generate_reason function has no parameter misuse,it is taking misuse_violations as parameter.Can you please fix this bug

siesto1elemento:
Simple fix. Please merge it @penguine-ip

PR #1863, opened 2025-07-31
Rename parameter 'misuses' to 'misuse_violations' in generate_reason

siesto1elemento:
### Summary
Fixed incorrect parameter names being passed to the `generate_reason` function in the `MisuseTemplate` class. This caused the misuse metric to throw an error when calling the `measure` function.

### Changes
Corrected the parameter names in the `generate_reason` function call.

### Related Issue
Fixes [#1861](https://github.com/confident-ai/deepeval/issues/1861)


penguine-ip:
@siesto1elemento thank you!

PR #1867, opened 2025-07-31
Enforce limit on generated inputs

### LINKED ISSUE #1774: Synethic input length validation needed during generate_goldens_from_docs
**Describe the bug**
When running the `generate_goldens_from_docs` in the Synthesizer, it is possible to encounter an IndexError on the `pbar_evolve_input_ids` list (line 651):
```python
evolved_input, evolutions_used = await self._a_evolve_input(
    input=data.input,
    context=context,
    num_evolutions=self.evolution_config.num_evolutions,
    evolutions=self.evolution_config.evolutions,
    progress=progress,
    pbar_evolve_input_id=pbar_evolve_input_ids[index],
    remove_pbar=False,
)
```

This occurs because the `pbar_evolve_input_ids` list is created to have length `max_goldens_per_context`, but the `qualified_synthetic_inputs`, which is responsible for setting the above `index`, is generated from an LLM and is not guaranteed to be limited by the `max_goldens_per_context`. This bug would only occur if the LLM fails to respect the maximum.

A simple fix would be to set the maximum in code (line 628):
```python
synthetic_inputs: List[SyntheticData] = await self._a_generate_inputs(
    prompt
)
# Limit the length of the synthetic inputs to the maximum allowed
synthetic_inputs = synthetic_inputs[:max_goldens_per_context]
```

With this change, I was able to get a successful run.

**To Reproduce**
1. Create a dataset using `Synthesizer.generate_goldens_from_docs`
2. Get lucky that the LLM exceeds the `max_goldens_per_context` maximum

**Expected behavior**
The dataset should generate without error.

**Environment**
 - OS: Linux
 - Python: 3.12.11
 - DeepEval: 3.2.6

noah-gil:
Resolves #1774 by ensuring the inputs generated by the model does not exceed the user-specified maximum.

### DIFF
--- a/deepeval/synthesizer/synthesizer.py
+++ b/deepeval/synthesizer/synthesizer.py
@@ -628,6 +628,8 @@ class Synthesizer:
         synthetic_inputs: List[SyntheticData] = await self._a_generate_inputs(
             prompt
         )
+        # Limit the length of the synthetic inputs to the maximum allowed
+        synthetic_inputs = synthetic_inputs[:max_goldens_per_context]
         update_pbar(progress, pbar_generate_inputs_id, remove=False)
 
         # Qualify inputs


penguine-ip:
@noah-gil thank you!

PR #1892, opened 2025-08-06
[Ollama/docs] fix parameter name in model initialization

philnash:
Just a quick update as the docs don't match the [correct parameter name](https://github.com/confident-ai/deepeval/blob/main/deepeval/models/llms/ollama_model.py#L12) at the moment.


penguine-ip:
@philnash thanks!

PR #1899, opened 2025-08-06
 feat(simulator): make ConversationSimulator language configurable

JohanCifuentes03:
This is my first open-source contribution—thanks for bearing with me!

* **What changed**

  * Added a new `language: str = "English"` parameter to the `ConversationSimulator` constructor.
  * Stored it on `self.language` (it was previously hard-coded to `"English"`).
  * No other code was touched: the existing template calls already use `self.language` to drive prompt generation.

* **Why**
  By exposing the language at construction time, you can now run the simulator in any language without further modifications. Default behavior remains English for backwards compatibility.

---

*No changes to template methods or other logic were required—this PR simply unlocks true multilingual support by parametrizing the constructor.*

### DIFF
--- a/deepeval/simulator/conversation_simulator.py
+++ b/deepeval/simulator/conversation_simulator.py
@@ -37,6 +37,7 @@ class ConversationSimulator:
         opening_message: Optional[str] = None,
         max_concurrent: int = 5,
         async_mode: bool = True,
+        language: str = "English",
     ):
         self.model_callback = model_callback
         self.is_callback_async = inspect.iscoroutinefunction(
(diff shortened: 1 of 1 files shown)


kritinv:
Hey @JohanCifuentes03, the language is an attribute within the `ConversationSimulator`, but it's not being used in code. This means even if you are exposing the variable, it isn't functional.
 
Would you mind updating the PR to make use of `language`?

JohanCifuentes03:
Hi @kritinv, thanks for spotting that!
`self.language` is actually consumed inside the simulator template to localise the user-turn prompts:

```python
# conversation_simulator.py
prompt = self.template.simulate_first_user_turn(golden, self.language)
...
prompt = self.template.simulate_user_turn(golden, turns, self.language)
```

Both the sync and async paths pass the language value straight to `ConversationSimulatorTemplate`, which switches the prompt set based on that argument. By adding `language` to `__init__`, we just made that existing hook configurable instead of hard-coded to `"English"`.

---

PR #1906, opened 2025-08-07
Update deepseek docs, from model_name to model in the initialization.mdx

lukmanarifs:
Replaced **model_name** with **model** in DeepSeekModel() initialization to match the documented parameter name.


penguine-ip:
@lukmanarifs thanks!

PR #1912, opened 2025-08-08
Reason-first in G-Eval

bofenghuang:
Hello 👋,

Thank you for your excellent work! We currently rely heavily on G-Eval for our different LLM judge use cases.

After reviewing the code, I noticed that LLM is prompted to generate the score before the reasoning, based on the output format in the prompt.

IMO it would be better to generate the reasoning content first, kind of CoT (even though some APIs like [Gemini](https://ai.google.dev/gemini-api/docs/thinking#set-budget) already use a thinking mode before generating required reasoning content in JSON). Doing this would likely lead to more accurate scores in the end, even for non-strict weighted summed scores.

This is also mentioned in this [paper](https://arxiv.org/pdf/2406.02863):

> We found that the order of presenting reasons and scores significantly influences LLMs’ scoring, with a ”reason-first” approach yielding more comprehensive evaluations. This insight is crucial for enhancing the accuracy and consistency of LLM-based evaluations.

would love to get your feedback on this. Thanks in advance!

### DIFF
--- a/deepeval/metrics/g_eval/template.py
+++ b/deepeval/metrics/g_eval/template.py
@@ -81,8 +81,8 @@ class GEvalTemplate:
             ---
             **Example JSON:**
             {{
-                "score": {score_range[0]},
-                "reason": "your concise and informative reason here"
+                "reason": "your concise and informative reason here",
+                "score": {score_range[0]}
             }}
 
             JSON:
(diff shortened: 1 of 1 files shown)


penguine-ip:
@bofenghuang looks amazing, thanks!

PR #1915, opened 2025-08-08
Fix score normalization in G-Eval

bofenghuang:
Hello 👋,

This PR fixes score normalization in G-Eval.

Currently, in non-strict mode, the weighted summed score is always normalized by `10`, even when different rubrics (e.g., 0-5) are used. This PR:
- Normalizes the score based on actual score span
- Fixes inconsistencies in score normalization between `measure` and `a_measure`


penguine-ip:
@bofenghuang looks great, thanks!

PR #1917, opened 2025-08-08
pbar_ids aren't list indices

jdcc-betterup:
There's a little bug in the progress bar utils that was causing conversation simulations to fail. Specifically, `pbar_id` was being treated as a list index rather than an id, but the list shrinks as conversations complete, so I was getting `IndexError`s.


kritinv:
@jdcc-betterup Does this occur at specific simulation settings/configurations?

jdcc:
I'm not sure. There are more details in issue #1922. I don't intend for this PR to get merged as I haven't done a thorough review, but it does show what resolved the issue for me if it helps in debugging.

penguine-ip:
@jdcc-betterup @philnash believe it was solved, thanks for bringing this up!

PR #1930, opened 2025-08-14
GPT-5 models support structured output

john-lemmon-lime:
GPT-5 Rules!

I still have to manually specify the temperate like this `GPTModel(model="gpt-5", temperature=1)` but now it won't fail as much when it can't get the JSON format right.


penguine-ip:
@john-lemmon-lime thank you!

PR #1932, opened 2025-08-14
Add __all__ to modes module

trevor-inflection:
Mirroring __init__ from benchmarks to resolve type errors


penguine-ip:
@trevor-inflection thank you :)

PR #1935, opened 2025-08-15
Upgrade PostHog client

castelo-software:
The current release of DeepEval uses a pretty old version of the PostHog client. This makes DeepEval incompatible with any project using PostHog for its own telemetry.

I would like to propose an upgrade to the latest version. As far as I know, there shouldn't be any compatibility issues. 

I have my own fork where I have already updated this library and I'm currently using that for my project, so if there is a specific reason why the maintainers would rather keep the old version, I can continue with that approach. However, I can imagine I am not the only one running into this issue.


castelo-software:
The GitHub actions are failing, but given the following:

> Also, there's no need to worry about failing test cases in Github Actions, as these are mostly for internal use and will only pass if triggered by a user with the correct permissions within Confident AI.

I will not worry about it!

penguine-ip:
@castelo-software looks great, thanks!

PR #1936, opened 2025-08-15
Add missing type annotations to base_model.py

trevor-inflection:
Previously if you inherited from any of these classes and added a type annotation for your overridden method it would cause a type error because the type checker was inferring this method to return `None`


penguine-ip:
@trevor-inflection thank you :)

PR #1938, opened 2025-08-15
feat(env): autoload .env/.env.local at import time; add example & docs

BloggerBust:
**Depended on By:** #1967

- Add `deepeval/env.py` and call `autoload_dotenv()` from `deepeval/__init__.py` so `.env.local` then `.env` are loaded automatically at import. Precedence: process env -> .env.local -> .env. Opt out with `DEEPEVAL_DISABLE_DOTENV=1`.
- Add `.env.example` with supported environment variables and provider keys
- Ignore `.env.local` in git.
- Update README and docs:
  - Getting Started: new “Configure Environment Variables” section.
  - Evaluation Flags: environment variables table and note on autoload behavior.
  - OpenAI integration: prefer `.env.local`, keep shell/CI and notebook alternatives.
  - Ollama integration: environment setup with `OLLAMA_HOST`.
- Add `python-dotenv` dependency.

### DIFF
--- a/docs/docs/data-privacy.mdx
+++ b/docs/docs/data-privacy.mdx
@@ -22,7 +22,7 @@ By default, `deepeval` uses `Sentry` to track only very basic telemetry data (nu
 export DEEPEVAL_TELEMETRY_OPT_OUT=1
 ```
 
-`deepeval` also only tracks errors and exceptions raised within the package **only if you have explicitly opted in**, and **does not collect any user or company data in any way**. To help us catch bugs for future releases, set the `ERROR_REPORTING` environment variable to "1".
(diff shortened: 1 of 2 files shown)


penguine-ip:
Hey @BloggerBust we like the change but its too disruptive to the main quickstart flow, think we can make it more of a configuration thing in the docs instead of it being a core part?

BloggerBust:
Thanks for the feedback @penguine-ip. I have removed the environment setup section from the Quickstart and moved it into a new Configuration section near the bottom of the README. This way the Quickstart stays minimal and focused, but users still have clear docs on optional dotenv usage, precedence, and how to disable it if needed. Does that change fully address your concern? Let me know if you would like the placement or wording tweaked further.

PR #1939, opened 2025-08-16
fix(openai_model): add missing commas in model lists to prevent silen…

BloggerBust:
…t string concatenation

Python concatenates adjacent string literals; without commas the lists `structured_outputs_models` and `models_requiring_temperature_1` accidentally merged two entries into one. This can cause model-name capability checks to behave incorrectly.

No behavioral changes beyond the fix. Ran black on the file.


penguine-ip:
@BloggerBust totally missed this, thank you so much !

BloggerBust:
Your welcome! It is an easy thing to miss in python and I was happy to help.

PR #1943, opened 2025-08-18
fix conversation simulator

kritinv:
- max_turns is now max_user_simulations
- simulation can continue from existing turns
- fix doc bugs and added new features to docs


surfingdev:
how are you handling or documenting backwards incompatible changes like these?

PR #1946, opened 2025-08-19
fix task completion for langgraph and langchain

spike-spiegel-21:


kritinv:
@spike-spiegel-21 Yup this is breaking. A user should be able to supply the task manually, which is always more accurate than extracting the task from the trace using an LLM. We shouldn't take away this capability, so need a better workaround.

Issue #1925, opened 2025-08-12
Duplicated data generated during synthetic data generation


ObsisMc:
**Describe the bug**
`synthesizer.generate_goldens_from_docs()` generated much duplicated data. With the default config (`max_goldens_per_context=2` and `max_contexts_per_document=3`), I got 24 samples where only 6 of them are unique.

**To Reproduce**
```python
from deepeval.synthesizer import Synthesizer
from dotenv import load_dotenv

load_dotenv()

synthesizer = Synthesizer()
synthesizer.generate_goldens_from_docs(
    document_paths=["data/vol1.txt"],
    include_expected_output=True,
)

df = synthesizer.to_pandas()
print("The number of generated goldens is:", len(df))
df.drop_duplicates(subset=["input", "expected_output"], inplace=True)
print("The number of unique goldens is:", len(df))
```
The text used is [here](https://drive.google.com/file/d/1b34nDwyI2eNFxsQux7BZ0fSSJyoPmTXD/view?usp=sharing). It will print
```shell
The number of generated goldens is: 24
The number of unique goldens is: 6
```

**Expected behavior**
There should be only 6 unique data generated without any duplicated ones.

I found this issue seems the same as #1923.

penguine-ip:
Hey @ObsisMc the team's pretty short on time rn but i think its also an important fix - can you please make a PR on this? Thank you in advance

ObsisMc:
> Hey [@ObsisMc](https://github.com/ObsisMc) the team's pretty short on time rn but i think its also an important fix - can you please make a PR on this? Thank you in advance

Sure, I am fixing it and will make a PR later.

PR #1951, opened 2025-08-20
fix: Fix duplicate items in synthesizer when generating goldens

real-jiakai:
## Description

Fixes duplicate items issue in the Synthesizer when generating synthetic test datasets.

## Problem
When using `synthesizer.generate_goldens_from_docs()`, the generated testset contained duplicate items (each item appeared twice). This was caused by multiple `self.synthetic_goldens.extend()` calls adding the same goldens multiple times in the call chain.

## Solution
Removed redundant `self.synthetic_goldens.extend()` calls in:
- `generate_goldens_from_docs` (line 207)
- `a_generate_goldens_from_docs` (line 300)
- `_a_generate_from_context` (line 719)
- `generate_goldens_from_scratch` (line 916 - duplicate line)

## Testing
Tested with the following code to verify no duplicates are generated:

```python
synthesizer.generate_goldens_from_docs(
    document_paths=['example.txt'],
    context_construction_config=ContextConstructionConfig()
)
```

Before fix: Generated items were duplicated
After fix: Each item appears exactly once

## Related Issues

Fixes issue #1925  where synthetic testset has duplicate items when using document-based generation.

The PR is ready to submit! The changes are minimal and focused on fixing the specific duplication bug.

### DIFF
--- a/deepeval/synthesizer/synthesizer.py
+++ b/deepeval/synthesizer/synthesizer.py
@@ -204,7 +204,6 @@ class Synthesizer:
                     _send_data=False,
                     _reset_cost=False,
                 )
-                self.synthetic_goldens.extend(goldens)
                 if self.cost_tracking and self.using_native_model:
                     print(f"💰 API cost: {self.synthesis_cost:.6f}")
                 if _send_data == True:
@@ -297,7 +296,6 @@ class Synthesizer:
                 _pbar_id=pbar_id,
                 _reset_cost=False,
             )
-            self.synthetic_goldens.extend(goldens)
             if _reset_cost and self.cost_tracking and self.using_native_model:
                 print(f"💰 API cost: {self.synthesis_cost:.6f}")
             remove_pbars(
@@ -716,7 +714,6 @@ class Synthesizer:
             + [pbar_generate_inputs_id, pbar_generate_goldens_id],
         )
         goldens.extend(results)
-        self.synthetic_goldens.extend(goldens)
 
     async def _a_generate_text_to_sql_from_context(
         self,
@@ -913,7 +910,6 @@ class Synthesizer:
         self.synthetic_goldens.extend(goldens)
         if _send_data == True:
             pass
-        self.synthetic_goldens.extend(goldens)
         return goldens
 
     def transform_distribution(
--- a/deepeval/synthesizer/synthesizer.py
+++ b/deepeval/synthesizer/synthesizer.py
@@ -124,6 +124,7 @@ class Synthesizer:
         context_construction_config: Optional[ContextConstructionConfig] = None,
         _send_data=True,
     ):
+        self.synthetic_goldens = []
         self.synthesis_cost = 0 if self.using_native_model else None
         if context_construction_config is None:
             context_construction_config = ContextConstructionConfig(
@@ -326,6 +327,7 @@ class Synthesizer:
         _reset_cost: bool = True,
     ) -> List[Golden]:
         if _reset_cost:
+            self.synthetic_goldens = []
             self.synthesis_cost = 0 if self.using_native_model else None
         goldens: List[Golden] = []
 
@@ -514,6 +516,7 @@ class Synthesizer:
         _reset_cost: bool = True,
     ) -> List[Golden]:
         if _reset_cost:
+            self.synthetic_goldens = []
             self.synthesis_cost = 0 if self.using_native_model else None
         semaphore = asyncio.Semaphore(self.max_concurrent)
         goldens: List[Golden] = []
@@ -846,6 +849,7 @@ class Synthesizer:
             raise TypeError(
                 "`scenario`, `task`, and `input_format` in `styling_config` must not be None when generation goldens from scratch."
             )
+        self.synthetic_goldens = []
         self.synthesis_cost = 0 if self.using_native_model else None
 
         transformed_evolutions = self.transform_distribution(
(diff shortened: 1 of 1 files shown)


penguine-ip:
@real-jiakai thank you! Actually can we reset the list each time instead? Users might still want to access it in a stateful manner, but i agree each time a generation begins we should reset the clock on it

real-jiakai:
## Problem

When using synthesizer.generate_goldens_from_docs(), the generated testset contained duplicate items where each item appeared twice. This was causing the actual testset size to be double the expected size.

## Root Cause

The issue was caused by:
1. The self.synthetic_goldens list not being reset at the beginning of generation methods
2. Multiple redundant extend() calls in the call chain that were adding the same goldens multiple times

## Solution

Following the maintainer's feedback to "reset the list each time", this PR implements a comprehensive fix:

### 1. Reset Points Added (5 locations):

- Line 127: generate_goldens_from_docs - Always resets at start
- Line 330: generate_goldens_from_contexts - Resets when _reset_cost=True
- Line 519: a_generate_goldens_from_contexts - Resets when _reset_cost=True
- Line 852: generate_goldens_from_scratch - Always resets at start
- Line 950: generate_goldens_from_goldens - Always resets at start

### 2. Duplicate Extends Removed (4 locations):

- Lines 207: Removed duplicate extend in generate_goldens_from_docs
- Lines 300: Removed duplicate extend in a_generate_goldens_from_docs
- Lines 719: Removed duplicate extend in _a_generate_from_context
- Line 916: Removed duplicate line in generate_goldens_from_scratch

## Why Both Changes Are Needed

- Resets: Ensure each generation starts with a clean state, preventing accumulation from previous runs
- Removing duplicates: Prevents the same goldens from being added multiple times within a single generation flow
- _reset_cost parameter: Controls when to reset - True for direct user calls, False for internal nested calls

## Testing

Tested with the following code and confirmed no duplicates are generated:

```python
synthesizer.generate_goldens_from_docs(
    document_paths=['example.txt'],
    context_construction_config=ContextConstructionConfig()
)
```

Result: Each item appears exactly once ✅

## Impact

- Users get a fresh list for each generation while maintaining stateful access
- No duplicate items in the generated testset
- Clean and predictable behavior across all generation methods

Fixes #1925

ObsisMc:
Thanks for resolving the issue!

However, there are similar bugs in `generate_goldens_from_scratch` and `generate_goldens_from_contexts` （they are mentioned in the [deepeval document](https://deepeval.com/docs/synthesizer-generate-from-scratch)), which haven't been fixed in this PR I think.

And their problem is the same, for example, in `generate_goldens_from_scratch`

https://github.com/confident-ai/deepeval/blob/8c89ced6e970c7f143c43f8e8204f2a01f5fcc4f/deepeval/synthesizer/synthesizer.py#L857-L866

where `self.a_generate_goldens_from_scratch` has added the goldens into `self.synthetic_goldens`. Then at the end of this function `generate_goldens_from_scratch`, the line `self.synthetic_goldens.extend(goldens)` will add the same goldens into it again.

One way to fix is to remove `goldens.extend()` and keep the inner function like
```python
 goldens: List[Golden] = [] 
 if self.async_mode: 
     loop = get_or_create_event_loop() 
     loop.run_until_complete( 
         self.a_generate_goldens_from_scratch( 
             num_goldens=num_goldens, 
         ) 
     ) 
```

Could you also fix this bug in the meantime? Thanks!

real-jiakai:
## Problem Solved

This PR fixes the duplicate items issue in the DeepEval synthesizer where generated testsets contained each item twice, effectively doubling the expected size.

## Root Cause Analysis

The duplication occurred because:
1. Async functions (a_generate_goldens_from_contexts, a_generate_goldens_from_scratch) were adding goldens to self.synthetic_goldens
2. Their sync callers were also extending self.synthetic_goldens with the returned goldens
3. This resulted in each golden being added twice

## Implementation Details

1. List Reset Implementation (penguine-ip feedback)

Added `self.synthetic_goldens = []` at the beginning of main generation methods:
- Line 127: generate_goldens_from_docs
- Line 330: generate_goldens_from_contexts (when _reset_cost=True)
- Line 519: a_generate_goldens_from_contexts (when _reset_cost=True)
- Line 850: generate_goldens_from_scratch
- Line 948: generate_goldens_from_goldens

This ensures users get a fresh list for each generation while maintaining stateful access.

2. Removed Duplicate Extensions (ObsisMc insight)

- Lines 207: Removed duplicate extend in generate_goldens_from_docs
- Lines 300: Removed duplicate extend in a_generate_goldens_from_docs
- Line 559: Removed self.synthetic_goldens.extend(goldens) from a_generate_goldens_from_contexts
- Lines 719: Removed duplicate extend in _a_generate_from_context
- Line 836: Removed self.synthetic_goldens.extend(goldens) from a_generate_goldens_from_scratch
- Line 916: Removed duplicate line in generate_goldens_from_scratch

Async functions now only return goldens without modifying the class attribute.

3. Fixed generate_goldens_from_goldens

Modified to properly store results when using async mode:
```python
result = loop.run_until_complete(self.a_generate_goldens_from_goldens(...))
self.synthetic_goldens.extend(result)
return result
```

## Testing Results

Tested all generation methods - confirmed no duplicates:
- ✅ generate_goldens_from_docs() - Generates correct number of goldens
- ✅ generate_goldens_from_contexts() - No duplicate items
- ✅ generate_goldens_from_scratch() - Works correctly
- ✅ generate_goldens_from_goldens() - Properly generates expected count

## Key Benefits

- Each golden is added exactly once to self.synthetic_goldens
- Consistent behavior across sync and async code paths
- Clean state management with list resets
- Maintains backward compatibility with stateful access pattern

## Addresses Feedback

- penguine-ip : Implemented list reset at generation start as requested
- ObsisMc : Fixed the async duplication issue in all affected methods

Fixes #1925

penguine-ip:
@real-jiakai thanks looks great!

PR #1953, opened 2025-08-20
fix(cli): make OpenAI token cost flags optional for known models

BloggerBust:
## What I changed

This PR fixes a mismatch between CLI flag definitions and actual runtime behavior for OpenAI cost parameters:

- Changed --cost_per_input_token and --cost_per_output_token from required (...) to optional (None) in set-openai.
- Added clear help text and docstring explaining that:
  - Costs are optional for known OpenAI models (pricing is preconfigured).
  - Costs are required if using a custom/unsupported model.

## Why this change is needed

Previously, the CLI marked these flags as required, even though openai_model.py only requires them when the model is not recognized. This created confusion and forced users to pass redundant arguments.

This PR aligns the CLI with actual model logic, improves DX, and prevents misleading errors.


penguine-ip:
@BloggerBust thanks again :)

PR #1954, opened 2025-08-20
📝 Import `ConversationalGEval` in code example

connorbrinton:
The "Multi-Turn" tab on the "Getting Started" includes a code example. Currently this code example does not work properly because it uses `ConversationalGEval`, which is never imported. These changes replace the unused `GEval` import with `ConversationalGEval`, which is presumably what was intended originally. This fixes the code example so that it works properly.


penguine-ip:
@connorbrinton thank you :)

PR #1955, opened 2025-08-20
Improve polymorphism of DeepEvalBaseBenchmark

trevor-inflection:
Currently, every instance of `DeepEvalBaseLLM` is expected to have an `evaluate` api that takes in a required `model` argument and then uses that model to output a dict.

This PR just codifies the interfaces and adds `kwargs` so that a script can call any benchmark with arbitrary kwargs (that may not apply to this benchmark, such as `batch_size`).

The main use case of this is suppose we have a list of benchmarks

```
benchmarks: list[DeepEvalBaseBenchmark] = [TruthfulQA(), SQuAD()]
kwargs = {"batch_size": 3}
chosen_benchmark_idx: int = ...
model = ...

benchmarks[chosen_benchmark_idx].evaluate(model, **kwargs)
```

This will allow us to run `SQuAD`, which does **NOT** have a `batch_size` argument without crashing, and if we choose to run `TruthfulQA`, then the `batch_size` will be parsed in the function.

### DIFF
--- a/deepeval/benchmarks/arc/arc.py
+++ b/deepeval/benchmarks/arc/arc.py
@@ -48,7 +48,7 @@ class ARC(DeepEvalBaseBenchmark):
         else:
             self.confinement_instructions = confinement_instructions
 
-    def evaluate(self, model: DeepEvalBaseLLM) -> Dict:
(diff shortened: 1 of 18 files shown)


penguine-ip:
@trevor-inflection looks great, thanks!

PR #1962, opened 2025-08-22
Aggregate results are getting duplicated in results.

john-lemmon-lime:
We were duplicating the aggregate results output for each test case. Now it's done 1 time at the end.


penguine-ip:
Thanks @john-lemmon-lime!

PR #1964, opened 2025-08-22
Test Run ID in Evaluate

john-lemmon-lime:
Specifically I want to generate a comparison link so I need the test runs of a control & treatment run. Right now I can only get the links and I'd rather not parse it out of the link.

### DIFF
--- a/deepeval/evaluate/evaluate.py
+++ b/deepeval/evaluate/evaluate.py
@@ -275,5 +275,7 @@ def evaluate(
         run_duration, display_table=False
     )
     return EvaluationResult(
-        test_results=test_results, confident_link=confident_link, test_run_id=global_test_run_manager.get_test_run().identifier
+        test_results=test_results,
+        confident_link=confident_link,
+        test_run_id=global_test_run_manager.get_test_run().identifier,
     )
--- a/deepeval/evaluate/evaluate.py
+++ b/deepeval/evaluate/evaluate.py
@@ -275,5 +275,5 @@ def evaluate(
         run_duration, display_table=False
     )
     return EvaluationResult(
-        test_results=test_results, confident_link=confident_link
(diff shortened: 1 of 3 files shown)


deepeval/test_run/test_run.py
```diff
@@ -755,6 +759,9 @@ def post_test_run(self, test_run: TestRun) -> Optional[str]:
         res = TestRunHttpResponse(
             id=data["id"],
         )
+        test_run.identifier = res.id
+        test_run.confident_ai_link = link
+        self.save_test_run(self.temp_file_path)
```
john-lemmon-lime, line 764: When we write to the cache the test run is refetched so I need to save it with the changes to the identifier and confident ai link.

I didn't need to include the confident AI link but it made sense to me to do so because it's tied so much to the test run itself.

deepeval/test_run/test_run.py
```diff
@@ -732,16 +733,19 @@ def post_test_run(self, test_run: TestRun) -> Optional[str]:
         ####################
         ### POST REQUEST ###
         ####################
+        # Create a copy of test_run to avoid modifying the original as we batch cases
+        test_run_copy = test_run.model_copy(deep=True)
```
john-lemmon-lime, line 737: Because we're saving the test run below I didn't want to save the changes around limiting the test cases to an initial batch. So I make a deep copy instead.

deepeval/evaluate/evaluate.py
```diff
@@ -275,5 +275,7 @@ def evaluate(
         run_duration, display_table=False
     )
     return EvaluationResult(
-        test_results=test_results, confident_link=confident_link
+        test_results=test_results,
+        confident_link=confident_link,
+        test_run_id=global_test_run_manager.get_test_run().identifier,
```
john-lemmon-lime, line 280: There's a test run object above but we recreate the object from a written cache when we fetch the test run from the global_test_run_manager. So I need the updated reference object to get the test run id.

PR #1965, opened 2025-08-22
GPT 5 Completion Params

john-lemmon-lime:
GPT 5 lets us control the reasoning effort but it's a new param. I don't want to make a new PR to DeepEval each time to support new params so this lets me supply it in a dict. I really want to supply it as regular kwargs but those are already sent to the OpenAIClient so I have to keep these as separate.

So now I can specify the reasoning_effort like this.
`gpt_5_model = GPTModel(model="gpt-5-mini", temperature=1, reasoning_effort="high")`


penguine-ip:
Hey @john-lemmon-lime this looks great, thank you so much!

PR #1969, opened 2025-08-23
Fix AzureOpenAIModel: add **self.generation_kwargs to API calls to fix LengthFinishReasonError

### LINKED ISSUE #1958: LengthFinishReasonError AzureOpenAI API
When evaluating multiple testcases, the error "LengthFinishReasonError" arise. The model that I used was GPT-4o in AzureOpenAI API. I tried to set `max_tokens` in the `deepeval.models.AzureOpenAIModel` but it errors and says `unexpected keyword "max_tokens"`. Is there any work around to solve this issue?

Image

Thanks in advance.

karankulshrestha:
## Problem
Fixes #1958 - LengthFinishReasonError in AzureOpenAIModel

## Solution
Added **self.kwargs to both generate() and a_generate() methods
so that parameters like max_tokens are properly passed to the Azure OpenAI API.

## Changes
- Modified generate() method to pass **self.kwargs
- Modified a_generate() method to pass **self.kwargs

## Why This Fixes the Issue
The LengthFinishReasonError occurs when Azure OpenAI API doesn't receive
proper parameters like max_tokens. By adding **self.kwargs, all parameters
passed to the model constructor are now properly forwarded to the API calls.

### DIFF
--- a/deepeval/models/llms/azure_model.py
+++ b/deepeval/models/llms/azure_model.py
@@ -33,6 +33,7 @@ class AzureOpenAIModel(DeepEvalBaseLLM):
         openai_api_version: Optional[str] = None,
         azure_endpoint: Optional[str] = None,
         temperature: float = 0,
+        generation_kwargs: Optional[Dict] = None, 
         **kwargs,
     ):
         # fetch Azure deployment parameters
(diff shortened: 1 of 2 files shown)


penguine-ip:
Hey @karankulshrestha actually can we name it `generation_kwargs` and make it a dictionary? Because we want to standardize kwargs to model initialization only.

penguine-ip:
@karankulshrestha something like this PR would be great: https://github.com/confident-ai/deepeval/pull/1965

karankulshrestha:
sure

penguine-ip:
@karankulshrestha looks great, thank you!

PR #1972, opened 2025-08-25
LiteLLMModel - max retry and proxy environment variables

hannex:
Max retry - current implementation has not stop condition, so in case of error it will be executed indefinitely.
I've set max retry to 6 - I assume you will change it, based on your experience.

Additionally, I've added support of LiteLLM proxy environment variables.

Finally, I moved retry related values to class level variables, to make future change more simple.


penguine-ip:
@hannex it looks amazing, appreeciate it!

PR #1975, opened 2025-08-25
Change benchmarks to use pydantic classes for evaluate results

### LINKED ISSUE #1948: Make benchmark results pydantic models instead of raw dicts
**Describe the solution you'd like**
Currently every benchmark returns a raw dictionary with arbitrary keys which makes it so there is very little uniformity across benchmarks. Ideally the result would be a pydantic model so that the user knows what to expect in the response. They don't necessarily need to be the same model across benchmarks, but just having a model in general would significantly improve type safety for the project.

It could also be a new API, such as `evaluate_typed` to maintain backwards compatability

**Describe alternatives you've considered**

The alternate way to get around this is to just make local pydantic models for classes that wrap DeepEval benchmarks, but this is a bit cumbersome, especially if the DeepEval benchmark ever changes

Happy to help implement this if you agree that it is beneficial.

trevor-inflection:
Adds strongly typed pydantic interfaces as the return values for all benchmarks to resolve #1948

Currently the `evaluate` function is annotated as `... -> Dict`, but in reality most of the benchmarks return a float. In the case of the `IFEval` it returns a dict, so this PR uses inheritance with pydantic models to support returning the base `overall_accuracy` based interface to the client. However, if they choose to specialize their response into `IFEvalResult` then they have the ability to access the `IFEval` specific fields.

I have manually tested both `IFEval` and random other benchmarks and verified that the code functions properly

An example block that can use this inheritance is something like

```
    def evaluate(
        self, model: BaseRESTModel, logger: BenchmarkLogger | None = None, **kwargs
    ) -> BenchmarkResult | None:
        results: DeepEvalBaseBenchmarkResult = self.benchmark.evaluate(model, **kwargs)
        bm_result = BenchmarkResult(
            overall_score=results.overall_accuracy, individual_scores={}, name=self.get_name()
        )
        return bm_result
```
where `BenchmarkResult` is another pydantic model. Essentially, these changes ensures that the result of each benchmark follows a specified format and also supports serializing to dict/json if the user needs that (hence why a Pydantic model is chosen over a raw dataclass).

It is worth noting that **this is a breaking change**, if that is undesirable then I can convert this PR to have new functions called `evaluate_typed` or something similar. Though in my opinion, it is better for the project long term to enforce every benchmark to only use typed models and warn users in release notes of this change (this could be something done later down the line).

This PR also pins datasets to `<4.0.0` as many benchmarks are reliant on data loader scripts which were deprecated in 4.0.0 causing issues.


penguine-ip:
@trevor-inflection thank you!

PR #1977, opened 2025-08-26
Fix benchmarks for Datasets 4.0.0

khannurien:
As per https://github.com/huggingface/datasets/pull/7592, `trust_remote_code` is no longer supported -- datasets which relied on custom initialization scripts cannot be used anymore.

This pull request:
- removes the boolean from `load_dataset` calls;
- updates MMLU to use an up-to-date Parquet dataset (along with required logic changes);
- updates MathQA to use an up-to-date Parquet dataset.

Please get back to me if further changes are necessary.


penguine-ip:
Hey @khannurien looking good :)

PR #1978, opened 2025-08-26
pydanticai new interface

spike-spiegel-21:


deepeval/integrations/pydantic_ai/agent.py
```diff
@@ -21,14 +41,218 @@ class PydanticAIAgent(Agent):
...
+                run_span.set_attribute("confident.agent.name", name)
+                run_span.set_attribute("confident.agent.input", input)
+                run_span.set_attribute("confident.agent.output", output)
+
+                # fallback for input and output not being set
+                run_span.set_attribute("confident.span.input", input)
+                run_span.set_attribute("confident.span.output", output)
+                
+                if metric_collection: # flattened argument to be replaced
+                    print("setting metric_collection from metric_collection", metric_collection)
```
kritinv: remove print statement

deepeval/integrations/pydantic_ai/agent.py
```diff
@@ -21,14 +41,218 @@ class PydanticAIAgent(Agent):
...
+                    run_span.set_attribute("confident.span.metric_collection", metric_collection)
+                elif self.metric_collection: # to be deprecated
+                    print("setting metric_collection from self.metric_collection", self.metric_collection)
```
kritinv: Remove since it is to be deprecated anyways, if someone is one the old version they can still use it, but they would have to change if they are on the new version.

deepeval/integrations/pydantic_ai/agent.py
```diff
@@ -21,14 +41,218 @@ class PydanticAIAgent(Agent):
...
+                    run_span.set_attribute("confident.span.metric_collection", self.metric_collection)
+
+                # set trace attributes (to be deprecated)
```
kritinv: We can just remove this as well, if we are no longer going with this interface.

deepeval/integrations/pydantic_ai/agent.py
```diff
@@ -21,14 +41,218 @@ class PydanticAIAgent(Agent):
...
+                            metrics=self.metrics,
+                        )
+                    )
+
+            return result
+        
+        # Replace the method only for this instance
+        self.run = patched_run
+
+    def _patch_run_method_sync(self):
```
kritinv, line 195: I don't see the same level of support for this method.  This span creation `with tracer.start_as_current_span("agent")`, as well as support for running local evals. Basically, what is presented above.

spike-spiegel-21: this is because `run_sync` eventually calls the above method in the `async run` in the pydantic code, so ultimately we are doing fine. It is tested in `test_docs`

kritinv:
Btw, for all the type conversions, is there any chance we can reuse functions from the OTEL files, the reason being there is already conversion and typing logic there, so don't want to duplicate too many things.

PR #1981, opened 2025-08-27
fix doc import error in example getting-started-llm-arena.mdx

raphaeluzan:
some import examples are wrong, i fixed it


penguine-ip:
@raphaeluzan tysm!

PR #1984, opened 2025-08-28
synthesizer fix

spike-spiegel-21:


penguine-ip:
@spike-spiegel-21 actually i think reset should always be true, i think it was just left out by mistake, so we can just remove the reset parameter all together and reset at all times

PR #1991, opened 2025-08-29
Added dataset 'save_as' support for ConversationalGoldens 

A-Vamshi:
Supports both save and load in csv and json format


penguine-ip:
@A-Vamshi can we move this into test_core instead of test_confident as these are unrelated to Confident Ai. Also we should create a json of goldens and conversational goldens instead, and csvs instead

A faulthy saving can lead to faulty loading, similarly if both methods are faulty it can mess up the unit test and make it look fine when it is actually failing. Please separate the tests

deepeval/dataset/utils.py
```diff
@@ -51,6 +51,26 @@ def convert_goldens_to_test_cases(
...
 
 
+def convert_convo_test_cases_to_convo_goldens(
+    test_cases: List[ConversationalTestCase],
+) -> List[ConversationalGolden]:
+    goldens = []
+    for test_case in test_cases:
+        if not test_case.scenario:
+            raise ValueError(
+                "Please provide a scenario in the test cases to convert them to goldens. Or toggle 'include_test_cases' to False"
```
penguine-ip: This makes the util function unusable for other future use cases, please change the error message

PR #1995, opened 2025-09-01
public dataset for cookbooks

spike-spiegel-21:


deepeval/dataset/dataset.py
```diff
@@ -659,6 +660,7 @@ def pull(
                     params={
                         "alias": alias,
                         "finalized": str(finalized).lower(),
+                        "isPublic": str(public).lower()
```
penguine-ip: just use `public` here is good

PR #2004, opened 2025-09-03
Use schema whenever we can for geval metrics

john-lemmon-lime:
I was running GEval metrics with `gpt-5-mini` and would randomly hit JSON errors. But `gpt-5-mini` is setup for structured outputs, which should pretty much never hit JSON parsing errors.

Turns out we're not giving the schema when running any native_models, causing it to not run in structured output mode.

I tested this with OpenAI models but I don't have the other native models setup to test against as well. It does look like the typehints aren't correct and the method can return a str or the pydantic model depending on how it's implemented. So I tried to handle that case.

### DIFF
--- a/deepeval/metrics/g_eval/g_eval.py
+++ b/deepeval/metrics/g_eval/g_eval.py
@@ -170,10 +170,14 @@ class GEval(BaseMetric):
             criteria=self.criteria, parameters=g_eval_params_str
         )
         if self.using_native_model:
-            res, cost = await self.model.a_generate(prompt)
+            res, cost = await self.model.a_generate(prompt, schema=Steps)
             self.evaluation_cost += cost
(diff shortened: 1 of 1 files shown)


penguine-ip:
Thanks @john-lemmon-lime ! we're in the middle of a release cycle for introducing more tests to our package atm so this will take a week or so before its merged, same for the other PR we have open

trevor-cai:
Thanks a lot for this contribution, @john-lemmon-lime! Structured output support is definitely something we want to improve, and your PR helped surface that gap clearly. In fact, this is on my road-map and I'm planning to address this in a broader refactor that will allow models to opt in to structured outputs consistently across all metrics. Closing this in favor of that upcoming work, but we really appreciate you raising the issue and pushing this forward.

Issue #2024, opened 2025-09-10
ToolCorrectnessMetric incorrectly returns 0.0 score when both expected_tools and tools_called are empty lists


qwchua:
## Describe the bug
The `ToolCorrectnessMetric` incorrectly returns a score of 0.0 when both `expected_tools` and `tools_called` are empty lists, instead of the expected perfect score of 1.0. This happens when evaluating test cases where no tools are expected to be called and no tools were actually called.

## To Reproduce
1. Create a test case with `expected_tools=[]` (empty list)
2. Run an agent that doesn't call any tools, resulting in `tools_called=[]` (empty list)  
3. Evaluate using `ToolCorrectnessMetric` with default parameters (non-exact matching mode)
4. Observe that the score is 0.0 instead of 1.0

## Expected behavior
When both `expected_tools` and `tools_called` are empty lists, the metric should return a perfect score of 1.0, indicating that the agent correctly avoided calling any tools when none were expected.

## Code Example
```python
from deepeval.test_case import LLMTestCase
from deepeval.metrics import ToolCorrectnessMetric

# Test case where no tools should be called
test_case = LLMTestCase(
    input="What is 2+2?",
    actual_output="4",
    expected_output="4",
    tools_called=[],  # No tools called
    expected_tools=[]  # No tools expected
)

metric = ToolCorrectnessMetric()
score = metric.measure(test_case)
print(f"Score: {score}")  # Currently prints 0.0, should print 1.0
```

## Root Cause
The bug is in the `_calculate_non_exact_match_score()` method around line 225-229:

```python
return (
    total_score / len(self.expected_tools)
    if self.expected_tools  # This evaluates to False for empty list
    else 0.0  # Bug: should handle the case where both lists are empty
)
```

When `expected_tools` is an empty list, `len(self.expected_tools)` is 0, so the condition `if self.expected_tools` evaluates to `False`, causing the method to return `0.0`. However, when both lists are empty, this should be considered a perfect match and return `1.0`.

## Suggested Fix
Replace the problematic return statement with:

```python
if not self.expected_tools:
    # If no tools were expected, check if no tools were called
    return 1.0 if not self.tools_called else 0.0
else:
    return total_score / len(self.expected_tools)
```

## Workaround
Currently, the issue can be worked around by using `should_exact_match=True` parameter, which has the correct logic for handling empty lists:

```python
metric = ToolCorrectnessMetric(should_exact_match=True)
```

## Environment
- OS: macOS
[...]

PR #2027, opened 2025-09-10
Fixed empty tools score

udaykiran2427:
this fixes the issue #2024


penguine-ip:
@udaykiran2427 thank you!

PR #2030, opened 2025-09-11
Document Mistral model integration in custom LLMs

leonardocrociani:
Added instructions for using Mistral models with the instructor package.

### DIFF
--- a/docs/guides/guides-using-custom-llms.mdx
+++ b/docs/guides/guides-using-custom-llms.mdx
@@ -697,7 +697,7 @@ pip install instructor
 ```python
 from pydantic import BaseModel
 from anthropic import Anthropic
-
+import instructor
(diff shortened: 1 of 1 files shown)


penguine-ip:
@leonardocrociani hey thank you! we already have a mistral example here via hugging face: https://deepeval.com/guides/guides-using-custom-llms#mistral-7b-instruct-v03-through-transformers

leonardocrociani:
@penguine-ip yep, it's through transformers. The one I wrote uses remote APIs.

penguine-ip:
@leonardocrociani i think we can include the remote api version using a dropdown at the bottom of the transfers example, like this:

```
<details>

<summary>See API Example...</summary>

[example]

</details>
```

Otherwise we won't be merging to have two of the same provider

PR #2044, opened 2025-09-13
portkey tracing 

spike-spiegel-21:


deepeval/integrations/portkey/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .patcher import instrument
+
+__all__ = ["instrument"]
```
penguine-ip: Need to standardize the name

spike-spiegel-21: We will be using types for all the new integrations going forward. Please check the docs: https://github.com/confident-ai/confident-docs/pull/11

spike-spiegel-21:
> Are there any tool calling available as well in portkey?

We are logging the tool calls in the response in the LLM span.

deepeval/integrations/portkey/__init__.py
penguine-ip: @spike-spiegel-21 is the idea that users will not be calling instrument? If that is the case i think we shouldn't include this as an export

PR #2060, opened 2025-09-16
Pydantic AI additional changes after docs

spike-spiegel-21:
- Also, sanitizes for RunContext

### DIFF
--- a/deepeval/integrations/pydantic_ai/patcher.py
+++ b/deepeval/integrations/pydantic_ai/patcher.py
@@ -133,16 +133,17 @@ def _patch_agent_init():
         result = original_init(*args, **kwargs)
         _patch_llm_model(args[0]._model, llm_metric_collection, llm_metrics, llm_prompt)  # runtime patch of the model
         _patch_agent_run(args[0], agent_metric_collection, agent_metrics)
-        # _patch_agent_run_sync(args[0], agent_metric_collection, agent_metrics)
+        _patch_agent_run_sync(args[0], agent_metric_collection, agent_metrics)
         return result
 
     Agent.__init__ = wrapper
 
 def _patch_agent_run_sync(
+    agent: Agent,
     agent_metric_collection: Optional[str] = None,
     agent_metrics: Optional[List[BaseMetric]] = None,
 ):
-    original_run_sync = Agent.run_sync
+    original_run_sync = agent.run_sync
     
     @functools.wraps(original_run_sync)
     def wrapper(
@@ -197,7 +198,7 @@ def _patch_agent_run_sync(
             )
             
         return result
-    Agent.run_sync = wrapper
+    agent.run_sync = wrapper
 
 
 def _patch_agent_run(
--- a/deepeval/integrations/pydantic_ai/patcher.py
+++ b/deepeval/integrations/pydantic_ai/patcher.py
@@ -35,6 +35,51 @@ except:
     pydantic_ai_installed = True
 
 RUN_SYNC_METHOD = False
+_INSTRUMENTED = False
+
+
+def instrument(otel: Optional[bool] = False, api_key: Optional[str] = None):
+    global _INSTRUMENTED
+    if api_key:
+        deepeval.login(api_key)
+
+    api_key = get_confident_api_key()
+
+    if not api_key:
+        raise ValueError("No api key provided.")
+
+    if otel:
+        instrument_pydantic_ai(api_key)
+    else:
+        with capture_tracing_integration("pydantic_ai"):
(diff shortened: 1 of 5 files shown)


deepeval/integrations/pydantic_ai/patcher.py
```diff
@@ -26,11 +26,13 @@
         ToolReturnPart,
         UserPromptPart,
     )
+    from pydantic_ai._run_context import RunContext
 
     pydantic_ai_installed = True
 except:
     pydantic_ai_installed = True
 
+RUN_SYNC_METHOD = False
```
kritinv: I'm not sure that this is the best, because if we have 2 agent calls in the same file, or even directory, we'll be overwriting this. We probably also need context management for this.

deepeval/integrations/pydantic_ai/patcher.py
```diff
@@ -130,10 +135,67 @@ def wrapper(
...
+
+        with Observer(
+            span_type="agent",
+            func_name="Agent",
+            function_kwargs={"input": args[1]},
+            metrics=agent_metrics,
+            metric_collection=agent_metric_collection,
+        ) as observer:
+            
+            global RUN_SYNC_METHOD
```
kritinv: need context management as per above

deepeval/integrations/pydantic_ai/patcher.py
```diff
@@ -130,10 +135,67 @@ def wrapper(
...
+        metadata: Optional[dict] = None,
+        thread_id: Optional[str] = None,
+        user_id: Optional[str] = None, 
+        **kwargs
+    ):
+
+        with Observer(
+            span_type="agent",
+            func_name="Agent",
+            function_kwargs={"input": args[1]},
```
kritinv: is this always the case? what is the user does run_sync(..., input="...") something like this

deepeval/integrations/pydantic_ai/patcher.py
```diff
@@ -144,21 +206,21 @@ def _patch_agent_run(
...
+        tags: Optional[List[str]] = None,
+        metadata: Optional[dict] = None,
+        thread_id: Optional[str] = None,
+        user_id: Optional[str] = None,
         **kwargs
     ):
         with Observer(
-            span_type="agent",
-            func_name="Agent",
+            span_type="agent" if not RUN_SYNC_METHOD else "custom",
```
kritinv: context mangement ^

deepeval/integrations/pydantic_ai/patcher.py
```diff
@@ -374,3 +436,22 @@ def _extract_tools_called(result: AgentRunResult) -> List[ToolCall]:
                         tool_calls.append(tool_call)
 
     return tool_calls
+
+def _sanitize_run_context(value):
```
kritinv: we can probably put this into its own file utils.py tbh

deepeval/integrations/pydantic_ai/patcher.py
spike-spiegel-21: We won't be overriting it, because the lifecycle for this variable is only at the start and stop of `run_sync`. And they wont be running parallely since they are synchronus. `RUN_SYNC_METHOD` is meant to revert back to false in the `finally` block.

deepeval/integrations/pydantic_ai/patcher.py
spike-spiegel-21: same comment

deepeval/integrations/pydantic_ai/patcher.py
spike-spiegel-21: same comment

deepeval/integrations/pydantic_ai/patcher.py
spike-spiegel-21: never mind, I introduced context variable

deepeval/integrations/pydantic_ai/patcher.py
```diff
@@ -26,11 +28,58 @@
...
     pydantic_ai_installed = True
 except:
     pydantic_ai_installed = True
 
+_IN_RUN_SYNC = ContextVar("deepeval_in_run_sync", default=False)
+_INSTRUMENTED = False
+
+def instrument(otel: Optional[bool] = False, api_key: Optional[str] = None):
+    global _INSTRUMENTED
+    if api_key:
```
penguine-ip, line 43: No change needed, but wanted to clarify on some login flows with @BloggerBust tomorrow on this

penguine-ip (approved):

Issue #2056, opened 2025-09-16
`add_goldens_from_json_file` method missing parameters for several Golden/ConversationalGolden fields


qwchua:
**Describe the bug**
  The `add_goldens_from_json_file` method in `EvaluationDataset` does not provide mapping parameters for all fields available in the `Golden` and `ConversationalGolden` classes. This makes it impossible to fully populate Golden objects from JSON files containing these fields.

  **To Reproduce**
  1. Create a JSON file with Golden data including fields like `comments`, `name`, or `custom_column_key_values`
  2. Use `dataset.add_goldens_from_json_file(file_path="data.json")` to load the data
  3. Check the created Golden objects
  4. The `comments`, `name`, and `custom_column_key_values` fields will be None/empty despite being present in the JSON

  **Expected behavior**
  The `add_goldens_from_json_file` method should accept optional key mapping parameters for ALL fields available in the Golden/ConversationalGolden classes, similar to how it currently handles `input_key_name`, `expected_output_key_name`,
  etc.

  Missing mapping parameters for `Golden`:
  - `comments_key_name`
  - `name_key_name`
  - `custom_column_key_values_key_name`

  Missing mapping parameters for `ConversationalGolden`:
  - `comments_key_name`
  - `name_key_name`
  - `custom_column_key_values_key_name`

**Suggested Fix**
```python
  def add_goldens_from_json_file(
      self,
      file_path: str,
      input_key_name: str = "input",
      actual_output_key_name: Optional[str] = "actual_output",
      expected_output_key_name: Optional[str] = "expected_output",
      context_key_name: Optional[str] = "context",
      retrieval_context_key_name: Optional[str] = "retrieval_context",
      tools_called_key_name: Optional[str] = "tools_called",
      expected_tools_key_name: Optional[str] = "expected_tools",
      source_file_key_name: Optional[str] = "source_file",
      additional_metadata_key_name: Optional[str] = "additional_metadata",
      comments_key_name: Optional[str] = "comments",  # NEW
      name_key_name: Optional[str] = "name",  # NEW
      custom_column_key_values_key_name: Optional[str] = "custom_column_key_values",  # NEW
      scenario_key_name: Optional[str] = "scenario",
      turns_key_name: Optional[str] = "turns",
      expected_outcome_key_name: Optional[str] = "expected_outcome",
      user_description_key_name: Optional[str] = "user_description",
      encoding_type: str = "utf-8",
  ):
      try:
[...]

PR #2066, opened 2025-09-18
Add name and comments to save_as and loading goldens

A-Vamshi:
Fixes issue mentioned in #2056


penguine-ip:
@A-Vamshi thank you! any chance we can add a test too

PR #2067, opened 2025-09-18
Fixed calculation of `pii_leakage` metric score

Sai-Suraj-27:
Based on this calculation formula, if `strict_mode=True` then `threshold must be = 1` and **score should be returned 0** if value of score is less than threshold (1 incase strict_mode=True)

image


penguine-ip:
@Sai-Suraj-27 thank you!

PR #2070, opened 2025-09-19
Fixed calculation of `NonAdvice` Metric score

Sai-Suraj-27:
Based on this calculation formula, if strict_mode=True then threshold must be = 1 and score should be returned 0 if value of score is less than threshold (1 incase strict_mode=True)

image

Missed changing [this](https://github.com/confident-ai/deepeval/blob/d42c4342b4c77c2b8d2921c0e5dec38c7843335e/deepeval/metrics/pii_leakage/pii_leakage.py#L38) in previous [PR](https://github.com/confident-ai/deepeval/pull/2067)


penguine-ip:
@Sai-Suraj-27 thank you!

PR #2077, opened 2025-09-20
Fixed sidebar headings not visible clearly in darkmode

Sai-Suraj-27:
Before

image

After

image


penguine-ip:
@Sai-Suraj-27 thank you!

PR #2081, opened 2025-09-21
Feature/bedrock structured outputs

deepakdalakoti:
This PR adds support for structured outputs in `AmazonBedrockModel`. Main changes are described below:

1. Adds an optional parameter to `AmazonBedrockModel` for structured output. When set as true, uses tool calling to enforce structured outputs. Added functionality to convert a pydantic output schema (in a new `amazon_bedrock_utils.py` file) to a Bedrock Converse API compatible tool specification
2. Removed hardcoded `max_tokens` and `top_p` from `generation_kwargs` for bedrock models. These default values can cause tests to artifically fail. For example, when the test output exceeds `max_tokens`, the generation is stopped which causes the schema parsing to fail and thus the test fails. Setting these values should be upto the user. If no values are provided, bedrock uses the default values for the model and max_tokens are set to maximum tokens allowed for a particular model
3. Updated documentation

### DIFF
--- a/deepeval/models/llms/amazon_bedrock_model.py
+++ b/deepeval/models/llms/amazon_bedrock_model.py
@@ -10,6 +10,9 @@ from deepeval.models.retry_policy import (
 )
 from deepeval.models import DeepEvalBaseLLM
 from deepeval.models.llms.utils import trim_and_load_json
+from deepeval.models.llms.amazon_bedrock_utils import (
+    convert_pydantic_to_tool_spec,
+)
 from deepeval.constants import ProviderSlug as PS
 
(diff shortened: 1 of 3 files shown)


trevor-cai:
Thanks a lot for this contribution, @deepakdalakoti! Structured outputs for Bedrock is definitely something we want to support, and your PR highlights both the approach and some important cleanup (like removing hardcoded generation params). We’re currently working on a broader refactor to make structured outputs opt-in across all models and metrics, and Bedrock will be included in that effort. Closing this for now so the changes can be integrated as part of that work, but I really appreciate you surfacing this and giving us a solid starting point.

PR #2083, opened 2025-09-22
Fixed wrong argument in `set_local_model_env` function call

Sai-Suraj-27:


trevor-cai:
Thank you @Sai-Suraj-27

PR #2091, opened 2025-09-24
feat(utils): add context task scheduler, remove create_task override

BloggerBust:
- introduce `set_task_scheduler` / `reset_task_scheduler` using ContextVar
- update `coerce_to_task` to use the installed scheduler
- bind coroutine/awaitable scheduling to the running loop

refactor(execute): route all task creation through coerce_to_task and install per-loop schedulers

- replace scattered calls to asyncio.create_task with coerce_to_task across async paths
- in agentic loop path, install scheduler that wraps via semaphore, names tasks, and attaches done callbacks
- in _a_evaluate_traces, install a lightweight scheduler (naming only); keep existing semaphore wrapper
- remove global create_task override and use context scheduler instead
- ensure gather/cancel/drain only touches tasks created on the owning loop

### DIFF
--- a/deepeval/dataset/utils.py
+++ b/deepeval/dataset/utils.py
@@ -212,12 +212,14 @@ def coerce_to_task(obj: Any) -> asyncio.Future[Any]:
         # type: ignore[return-value]  # it is an awaitable, gather accepts it
         return obj
 
-    loop = asyncio.get_running_loop()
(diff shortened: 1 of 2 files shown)


BloggerBust:
Going to close this. If we descide we want to move in this direction, we can re-visit it then.

PR #2099, opened 2025-09-25
Refactored KeyFileHandler class to reduce redundant file I/O and JSON parsing

Sai-Suraj-27:
- Added `fetch_multiple_keys()` method to KeyFileHandler class to retrieve multiple key values in a single call.
- Reduces redundant file I/O and JSON parsing compared to multiple `fetch_data()` calls, simplifies model initialization.

### DIFF
--- a/deepeval/models/llms/gemini_model.py
+++ b/deepeval/models/llms/gemini_model.py
@@ -56,25 +56,22 @@ class GeminiModel(DeepEvalBaseLLM):
         generation_kwargs: Optional[Dict] = None,
         **kwargs,
     ):
-        model_name = (
-            model_name
-            or KEY_FILE_HANDLER.fetch_data(ModelKeyValues.GEMINI_MODEL_NAME)
-            or default_gemini_model
+        
+        values = KEY_FILE_HANDLER.fetch_multiple_keys(
+            ModelKeyValues.GEMINI_MODEL_NAME,
+            ModelKeyValues.GOOGLE_API_KEY,
+            ModelKeyValues.GOOGLE_CLOUD_PROJECT,
+            ModelKeyValues.GOOGLE_CLOUD_LOCATION,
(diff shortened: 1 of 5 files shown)


BloggerBust:
Hey @Sai-Suraj-27, 

Thank you for the refactor! Your refactor of the key file handler is a robust improvement for legacy reads. However, we are actively moving runtime reads to `Settings` and away from the legacy `.deepeval` JSON keystore. The CLI already persists only to dotenv for secrets, and our models/config should read via `get_settings` for everything (both secrets and non-secrets alike).

**A few alignment notes before we can merge:**

1. **Please drop `fetch_multiple_keys()`**
   We are deprecating the `.deepeval` JSON keystore and not adding new API surface to it. Batch reads optimize the wrong layer now that secrets persist to dotenv via the CLI. Keeping this method would encourage new call sites against a store we are moving away from.

2. **Keep only the safe JSON load improvements**
   The `_load_data`/single-read refactor and corruption safe handling are welcome as legacy hardening. That is useful for older installs that still have secrets in a `.deepeval` file lying around.

3. **Revert all model file changes**
   Models should read exclusively from `Settings` (`get_settings()`), not from `KeyFileHandler`. The CLI already demonstrates the source-of-truth pattern. If you want to help here, swapping those reads to `get_settings()` would be perfect.

4. **Mark the keyfile handler as legacy via docstring**
   Add a short module/class docstring: “Legacy JSON keystore for backwards compatibility. Do not add new call sites; use `get_settings()` instead.”
   (No runtime warnings needed, we just clear documentation to avoid churn.)

5. **Tests** (minimal):
   * Don't forget to update the test after removing `fetch_multiple_keys`

If you make the above changes, I am happy to approve. This way we keep your nice I/O/parse reduction for legacy reads without increasing coupling to the legacy store, and we stay aligned with the Settings config direction.

PR #2102, opened 2025-09-26
[Mergeable] Prompt Changes

kritinv:


deepeval/test_run/test_run.py
```diff
@@ -71,6 +82,16 @@ class TraceMetricScores(BaseModel):
     base: Dict[str, Dict[str, MetricScores]] = Field(default_factory=dict)
 
 
+class PromptData(BaseModel):
```
penguine-ip, line 85: Do we need to log prompt type here? Seems like this means a prompt can have both text and messages templates

PR #2113, opened 2025-09-30
feat(evaluate,test_run): display conversational turns in multi-turn evals

BloggerBust:
Print and log sorted turns, include role and truncated content, and tools used. Pass turns (`TurnApi`) through `TestResult` for conversational cases. Include focused tests for CLI/file output and turns table.

### Screen shot of rich table formatting:
image

### Screen shot of terminal output:
image

### DIFF
--- a/deepeval/evaluate/utils.py
+++ b/deepeval/evaluate/utils.py
@@ -35,14 +35,13 @@ from deepeval.tracing.utils import (
 )
 
 
-def _is_metric_successful(metric_data) -> bool:
+def _is_metric_successful(metric_data: MetricData) -> bool:
     """
     Robustly determine success for a metric row.
 
     Rationale:
     - If the metric recorded an error, treat as failure.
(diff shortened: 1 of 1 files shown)


deepeval/evaluate/utils.py
```diff
@@ -34,6 +35,30 @@
 )
 
 
+def _is_metric_successful(metric_data) -> bool:
```
penguine-ip: I think we need this more strictly typed `: MetricData`

deepeval/evaluate/utils.py
```diff
@@ -34,6 +35,30 @@
...
+    """
+    Robustly determine success for a metric row.
+
+    Rationale:
+    - If the metric recorded an error, treat as failure.
+    - Try/except guards against custom metrics that may omit or mistyped `success`.
```
penguine-ip: ^doesn't match implementation

PR #2117, opened 2025-10-01
Updated `async_mode=True` by default for Hallucination Metric

Sai-Suraj-27:
[Docs](https://deepeval.com/docs/metrics-hallucination#usage) are already correctly saying default value is True.


penguine-ip:
@Sai-Suraj-27 thanks!

PR #2132, opened 2025-10-03
fix: broken doc code examples

denis-snyk:
I was trying out this library using some of the examples such as the one below.

```python
@pytest.mark.parametrize(
    "test_case",
    dataset,
)
def test_customer_chatbot(test_case: LLMTestCase):
    assert_test(test_case, [AnswerRelevancyMetric()])
```

I was getting this error: `TypeError: 'EvaluationDataset' object is not iterable`

I believe this is because of a mistake in the docs. This PR attempts to fix all occurences of this docs typo.

Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -309,7 +309,7 @@ for golden in dataset.goldens:
 
 @pytest.mark.parametrize(
     "test_case",
-    dataset,
+    dataset.test_cases,
 )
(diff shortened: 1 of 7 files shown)


penguine-ip:
@denis-snyk nice catch, thank you!

PR #2134, opened 2025-10-03
feat(execute): add per-task timeouts to semaphore-guarded awaits

### LINKED ISSUE #2125: feat: Add per-task timeouts to semaphore guarded awaits in execute.py
### Problem
In `deepeval/evaluate/execute.py`, several coroutines are awaited inside `async with semaphore:` blocks without a timeout. If a downstream call stalls (model, user callback, etc.), the await can block forever and the whole run hangs.

### Goal
Guard semaphore-wrapped awaits with `asyncio.wait_for` using our existing config: `DEEPEVAL_PER_TASK_TIMEOUT_SECONDS`

**What to change**
Find patterns like:

```py
async def execute_with_semaphore(func, *args, **kwargs):
    async with semaphore:
        return await func(*args, **kwargs)
```

Replace the inner `await` with a bounded wait:

```py

async def execute_with_semaphore(func, *args, **kwargs):
    async with semaphore:
        try:
            return await asyncio.wait_for(
                func(*args, **kwargs),
                timeout=settings.DEEPEVAL_PER_TASK_TIMEOUT_SECONDS,
            )
        except asyncio.TimeoutError:
            # Optional: add structured logging here with enough context
            # Then surface a friendly error object / re-raise for caller handling
            # see existing `except asyncio.TimeoutError:` block
            raise
```

If the helper takes a coroutine instead of `func,*args`, use the same `wait_for` around the coroutine:

```py
async def execute_callback_with_semaphore(coro):
    async with semaphore:
        return await asyncio.wait_for(
            coro,
            timeout=settings.DEEPEVAL_PER_TASK_TIMEOUT_SECONDS,
        )
```

### Scope
- File: `deepeval/evaluate/execute.py`
- Only semaphore-guarded awaits that do not already have an enclosing timeout.
- **Don’t** change function signatures or settings names.
- **Don’t** add new settings.

**Testing hints**
Create a dummy coroutine that `await asyncio.sleep(9999)` inside one of the affected paths and verify it times out per setting. For local testing, temporarily set `DEEPEVAL_PER_TASK_TIMEOUT_SECONDS=2` via env or settings to speed up.

**Notes**

* It’s okay if some call sites end up “double guarded” (inner per-task plus outer group timeout), the inner wins for per-task cutoff, outer protects the batch.

yujiiroo:
Adds per-task timeouts to all semaphore-guarded awaits in `execute.py` using 
`asyncio.wait_for` and the existing `DEEPEVAL_PER_TASK_TIMEOUT_SECONDS`.

Fixes #2125

### DIFF
--- a/deepeval/evaluate/execute.py
+++ b/deepeval/evaluate/execute.py
@@ -490,14 +490,10 @@ async def a_execute_test_cases(
 
     async def execute_with_semaphore(func: Callable, *args, **kwargs):
         async with semaphore:
-            try:
-                return await asyncio.wait_for(
-                    func(*args, **kwargs),
-                    timeout=settings.DEEPEVAL_PER_TASK_TIMEOUT_SECONDS,
-                )
-            except asyncio.TimeoutError:
-
-                raise
+            return await asyncio.wait_for(
+                func(*args, **kwargs),
+                timeout=_per_task_timeout(),
+            )
 
     global_test_run_cache_manager.disable_write_cache = (
         cache_config.write_cache is False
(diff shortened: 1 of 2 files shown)


trevor-cai:
Hey @yujiiroo,

Thank you for your contribution. Can you please follow [these instructions](https://github.com/confident-ai/deepeval/pull/2141#issuecomment-3374221016) to fix the lint issues in the code that you changed.

Thank you!

yujiiroo:
Alright sure

yujiiroo:
Heyy @trevor-cai, i've fixed the lint issues

Issue #2133, opened 2025-10-03
AmazonBedrockModel fails when using Claude Sonnet 4.5 model due to conflicting temperature and topP parameters


eppeters:
__Describe the bug__

AmazonBedrockModel fails when using Claude Sonnet 4.5 model due to conflicting temperature and topP parameters. The model throws a ValidationException because Claude Sonnet 4.5 cannot accept both `temperature` and `top_p` parameters simultaneously, but AmazonBedrockModel automatically sets both parameters regardless of model compatibility.

__To Reproduce__

1. Create an AmazonBedrockModel instance with Claude Sonnet 4.5:

   ```python
   from deepeval.models import AmazonBedrockModel
   bedrock_model = AmazonBedrockModel(
       model_id="anthropic.claude-sonnet-4-5-20250929-v1:0",
       region_name="us-east-1",
       temperature=0
   )
   ```

2. Attempt to generate content using the model

3. Observe ValidationException from Bedrock API

__Expected behavior__ The AmazonBedrockModel should successfully work with Claude Sonnet 4.5 model by only sending compatible parameters. For models that don't support both temperature and top_p simultaneously, only one parameter should be sent.

__Error Message__

```javascript
botocore.errorfactory.ValidationException: An error occurred (ValidationException) when calling the Converse operation: The model returned the following errors: `temperature` and `top_p` cannot both be specified for this model. Please use only one.
```

__Root Cause Analysis__

In `get_converse_request_body()`, the method sets both:

- `temperature`: from the constructor parameter (defaults to 0)
- `topP`: from generation_kwargs with a default of 0

This causes conflicts for models like Claude Sonnet 4.5 that only accept one of these parameters.

__Desktop (please complete the following information):__

- deepeval Version: 3.5.6

A-Vamshi:
Hey @eppeters, this issue was recently solved. We decided to remove the default `temperature` they are now defaulted to what the amazon bedrock model offers, you can set them using the `generation_kwargs` parameter. Here's the docs: https://deepeval.com/integrations/models/amazon-bedrock#python

Let us know if you run into any other issues.

SalMireles:
Hi, the prompt management platform has both on by default without the option to remove and this same issue surfaces there.

SalMireles:
Hi, the prompt management platform has both on by default without the option to remove and this same issue surfaces there.

PR #2137, opened 2025-10-04
fix/bedrock-model-param-conflict

Vidushi2709:
## Fix Claude Sonnet 4.5 Parameter Conflict in AmazonBedrockModel
Issue -> https://github.com/confident-ai/deepeval/issues/2133

### Problem
AmazonBedrockModel fails when using Claude Sonnet 4.5 due to conflicting `temperature` and `topP` parameters. The model throws a `ValidationException` because Claude Sonnet 4.5 cannot accept both parameters simultaneously, but AmazonBedrockModel automatically sets both regardless of model compatibility.

**Error:**
```
botocore.errorfactory.ValidationException: The model returned the following errors: 
`temperature` and `top_p` cannot both be specified for this model. Please use only one.
```

### Solution
- **Model Detection**: Added case-insensitive detection for Claude Sonnet 4.5 models using `"claude-sonnet-4-5"` in model ID
- **Parameter Logic**: For Claude Sonnet 4.5, only send one parameter:
  - If `temperature > 0`: send only `temperature`
  - If `temperature == 0` and `top_p > 0`: send only `topP`
  - Otherwise: send `temperature` (default)
- **Backward Compatibility**: Other models continue to receive both parameters as before
- **Flexible Input**: Support both `"top_p"` and `"topP"` parameter names

### Changes
- Modified `get_converse_request_body()` in `AmazonBedrockModel`
- Added comprehensive tests for Claude Sonnet 4.5 compatibility
- Maintained existing behavior for all other models

### DIFF
--- a/deepeval/models/llms/amazon_bedrock_model.py
+++ b/deepeval/models/llms/amazon_bedrock_model.py
@@ -14,8 +14,8 @@ from deepeval.constants import ProviderSlug as PS
 
 # check aiobotocore availability
 try:
-    from aiobotocore.session import get_session
-    from botocore.config import Config
+    from aiobotocore.session import get_session  # type: ignore
+    from botocore.config import Config  # type: ignore
 
     aiobotocore_available = True
 except ImportError:
@@ -29,7 +29,8 @@ def _check_aiobotocore_available():
     if not aiobotocore_available:
         raise ImportError(
             "aiobotocore and botocore are required for this functionality. "
-            "Install them via your package manager (e.g. pip install aiobotocore botocore)"
+            "Install them via your package manager "
+            "(e.g. pip install aiobotocore botocore)"
         )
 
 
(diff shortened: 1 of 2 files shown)


penguine-ip:
@Vidushi2709 thanks for the PR! @BloggerBust i'm wondering is there a cleaner way to handle such cases? such as delegating all generation kwargs away to the user

eppeters:
I had considered some possible solutions. Of course you can add special case handling, but if Sonnet 4.5 does this, future models probably will as well, or they'll have other parameter quirks. Plus, maintaining a hardcoded list requires the library to be updated every time a new model comes out that only accepts one of these parameters.

My opinion about how this might be fixed: Right now (on main, not this PR), the class takes in one of these parameters as a function parameter and the other as a key of the `generation_kwargs` parameter. Given the different requirements for different Bedrock models, it probably makes the most sense to take in both of these as `generation_kwargs`. It would be best to then deprecate the special temperature param. For now, you could default to `None` and log that it will be removed in a future version (not sure how deepeval normally handles this)

For models that accept both parameters (most models right now), users may want to default to 0, which is understandable. Perhaps you could have a default generation params dict that is only used if a user doesn't pass in their own dict. Of course this [shouldn't be the parameter's default value - it should be set by the code to avoid issues with mutable types used as parameter defaults](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments).

penguine-ip:
Hey @Vidushi2709 thanks for the detailed explanation. We also agree and have moved the temperature to generation kwargs as well, its out in the latest release and reflected in the docs: https://deepeval.com/integrations/models/amazon-bedrock#python

I'm going to close this PR right now since it is solved, thank you!

PR #2140, opened 2025-10-05
Saving trace in Pydanic AI instrumentator for accessing in the future 

spike-spiegel-21:
This will allow us to access the Trace ID in future to access the annotation. Not the best solutions, but it is very hard to find a way to callback trace ID for a given `run` context in Pydantic AI agent.

### DIFF
--- a/tests/test_integrations/test_pydanticai/simple_agent.py
+++ b/tests/test_integrations/test_pydanticai/simple_agent.py
@@ -29,12 +29,13 @@ agent = Agent(
 
 
 async def execute_simple_agent():
-    result = await agent.run("What are the LLMs?")
-    print("===============Simple agent output:===============")
-    print(result)
+    with TraceContext() as trace:
+        result = await agent.run("What are the LLMs?")
+        print("===============Simple agent output:===============")
+        print(result)
+        print("===============Trace ID:===============")
+        print(trace.get_uuid())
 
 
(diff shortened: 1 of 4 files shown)


tests/test_integrations/test_pydanticai/simple_agent.py
```diff
@@ -27,7 +27,12 @@
...
 
 
-def execute_simple_agent():
-    result = asyncio.run(agent.run("What are the LLMs?"))
+async def execute_simple_agent():
+    result = await agent.run("What are the LLMs?")
     print("===============Simple agent output:===============")
     print(result)
+    print("===============Trace ID:===============")
+    print(confident_instrumentation_settings.get_trace_id())
```
penguine-ip: @spike-spiegel-21 does this still work if `agent.run()` is called somewhere else in between `await agent.run("What are the LLMs?")` and `confident_instrumentation_settings.get_trace_id()`?

penguine-ip:
```python

with ContextManager():
  result = await agent.run("What are the LLMs?")
  print(confident_instrumentation_settings.get_trace_id())

with ContextManager():
  result = await agent.run("What are the LLMs?")
  print(confident_instrumentation_settings.get_trace_id())
```

PR #2151, opened 2025-10-06
Amazon bedrock fix to remove default temperature

A-Vamshi:


penguine-ip:
@A-Vamshi can we please add the docs for this: https://deepeval.com/integrations/models/amazon-bedrock#python

Issue #2164, opened 2025-10-08
MultimodalToolCorrectnessMetric cannot accept MLLMTestCase


EZLOGIC:
**Describe the bug**
`MultimodalToolCorrectnessMetric` raises a `ValueError:` when running the test with `MLLMTestCase`

This happens because the MultimodalToolCorrectnessMetric's method `mesure` uses the `check_llm_test_case_params` function. 

```
def measure(
        self,
        test_case: MLLMTestCase,
        _show_indicator: bool = True,
        _in_component: bool = False,
    ) -> float:
        check_llm_test_case_params(test_case, self._required_params, self)
```

This function rejects all non-`LLMTestCase` objects.

**To Reproduce**
Run the following minimal script:

```
from deepeval.metrics import MultimodalToolCorrectnessMetric
from deepeval.test_case import MLLMTestCase, ToolCall


test_case = MLLMTestCase(
    input=["What's in this image?"],
    actual_output=["The image shows a pair of running shoes."],
    tools_called=[
        ToolCall(name="ImageAnalysis"),
        ToolCall(name="ToolQuery")
    ],
    expected_tools=[
        ToolCall(name="ImageAnalysis")
    ]
)

metric = MultimodalToolCorrectnessMetric(threshold=0.5)
metric.measure(test_case)
```
Observed behavior:
Raising the `ValueError: Unable to evaluate test cases that are not of type 'LLMTestCase' using the non-conversational 'Tool Correctness' metric.`

**Expected behavior**
The metric should successfully evaluate `MLLMTestCase` objects and return a score based on tool correctness.

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser N/A
 - Version 3.6.4

**Additional context**
The issue originates in `check_llm_test_case_params(test_case, self._required_params, self)` within deepeval/metrics/utils.py

```
if isinstance(test_case, LLMTestCase) is False:
        error_str = f"Unable to evaluate test cases that are not of type 'LLMTestCase' using the non-conversational '{metric.__name__}' metric."
        metric.error = error_str
        raise ValueError(error_str)
```
Possible solution
Add a new function in the `utils.py` module. Something like this:

```
def check_mllm_test_case_parameters(
    test_case: MLLMTestCase,
    test_case_params: List[MLLMTestCaseParams],
    metric: Union[BaseMetric, BaseArenaMetric],
):

    if isinstance(test_case, MLLMTestCase) is False:
[...]

PR #2170, opened 2025-10-09
Correct function name to `check_mllm_test_case_params`

Aisha630:
This pull request updates the parameter validation logic in the `multimodal_tool_correctness.py` metric implementation. The main change is to switch from using the LLM-specific test case parameter checker to the MLLM-specific version, ensuring the correct validation for multimodal test cases.

Validation logic update:

* Replaced `check_llm_test_case_params` with `check_mllm_test_case_params` in both the import statement and the `measure` method, so the code now validates parameters specifically for multimodal large language models (MLLMs).

Fixes: #2164

### DIFF
--- a/deepeval/metrics/multimodal_metrics/multimodal_tool_correctness/multimodal_tool_correctness.py
+++ b/deepeval/metrics/multimodal_metrics/multimodal_tool_correctness/multimodal_tool_correctness.py
@@ -47,7 +47,7 @@ class MultimodalToolCorrectnessMetric(BaseMetric):
         _show_indicator: bool = True,
         _in_component: bool = False,
     ) -> float:
-        check_mllm_test_case_params(test_case, self._required_params, self)
+        check_mllm_test_case_params(test_case, self._required_params, None, None, self)
         self.test_case = test_case
         with metric_progress_indicator(
             self, _show_indicator=_show_indicator, _in_component=_in_component


deepeval/metrics/multimodal_metrics/multimodal_tool_correctness/multimodal_tool_correctness.py
```diff
@@ -47,7 +47,7 @@ def measure(
         _show_indicator: bool = True,
         _in_component: bool = False,
     ) -> float:
-        check_llm_test_case_params(test_case, self._required_params, self)
+        check_mllm_test_case_params(test_case, self._required_params, self)
```
EZLOGIC: This fix is incomplete. The function `check_mllm_test_case_params` requires `input_image_count` and `actual_output_image_count` parameters.       
The correct call should be:

check_mllm_test_case_params(test_case, self._required_params, None, None, self)

Otherwise it throws: `TypeError: check_mllm_test_case_params() missing 2 required positional arguments: 'actual_output_image_count' and 'metric'` while trying to test multimodal tool correctness

Aisha630: Ah my bad, I missed that. Fixed it

PR #2198, opened 2025-10-14
Fix typos in metrics-dag documentation

simoneb:


penguine-ip:
@simoneb thank you :)

PR #2215, opened 2025-10-17
Move Hallucination required params to class var like all other metrics

AnuragGowda:
Moved module level required params to be a class level variable, this is the implementation across all other metrics except this one.

Why is this attribute denoted to be "private"? This can pretty useful as it can help with input validation on custom observability frameworks that leverage Deepeval metrics


penguine-ip:
hey @AnuragGowda thank you!

PR #2230, opened 2025-10-22
fix: 2229 json.dumps in ToolCall.__repr__ needs ensure_ascii=False 

danerlt:


penguine-ip:
@danerlt thank you :)

Issue #2231, opened 2025-10-22
Backward compatibility of DEEPTEAM_TELEMETRY_OPT_OUT


agrudkow:
The recent modification to environment variable parsing — where only numeric truthy values like 1 are accepted — introduces a security and compliance risk rather than just a configuration issue.

Previously, many teams used DEEPEVAL_TELEMETRY_ENABLED=YES (a conventional and widely accepted pattern) to explicitly disable telemetry or control outbound data flows in accordance with company policies. After the change, this flag is no longer recognized as truthy, which can silently alter the intended telemetry behavior.

In organizations with strict data governance, network egress controls, or regulatory compliance constraints, such a subtle behavioral change can result in unauthorized outbound communication — a serious issue in environments where telemetry transmission to third-party servers is disallowed.

Moreover, the related deepteam library still uses the YES variant in its configuration, creating inconsistency across dependent components. This inconsistency can easily lead to misconfigurations that are hard to diagnose and error-prone, especially when both libraries are used together in the same ecosystem.

> 🎨 It might have been a good idea to support both YES and 1 for backward compatibility. This could break in some environments , including read-only filesystems 

 _Originally posted by @fisabelle-lmi in [1ff6f8d](https://github.com/confident-ai/deepeval/commit/1ff6f8dd9b0c0903f01f55d05e8c5f3bff057805#r165437477)_

penguine-ip:
@trevor-cai could we somehow support this? something really quick would help...

trevor-cai:
Hi @agrudkow ,

Thank you for raising this and for pointing out your concerns about compliance. DeepEval continues to support flexible boolean values such as 'yes', 'no', '1', '0', and similar representations. What has caused some confusion is that the environment variable `DEEPEVAL_TELEMETRY_ENABLED` was renamed `DEEPEVAL_TELEMETRY_OPT_OUT` and with it the semantics changed.

We have reintroduced support for the legacy flag `DEEPEVAL_TELEMETRY_ENABLED` and mapped it to the current `DEEPEVAL_TELEMETRY_OPT_OUT` with the following precedence rules:

- If both are set, **any OFF signal** wins

   - `DEEPEVAL_TELEMETRY_OPT_OUT` truthy -> telemetry OFF
   - `DEEPEVAL_TELEMETRY_ENABLED` falsy -> telemetry OFF
-  Else, ON signal applies
   * `DEEPEVAL_TELEMETRY_OPT_OUT` falsy -> telemetry ON
[...]

PR #2249, opened 2025-10-27
feat(settings): auto-refresh on env change and unskip test_settings

BloggerBust:
- test(core): add `enable_dotenv` marker and env sandbox
- update dotenv boolean tests to ensure coercsion works as intended
- chore(pyproject): register marker
- ref #2231: use `DEEPEVAL_TELEMETRY_OPT_OUT` (not `DEEPEVAL_TELEMETRY_ENABLED`)


agrudkow:
@trevor-cai wdyt about logging some warning when user uses deprecated env var? It seem like a good idea to give user a hint that this is deprecated and will be removed in a future major releases

penguine-ip:
Hey @agrudkow love the suggestion, was wondering could you quickly make this a new issue? we're merging this now to get  a new deepeval release out asap for this fix...

PR #2264, opened 2025-10-29
Update `bool | None` -> `Optional[bool]`

OwenKephart:
The `bool | None` syntax is not valid in `python3.9`, which is marked as supported in the current [pyproject.toml](https://github.com/confident-ai/deepeval/blob/main/pyproject.toml#L19).

This line (introduced in 3.6.8) causes errors when attempting to use this package with py39.


penguine-ip:
@OwenKephart thank you!

PR #2272, opened 2025-10-30
fix: migrate Pydantic models to use ConfigDict instead of class-based Config

andres-ito-traversal:
Fixes Pydantic V2 deprecation warnings by migrating all models from class-based `Config` to `ConfigDict`.

## Problem

When using deepeval, users encounter multiple deprecation warnings:

## Testing Methodology
I created this test script:
```python
"""Test script to replicate Pydantic deprecation warnings"""
import warnings
warnings.filterwarnings("default", category=DeprecationWarning)

# Import the modules that have the deprecated configs
from deepeval.prompt.api import OutputSchemaField, PromptPushRequest, PromptUpdateRequest
from deepeval.prompt.prompt import CachedPrompt
from deepeval.tracing.types import BaseSpan, Trace
from deepeval.tracing.api import BaseApiSpan, TraceApi
from deepeval.test_run.cache import MetricConfiguration

print("All modules imported successfully")
```

I was able to replicate the deprecation errors. Then I made the proposed fix and re-ran the script and saw no warnings

### DIFF
--- a/deepeval/benchmarks/ifeval/ifeval.py
+++ b/deepeval/benchmarks/ifeval/ifeval.py
@@ -1,8 +1,8 @@
-from pydantic.config import ConfigDict
 from deepeval.benchmarks.base_benchmark import (
     DeepEvalBaseBenchmark,
     DeepEvalBaseBenchmarkResult,
 )
+from deepeval.utils import make_model_config
 from typing import List, Optional, Dict, Any, Tuple
 from tqdm import tqdm
 import re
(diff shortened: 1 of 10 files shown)


penguine-ip:
Hey @andres-ito-traversal is it possible to conditionally do both depending on the pydantic version? I'm assuming this will break things for V1 users...

andres-ito-traversal:
> Hey @andres-ito-traversal is it possible to conditionally do both depending on the pydantic version? I'm assuming this will break things for V1 users...

I suppose you're right @penguine-ip, but I thought this was okay since in your pyproject.toml file the pydantic version the code is using is 2+. I suppose clients of this library could be using an older pydantic version, however. I've made the code changes to make this backwards compatible.

Just another thing, but it looks like there are other places in the repo that use `ConfigDict`. Presumably it's not a problem already. I attached a screenshot

Screenshot 2025-10-31 at 8 13 50 AM

andres-ito-traversal:
@penguine-ip friendly ping here :) Thanks!

penguine-ip:
Hey @andres-ito-traversal yup we had a look internally and seems like pydantic 1 is uncessary, thank you!

PR #2273, opened 2025-10-30
docs(contributing): add 12-month stale issue policy

BloggerBust:
label exclusions: security, roadmap


penguine-ip:
@BloggerBust can we make it so that it doesn't touch issues with any label? :)

trevor-cai:
> @BloggerBust can we make it so that it doesn't touch issues with any label? :)

Done!

PR #2304, opened 2025-11-12
Allow any version of pytest-rerunfailures

bostadynamics:
The following is an LLM's answer to the question "what are the changes in `pytest-rerunfailures` from 12 to latest?".
IMO it is safe to allow any version of said package as they didn't break any apis.

---

Here is a comprehensive summary of all notable changes from version 12.0 to the latest (16.1).

### ❗ Breaking Changes

The breaking changes are **entirely related to dropping support for end-of-life Python and pytest versions**. This is a common practice to reduce maintenance overhead.

Here is the timeline of dropped support:

* **v16.1:** Dropped support for **Python 3.9**.
* **v16.0:** Dropped support for **pytest < 8.0**.
* **v15.0:** Dropped support for **Python 3.8** and **pytest < 7.4**.
* **v14.0:** Dropped support for **Python 3.7** and **pytest < 7.2**.
* **v13.0:** Dropped support for **Python 3.6** and **pytest < 6.0**.
* **v12.0:** (The version you started at) Dropped support for **pytest < 6.2**.

---

### ✨ New Features & Enhancements

Several new command-line options and marker arguments were added to give you more control over reruns.

* **v16.1:**
    * Added `--force-reruns` to globally override the rerun count, ignoring any `@pytest.mark.flaky` settings.
* **v16.0:**
    * The `condition` argument for `@pytest.mark.flaky` can now be a callable (function) or an evaluatable string, which has access to the `error` object.
* **v15.0:**
    * Added `--fail-on-flaky` option. If a test passes only after being rerun, this option will still make the test run fail with a custom exit code (8).
* **v13.0:**
    * Configuration settings (like `reruns` and `reruns_delay`) can now be set in `pytest.ini` or `pyproject.toml` files, not just on the command line.
* **v12.0:**
    * Added `only_rerun` and `rerun_except` arguments to the `@pytest.mark.flaky` marker. This allows you to specify, on a per-test basis, which exception types should or should not trigger a rerun.

---

### 🐞 Bug Fixes & Other Changes

This category includes fixes for crashes, incorrect behavior, and compatibility issues.

* **v16.1:**
    * Changed "localhost" to "127.0.0.1" to avoid hostname resolution issues.
* **v16.0:**
    * Fixed a bug where error attributes were inaccessible when using `pytest-xdist`.
* **v15.1:**
    * Fixed the `--fail-on-flaky` option to only trigger if reruns actually happened.
    * Improved exception filtering logic.
* **v15.0:**
    * Added support for pytest 8.x.
* **v14.0:**
    * Fixed a bug where teardown for non-function-scoped fixtures was missed when using `only_rerun` or `rerun_except`.
* **v13.0:**
    * Fixed a crash when a setup function failed while the `--rerun-except` flag was used.
    * Fixed a crash when `pytest-xdist` was installed but disabled.
* **v12.0:**
    * Fixed an issue where teardown of session, class, and module-scoped fixtures would run multiple times. Now, they run only once after all reruns are complete.
    * Fixed a crash during teardown when the `runtest` protocol hook was in use by another plugin.


penguine-ip:
@bostadynamics thank you!

Issue #1419, opened 2025-03-06
pytest plugin 'plugins' has generic name, triggers telemetry by default.


Ben-grmbl:
Deepeval includes a pytest plugin called ["plugins"](https://github.com/confident-ai/deepeval/blob/80c978386c916b37995d156b8f40721b960844ab/pyproject.toml#L15).

The plugin automatically loads when executing pytest, regardless of whether any tests using deepeval are executed.

The plugin will by default enable telemetry and update-check, and start making outgoing connections, before any pytest fixture or other python code can modify the variables `DEEPEVAL_TELEMETRY_OPT_OUT`, or `DEEPEVAL_UPDATE_WARNING_OPT_OUT`.

This may cause serious problems when executing any tests, even those that make no use of deepeval, in a sensitive or restricted environment.

Determining the root cause of these network requests was quite challenging

The plugin can be disabled by passing `-p no:plugins` to pytest. 

However, it is very hard to find out that this plugin is related to deepeval at all. It is shown when executing `pytest --trace-config`, but only by name, without any reference to deepeval.

**Describe the solution you'd like**

I would like a less generic name, such as `deepeval-pytest-plugin` instead of `plugins`, and some mention in the documentation, that a plugin with this name exists, and how it can be disabled.

The plugin should not start accessing the network, checking for updates, or sending telemetry, unless any deepeval features are actually used in a test.

A way to configure the telemetry behaviour from within the tests, without relying on environment variables from the OS, would be great

max-sixty:
Similarly:
- The plugin also prints stuff to stdout, even when deepeval isn't part of the test. `]Running teardown with pytest sessionfinish...`, and prints some more details about tests that have been skipped
- It makes the tests much much slower. Running a test goes from 1.2 seconds to 3.6 seconds — what's it doing? Again deepeval is not involved with this test (the test passes even with it uninstalled...)

```
________________________________________________________
Executed in    3.65 secs    fish           external
   usr time    1.46 secs    0.04 millis    1.46 secs
   sys time    0.35 secs    1.19 millis    0.35 secs

```

```
________________________________________________________
Executed in    1.26 secs      fish           external
   usr time  537.05 millis    0.05 millis  537.00 millis
   sys time  133.42 millis    1.46 millis  131.96 millis

```

I'm not some privacy or free-software activist — I support companies making money from their open-source efforts, and even think some light telemetry is OK. 

[...]

PR #2308, opened 2025-11-14
Rename plugin entity from "plugins" to "deepeval"

gavmor:
This resolves part of [pytest plugin 'plugins' has generic name](https://github.com/confident-ai/deepeval/issues/1419), a problem which has significantly inconvenienced me.


penguine-ip:
@gavmor thanks for the PR!

PR #2312, opened 2025-11-16
Fix code block formatting in README.md

DevilsAutumn:


penguine-ip:
@DevilsAutumn thank you!

PR #2317, opened 2025-11-20
Fix: NameError in CrewAI integration when optional dependency is missing

RajaMuhammadAwais:


penguine-ip:
Hey @RajaMuhammadAwais thanks for adding this, i'm wondering does this error occur only when you are importing crewai?

PR #2323, opened 2025-11-24
fix: verbose_mode argument and creation of predictions dataframe in HumanEval

dermodmaster:
## HumanEval `verbose_mode` is always true currently
The `verbose_mode` parameter in the constructor of class `HumanEval` is currently unused and will be set to `(False,)` which if casted to bool is `bool((False,)) => True`.

## Column mismatch for building .predictions
Fixes the human_eval Evaluation because of a column mismatch while gathering the prediction results for human_eval.

Column mismatch between:
```
                predictions_row.append(
                    (
                        task.value,
                        golden.input,
                        prediction,
                        golden.expected_output,
                        score,
                    )
                )
```
and
```
            # Create a DataFrame from task_results_data
            # Columns: 'Task', 'Input', 'Prediction', 'Score'
            self.predictions = pd.DataFrame(
                predictions_row,
                columns=[
                    "Task",
                    "Input",
                    "Prediction",
                    "Correct",
                    "Expected Output",
                    "Score",
                ],
            )
```

Occured exception:
```
ValueError: 6 columns passed, passed data had 5 columns
```

### DIFF
--- a/deepeval/benchmarks/human_eval/human_eval.py
+++ b/deepeval/benchmarks/human_eval/human_eval.py
@@ -92,7 +92,7 @@ class HumanEval(DeepEvalBaseBenchmark):
         self.predictions: Optional[pd.DataFrame] = None
         self.task_scores: Optional[pd.DataFrame] = None
         self.overall_score: Optional[float] = None
-        self.verbose_mode: bool = (False,)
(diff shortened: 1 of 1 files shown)


penguine-ip:
@dermodmaster thank you!

PR #2335, opened 2025-11-28
fix(execute): avoid raising on gather timeout in non-indicator path

### LINKED ISSUE #2315: a_execute_test_cases raises only when not show_indicator
https://github.com/confident-ai/deepeval/blob/79723138a662d181cf22d5e2bcbdcc94b84f368c/deepeval/evaluate/execute.py#L806

When the method branches at `if display_config.show_indicator and _use_bar_indicator:` it enters two very similar pieces of boilerplate code, however in the `else` path there is a `raise` within the `except (asyncio.TimeoutError, TimeoutError)`, but not in the `if` path.

I can't see a reason for this divergent behaviour, and it means that `a_execute_test_cases` will raise only in the case that we don't show the indicator.

BloggerBust:
Fixes #2315

### DIFF
--- a/deepeval/evaluate/execute.py
+++ b/deepeval/evaluate/execute.py
@@ -718,6 +718,8 @@ async def a_execute_test_cases(
                     "Gather timed out after %.1fs. Some metrics may be marked as timed out.",
                     _gather_timeout(),
                 )
+                if not error_config.ignore_errors:
+                    raise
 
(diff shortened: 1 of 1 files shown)


trevor-cai:
Just noticed a test was failing in this PR, so I set it to draft until I have pushed a fix for the test.

PR #2339, opened 2025-11-30
fix(metrics): enforce valid JSON templates and discourage fences

### LINKED ISSUE #2332: Issue: Bad JSON Templates in DeepEval
## Problem

DeepEval's prompt templates in `dag/templates.py` had two issues:

1. Used Python boolean syntax (`True`/`False`) in JSON examples instead of valid JSON (`true`/`false`)
2. No explicit instruction to avoid markdown code fences

This caused LLMs to return responses like:
```
\`\`\`json
{
    "verdict": True,
    "reason": "..."
}
\`\`\`
```

### Reproduction

I used `gpt-4o` (via internal company's API). \
The simple evaluation with GEval and DAGMetric from `deepeval` library failed for me **because of parsing error**.

## Proposed Fix

Improve the templates to (example from `deepeval/metrics/dag/templates.py` for `TaskNodeTemplate`):

Before:
```text
===END OF INSTRUCTIONS===

**
IMPORTANT: Please make sure to only return in JSON format, with the 'output' key as the output from the instructions.
Example JSON:
{{
    "output": "your output goes here"
}}
**

JSON:
```

After:
```text
===END OF INSTRUCTIONS===

**
IMPORTANT: Return ONLY raw JSON without markdown code fences. Use valid JSON values (true/false/null, not Python True/False/None).
Example JSON:
{{
    "output": "your output goes here"
}}
**

JSON:
```

### Files to modify

- `deepeval/metrics/dag/templates.py`
- `deepeval/metrics/g_eval/template.py`

Note: probably the similar problems exist in other templates.

BloggerBust:
- Update DAG metric templates to use valid JSON examples (true/false, string verdicts) and explicitly forbid markdown code fences
- Strengthen GEval and Conversational GEval templates to require "valid and parseable JSON" and avoid markdown code fences and extra text
- Validate with tests

Fixes #2332

### DIFF
--- a/deepeval/metrics/conversational_g_eval/template.py
+++ b/deepeval/metrics/conversational_g_eval/template.py
@@ -10,10 +10,10 @@ Evaluation Criteria:
 {criteria}
 
 **
-IMPORTANT: Please make sure to only return in JSON format, with the "steps" key as a list of strings. No words or explanation is needed.
+IMPORTANT: Please make sure to only return in valid and parseable JSON format, with the "steps" key as a list of strings. No words or explanation is needed. Do not wrap the JSON in markdown code fences (for example ```json ... ```), and do not include any extra text, commentary, or formatting.
 Example JSON:
 {{
-    "steps": <list_of_strings>
+    "steps": ["Check if the answer addresses the user query.", "Verify factual correctness where applicable.", "Assess clarity and coherence of the answer."]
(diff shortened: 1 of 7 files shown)


trevor-cai:
FIxed by https://github.com/confident-ai/deepeval/pull/2365

PR #2350, opened 2025-12-03
feat(models): update all models to have consistent argument names and positioning

BloggerBust:
- Refactored all models to have consistent argument names and ordering
- Refactored all relevant models to support both old and new keyword arguments for backwards compatability
- Implemented alias mapping for deprecated keywords and log warning when used
- Ensured compatibility by updating constructor logic and adjusting client initialization.
- Added tests to verify compatibility with legacy keyword arguments.
- Updated model handling to ensure proper integration with existing and new parameters.


trevor-cai:
Closed because this PR is now part of #2356

PR #2355, opened 2025-12-04
fix(gemini_model): update use_vertexai check to bool

nwhite365:
Removing string check for use_vertexai, as it is based on the `GOOGLE_GENAI_USE_VERTEXAI` boolean environment variable.


trevor-cai:
Hi @nwhite365,

thanks a lot for the contribution!

Your fix is exactly right. This same change was included in a recent larger refactor that has since merged (#2378 ), along with a few other GeminiModel fixes.

It’s now released in deepeval 3.7.6, so upgrading to that version will include the fix.

Issue #2366, opened 2025-12-08
AttributeError: 'SummarizationAlignmentVerdict' object has no attribute 'verdicts' -- Async False


dgrahn:
**Describe the bug**
When `run_async=False`, `SummarizationMetric` has an attribute error when it is evaluated.

**To Reproduce**

1. Run `pip install -qU deepeval==3.7.4`
2. Run the following code:

```py
from dotenv import load_dotenv
from deepeval import evaluate
from deepeval.metrics import SummarizationMetric
from deepeval.evaluate import AsyncConfig
from deepeval.test_case import LLMTestCase

load_dotenv('.env')

tests = [
    LLMTestCase(
        input='Marley was dead: to begin with...',
        actual_output="The text opens by firmly establishing that Marley is dead...",
        expected_output=None,
        context=None,
        retrieval_context=None,
        additional_metadata=None,
        tools_called=None,
        comments=None,
        expected_tools=None,
        token_cost=None,
        completion_time=None,
        name=None,
        tags=None,
        mcp_servers=None,
        mcp_tools_called=None,
        mcp_resources_called=None,
        mcp_prompts_called=None
    )
]

metric = SummarizationMetric(model='gpt-4.1-nano')
results = evaluate(tests_41, [metric], async_config=AsyncConfig(run_async=False))
```

3. View error:

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
/tmp/ipykernel_32063/195509519.py in <cell line: 34>()
     32 
     33 metric = SummarizationMetric(model='gpt-4.1-nano')
---> 34 results = evaluate(tests_41, [metric], async_config=AsyncConfig(run_async=False))

[~/anaconda3/envs/python3/lib/python3.10/site-packages/deepeval/evaluate/evaluate.py](https://gai-1101-rwzk.notebook.us-west-2.sagemaker.aws/home/ec2-user/anaconda3/envs/python3/lib/python3.10/site-packages/deepeval/evaluate/evaluate.py) in evaluate(test_cases, metrics, metric_collection, hyperparameters, identifier, async_config, display_config, cache_config, error_config)
    246                 )
    247             else:
--> 248                 test_results = execute_test_cases(
    249                     test_cases,
    250                     metrics,

[...]

Issue #2318, opened 2025-11-21
Structured output for knowledge retention metric


tasneem200:
**Describe the bug**
Using structured output with schema does not work properly for the knowledge retention metric and the schema defined and passed in for it.

**To Reproduce**
1. Set up a DeepEval KnowledgeRetentionMetric with a custom LLM wrapper that supports structured output
2. Enable structured output by passing the schema argument to the LLM's generate method
3. Run the knowledge retention evaluation on conversational data
4. See ValidationError: "1 validation error for Knowledge data Field required"

**Expected behavior**
Model should return a response with structured output containing the evaluation result, score, and reason for the score. 

**Screenshots/Examples**
Examples of **truncated** model responses in the case where structured output is attempted by passing in the schema argument and without the schema are shown below. Different models were tested and in all cases the LLM returned an empty dictionary in the "content" field of the response when schema is used. 

Example 1: MODEL RESPONSE WITH STRUCTURED OUTPUT:  content='{}' additional_kwargs={'refusal': None} response_metadata={'token_usage': {'completion_tokens': 2, 'prompt_tokens': 670, 'total_tokens': 672, 'completion_tokens_details': None, 'prompt_tokens_details': None}, 'model_name': 'gpt-4.1-2025-04-14', ...

Example 2: MODEL RESPONSE WITHOUT STRUCTURED OUTPUT:  {
    "reason": "The score is 1.00 because there are no attritions, indicating no forgetfulness or loss of previously established knowledge in the LLM's responses."} content='{\n    "reason": "The score is 1.00 because there are no attritions, indicating no forgetfulness or loss  of previously established knowledge in the LLM\'s responses."\n}' additional_kwargs={'refusal': None} response_metadata={'token_usage': {'completion_tokens': 39, 'prompt_tokens': 151, 'total_tokens': 190, 'completion_tokens_details': None, 'prompt_tokens_details': None}, 'model_name': 'gpt-4.1-2025-04-14', ...

Ideally, the case where structured output is attempted should also include a real evaluation result in the response. 

**Desktop (please complete the following information):**
 - OS: macOS
 - Python: 3.10
 - DeepEval: Latest version
 - LangChain with Azure OpenAI
 - Model: GPT 4.1: 04-14-2025

penguine-ip:
Hey @tasneem200 can you please show the error stack? Also which DeepEval version are you on? I know for sure that a few months back not everything was structured so maybe this could be the issue

tasneem200:
[...]

PR #2367, opened 2025-12-09
refactor(metrics): centralize structured output handling via shared helpers

BloggerBust:
This PR updates metric implementations to use the new structured output helpers so that all schema / JSON-mode logic is centralized instead of being duplicated in each metric.

At a high level:

* All metrics that previously called `model.generate` / `model.a_generate` with ad hoc schema / JSON handling now use:

  * `generate_with_schema_and_extract`
  * `a_generate_with_schema_and_extract`

This PR updates metric implementations to use the new structured-output helpers so that all schema / JSON-mode logic is centralized instead of being duplicated in each metric.

At a high level:

* All metrics that previously called `model.generate(...)` / `model.a_generate(...)` with ad-hoc schema / JSON handling now use:

  * `generate_with_schema_and_extract(...)`
  * `a_generate_with_schema_and_extract(...)`

* Expose capability methods so that we can test for structured output / json-mode capability in a consistent way:

  * `supports_structured_outputs()`
  * `supports_json_mode()`
  
Fixes: #2318 
Fixes: #2366

### DIFF
--- a/deepeval/metrics/knowledge_retention/knowledge_retention.py
+++ b/deepeval/metrics/knowledge_retention/knowledge_retention.py
@@ -198,7 +198,7 @@ class KnowledgeRetentionMetric(BaseConversationalMetric):
             accumulated_knowledge = [
                 knowledge.data
                 for knowledge in self.knowledges[:i]
-                if knowledge is not None
+                if knowledge is not None and knowledge.data
             ]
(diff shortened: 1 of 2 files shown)


BloggerBust:
Closing as this PR is now part of https://github.com/confident-ai/deepeval/pull/2378

PR #2379, opened 2025-12-15
Adding naive truncation to prettify list

DanielYakubov:
# Naive Truncation to prettify_list

Adding an option for truncation in prettify_list util function + some documentation. It's naive because it can cut off pprints of JSON outputs. Useful for debug displays where the whole output is not needed in repeated inference.


penguine-ip:
Hey @DanielYakubov going to close this PR since its not being used anywhere, so change no necessary

Issue #2385, opened 2025-12-17
Answer Relevancy metric returns 1.0 even if `actual_output` is empty


GeisaFaustino:
**Describe the bug**
Answer Relevancy metric returns 1.0 even if `actual_output` is empty

**To Reproduce**
```
from deepeval.metrics import AnswerRelevancyMetric
from deepeval.test_case import LLMTestCase

answer_relevancy = AnswerRelevancyMetric(model=MODEL)

test_case = LLMTestCase(
    input="What if these shoes don't fit?",
    actual_output="" # an empty output
)

answer_relevancy.measure(test_case)
print(f"Score:{answer_relevancy.score}\nReason: {answer_relevancy.reason}")
```
Score:1
Reason: The score is 1.00 because the response is fully relevant and directly addresses the input without any irrelevant statements. Great job!

**Expected behavior**
For an empty output, the expected behavior would be a score of 0.0 or close to it, as the output provides no relevant information to address the input.

trevor-cai:
@GeisaFaustino thank you for reporting this. I will investigate and will reach out if I have any questions.

trevor-cai:
@GeisaFaustino The fix is in review. Thanks again for reporting this and providing repro steps, I really appreciate it.

PR #2386, opened 2025-12-17
fix(answer-relevancy): score empty actual_output as 0 and skip LLM calls

BloggerBust:
- Short-circuit measure/a_measure when actual_output is empty (incl. multimodal)
- Return 0 when no verdicts are produced
- Add shared test make_metric stub and regression tests for sync/async paths

Fixes: #2385


trevor-cai:
Closing in favor of https://github.com/confident-ai/deepeval/pull/2451

PR #2388, opened 2025-12-18
feat(cli): standardization of cli commandline interface

BloggerBust:


deepeval/cli/main.py
```diff
@@ -660,352 +972,759 @@ def unset_openai_env(
...
+        edit_ctx.switch_model_provider(
+            EmbeddingKeyValues.USE_AZURE_OPENAI_EMBEDDING
         )
+        if model is not None:
+            settings.AZURE_EMBEDDING_MODEL_NAME = model
+        if deployment_name is not None:
+            settings.AZURE_EMBEDDING_DEPLOYMENT_NAME = deployment_name
+
+    handled, path, updates = edit_ctx.result
+    effective_model = settings.AZURE_EMBEDDING_MODEL_NAME or "<unset>"
```
penguine-ip: What will be printed when it is `<unset>`?

trevor-cai: This clears the settings, so that the model name is unset
```
dustfinger@primus:~/dev/deepeval$ poetry run deepeval unset-azure-openai-embedding
🙌 OpenAI will still be used by default because OPENAI_API_KEY is set.
```

Here is what prints when `set-azure-openai` is called with the model name unset.
```
dustfinger@primus:~/dev/deepeval$ poetry run deepeval set-azure-openai
Tip: persist these settings to a dotenv file with --save=dotenv[:path] (default 
.env.local) or set DEEPEVAL_DEFAULT_SAVE=dotenv:.env.local
🙌 Congratulations! You're now using Azure OpenAI's `<unset>` for all evals that 
require an LLM.
```

PR #2402, opened 2026-01-02
bug/issue-2401: batch scoring on DROP Benchmark

### LINKED ISSUE #2401: Bug: Batch Predict on DROP Benchmark
**Describe the bug**
`batch_predict` on DROP benchmark expects exact match with a list of possible answers, rather than marking correct if the prediction matches just one of the possible answers.

**To Reproduce**
Run DROP benchmark with batch_size > 1
Will have a score much lower than expected, and much lower than having no batch_size

**Expected behavior**
Mark correct if prediction is in the list of answers

AadamHaq:
Closes #2401 

Logic works similar to scoring with batch_predict=None. Tested and it works.

Example of bug:
Prediction: '2', Answer: '2, 2-yards'
Prediction wanted model to return the full '2, 2-yards' instead of correctly marking '2' as correct. Batch prediction used `quasi_exact_match_score` instead of `quasi_contains_score` like the predict function.

### DIFF
--- a/deepeval/benchmarks/drop/drop.py
+++ b/deepeval/benchmarks/drop/drop.py
@@ -282,7 +282,9 @@ class DROP(DeepEvalBaseBenchmark):
             expected_output = DROPTemplate.parse_str_to_list(
                 golden.expected_output, DELIMITER
             )
-            score = self.scorer.quasi_contains_score(expected_output, prediction)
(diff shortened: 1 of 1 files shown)


penguine-ip:
Hey @AadamHaq thank you!

PR #2406, opened 2026-01-05
docs : Update in the GEval - docs - to include the evaluation_steps instead …

Vishnu-sai-teja:
…of the evaluation_params.


penguine-ip:
@Vishnu-sai-teja thank you!

PR #2408, opened 2026-01-05
Bug/issue-2407: MMLU batch_predict type

### LINKED ISSUE #2407: Bug: Batch predict in MMLU requires List[List[BaseModel]] rather than List[BaseModel]
**Describe the bug**
Batch predict in MMLU requires List[List[BaseModel]] rather than List[BaseModel]

**To Reproduce**
Bug in code, have fixed in a PR. To reproduce, run MMLU code using customer adapter that returns List[BaseModel] in batch generation

**Expected behavior**
List[BaseModel] should work (as in type hint)

AadamHaq:
Closes #2407 

Bug in MMLU where only List[List[BaseModel]] or List[Tuple(BaseModel)] would work rather than List[BaseModel] as expected and used in all other benchmarks


penguine-ip:
Hey @AadamHaq thanks for catching this!!

PR #2410, opened 2026-01-05
fix: trimAndLoadJson failure for llm outputs with <think> tags

abhijitmamarde:
Some evaluation models (ex: llama-3_3-nemotron-super-49b-v1_5) emit reasoning wrapped in `<think>`...`</think>` tags, which caused JSON parsing to fail. This change strips those blocks before running the existing JSON trimming and parsing logic.


abhijitmamarde:
admins, Any feedback you can give in order to merge this PR?  @tanayvaswani @trevor-cai  @kritinv could you please have a look? Thanks in advance!

abhijitmamarde:
Gentle nudge to core members!!! 
Let me know if anything I could do to help merge this PR.

Thanks!

abhijitmamarde:
authors need your action pls!

penguine-ip:
Thanks for the contribution @abhijitmamarde, and apologies for the long wait! The motivation is valid — reasoning models (Nemotron, DeepSeek-R1, QwQ, etc.) do emit `<think>...</think>` blocks in raw content and they can break parsing. Before we can merge, though, a few things need addressing:

1. **Rebase onto latest `main` (currently has conflicts).** `trimAndLoadJson` was just restructured in #2701 to a "parse strictly first, clean up only on failure" approach (to avoid corrupting valid JSON). Your branch predates that.

2. **Move the `<think>` stripping into the fallback path rather than running it unconditionally.** As written it mutates every input on every call, which goes against the #2701 direction and could (rarely) corrupt a valid JSON object whose string value legitimately contains the literal text `<think>...</think>`. It should only kick in after a direct `json.loads` fails, then re-run the `{`/`}` extraction.

   Worth noting: the existing `find("{")` / `rfind("}")` extraction already handles the common case where a brace-free think block precedes the JSON (e.g. `<think>reasoning</think>{"score": 1}` already parses today). Your change really only adds value when the think block itself **contains braces** — so it'd be great to call that out and cover it in a test.

3. **Handle the closing-tag-only variant.** Many chat templates inject the opening `<think>` so the model only emits the closing `</think>` — e.g. `reasoning...</think>{...}`. The current `<think>.*?</think>` regex won't match that. Consider stripping everything up to and including a `</think>` when present, not just fully-paired tags.

4. **Add regression tests** under `tests/test_core/` — at minimum: (a) a `<think>` block containing braces round-trips to the correct JSON, (b) valid JSON whose string value contains literal `<think>` text is left untouched, and (c) the closing-tag-only case.

If you'd like, point them at `tests/test_core/test_trim_and_load_json.py` (added in #2701) as a template. Happy to re-review once these are in. Thanks again!

Issue #2413, opened 2026-01-05
`synthesizer.to_pandas()` returns `KeyError: 'evolutions'`


GeisaFaustino:
**Describe the bug**
It is not possible to convert the synthetically generated goldens into a DataFrame as illustrated in the [documentation](https://deepeval.com/docs/synthesizer-introduction).
`synthesizer.to_pandas()` returns `KeyError: 'evolutions'`.

**To Reproduce**
````python
from deepeval.synthesizer import Synthesizer

print(f"We are using the model: {aoai_model.get_model_name()}")
synthesizer = Synthesizer(model=aoai_model)

conversational_goldens = synthesizer.generate_conversational_goldens_from_contexts(
    # Provide a list of context for synthetic data generation
    contexts=[
        ["The Earth revolves around the Sun.", "Planets are celestial bodies."],
        ["Water freezes at 0 degrees Celsius.", "The chemical formula for water is H2O."],
    ]
)

print(f"Generated {len(conversational_goldens)} conversational goldens.")

df = synthesizer.to_pandas()
df.head()
````

Output:
```
We are using the model: gpt-4o (Azure)

Generated 4 conversational goldens.

---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Cell In[5], [line 1](vscode-notebook-cell:?execution_count=5&line=1)
----> [1](vscode-notebook-cell:?execution_count=5&line=1) df = synthesizer.to_pandas()
      2 df.head()

File /opt/anaconda3/envs/deep-eval-378/lib/python3.11/site-packages/deepeval/synthesizer/synthesizer.py:1438, in Synthesizer.to_pandas(self)
   1435 df = pd.DataFrame(data)
   1437 # Optional: Fill NaN evolutions for better clarity
-> [1438](https://file+.vscode-resource.vscode-cdn.net/opt/anaconda3/envs/deep-eval-378/lib/python3.11/site-packages/deepeval/synthesizer/synthesizer.py:1438) df["evolutions"] = df["evolutions"].apply(
   1439     lambda x: x if x is not None else "None"
   1440 )
   1442 return df

File /opt/anaconda3/envs/deep-eval-378/lib/python3.11/site-packages/pandas/core/frame.py:4113, in DataFrame.__getitem__(self, key)
   4111 if self.columns.nlevels > 1:
   4112     return self._getitem_multilevel(key)
[...]

PR #2415, opened 2026-01-06
Fix synthesizer

A-Vamshi:
Synthesizer now supports `save_as` and `to_pandas` for conversational goldens as well. Fixes the issue mentioned in #2413


satyakic22-crypto:
We are using 4.0.7 version and still getting the same error with to_pandas()

PR #2441, opened 2026-01-17
Fixed Return Type Annotations

yzhao244:
### **Fixed Return Type Annotations**
   - **File**: `deepeval/annotation/annotation.py`
   - **Issue**: Functions `send_annotation()` and `a_send_annotation()` had incorrect return type `-> str` but don't return anything
   - **Fix**: Changed return type to `-> None` to match actual behavior


penguine-ip:
@yzhao244 thank you!

PR #2444, opened 2026-01-19
fix(crewai): restore trace/span context across event callbacks

BloggerBust:
CrewAI events can fire in different execution contexts where ContextVars don’t propagate, causing Observer.__exit__ to mismatch the current span. Capture and restore trace/span context for LLM/tool/knowledge events and use trace_manager span lookups to finalize spans safely. A similar issue was seen in our langchain integration and a similar strategy was used to resolve the issue here.

Also:
- fallback to pysqlite3-binary when Chroma fails due to old system sqlite
- pass metric_collection through to created spans
- ensure wrappers set observer.result so Observer can finalize output
- handle CrewAI LLM factory __new__ by filtering DeepEval kwargs

### DIFF
--- a/deepeval/integrations/langchain/callback.py
+++ b/deepeval/integrations/langchain/callback.py
@@ -1,4 +1,7 @@
-from typing import Any, Optional, List, Dict
+import logging
+import threading
+
+from typing import Any, Optional, List, Dict, Tuple
 from uuid import UUID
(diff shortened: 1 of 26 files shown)


BloggerBust:
Closing this draft as the approach changed / was superseded elsewhere. Happy to reopen or rework if still useful.

PR #2454, opened 2026-01-22
Fix variable name from 'golden' to 'goldens'

himanshutech4purpose:
golden is typo 
for goldens in dataset_item:
  trip_planner_agent(goldens.input) # it is correct
  # trip_planner_agent(golden.input) # it is incorrect


kritinv:
Hey @himanshutech4purpose, thanks for flagging. Although golden is correct in the iterator, and we would want `goldens.input` to be `golden.input` instead.

Issue #2443, opened 2026-01-19
Change cache folder path


vection:
Is your feature request related to a problem? Please describe.
1. When running DeepEval inside a pod, I would like to configure the cache directory to point to a persistent storage location.
2. When scaling DeepEval across multiple workers, I would like them to share a single cache directory.

Describe the solution you'd like
Make the cache directory configurable via an environment variable.

Describe alternatives you've considered
The second use case is more complex, since I’m not sure how DeepEval currently handles concurrent reads and writes from multiple processes or machines. Any guidance or recommendations on this would be appreciated.

Additional context
This would make it much easier to run DeepEval in containerized and distributed environments (e.g., Kubernetes) with persistent or shared storage.

kritinv:
Hey @vection. Yup, 1.) seems like something that we could add pretty quickly. I'm wondering if there is a specific feature you are looking to utilize with cache? Is it running evaluations?

vection:
Hi @kritinv 
Thank you for the quick response , it would be great if you could add this option.
My use case involves running evaluation flows on 10K+ rows with multiple metrics.
What I’m trying to achieve is the ability to resume from the point of failure if an evaluation crashes or at least reuse previous responses to avoid repeating calls.

This is especially important since these evaluations are running in a Kubernetes environment, where the local storage path is not persistent.

penguine-ip:
Hey @vection, this looks useful, we're more than happy to accept a PR to help facilitate this if you could contribute one? thank you in advance

vection:
Hi @penguine-ip 
Sounds reasonable, I can do it.

penguine-ip:
Hey @vection thank you, it looks great, merging and cutting a release for this soon.

PR #2455, opened 2026-01-23
Able to change cache folder path 

vection:
PR for issue: #2443 
I implemented the simplest solution by allowing the HIDDEN_DIR constant to be overridden via an environment variable.

Ideally, the cache path should be taken directly from the settings object, but that currently causes circular import issues since settings already depends on constants. I also preferred not to remove HIDDEN_DIR, as it is used in many places in the codebase.

If you think there’s a better approach, I’m open to talk on it


penguine-ip:
Hey @vection thank you!

PR #2464, opened 2026-01-31
Fix AzureOpenAI keyless auth by deferring credential validation to SDK

### LINKED ISSUE #2431: Deepeval AzureOpenAI requires an api_key even when key based auth is disabled
**Describe the bug**
I am trying to construct a Haystack evaluation pipeline. One of the evaluation steps of this former requires a `DeepEvalEvaluator`. This latter requires a `metric_params={"model": [insert evaluation_model]` parameter - which can either be a string of the name of a model (with its relative environment variables saved) or an instance of a `deepeval.models` class. In my case, I would use an `AzureOpenAIModel` instance. 

However, once running the haystack pipeline, a PipelineRuntimeError appears from the component `DeepEvalEvaluator`: 
```
The following component failed to run:
Component name: 'contextual_recall_evaluator'
Component type: 'DeepEvalEvaluator'
Error: Error code: 403 - {'error': {'code': 'AuthenticationTypeDisabled', 'message': 'Key based authentication is disabled for this resource.'}}
```

What if I remove the azure api key? Well if I do, an error appears once instantiating the `AzureOpenAIModel`:
```
Exception has occurred: DeepEvalError
AzureOpenAI API key is not configured. Set AZURE_OPENAI_API_KEY in your environment or pass `api_key` to AzureOpenAIModel(...).
```

NB: My Azure OpenAI models cannot be changed to be key-based. 
Therefore, my question becomes **Would it be possible to initialise a deepeval AzureOpenAIModel that is not key-based? Or is there another way to initialise DeepEvalEvaluators with not key-based AzureOpenAIModel environment variables?**

**To Reproduce**
1. Have an Azure OpenAI deployed model that is not key-based.
2. Initialise your model

The model gets initialised as such:
```
settings = get_settings() # PydanticSettings object loading a .env

azure_model = AzureOpenAIModel(
    deployment_name=settings.azure_deployment_name,
    model=settings.azure_deployment_name,
    api_version=settings.azure_openai_version_okf,
    base_url=settings.azure_openai_endpoint,
)
```
Without any api_key, the error appears here.

**Expected behavior**
I expect that the `AzureOpenAIModel` class can be initialised similarily as the `AzureOpenAI` class from `openai` library when using none-key-based authentication. I.e.: 
```
from openai import AzureOpenAI
from azure.identity import ClientSecretCredential, get_bearer_token_provider
credential = ClientSecretCredential(
    tenant_id=self.tenant_id,
    client_id=self.client_id,
    client_secret=self.client_secret,
)

token_provider = get_bearer_token_provider(
    credential, "https://cognitiveservices.azure.com/.default"
)

self._client = AzureOpenAI(
    azure_ad_token_provider=token_provider,
    api_version=self.api_version,
    azure_endpoint=self.endpoint,
)
```

 - OS: Linux, Ubuntu 22.04
 - Browser: Firefox
 - Version: deepeval-haystack>= 1.0.0 | haystack-ai>= 2.20.0

ppon1086:
### What changed

- Defer Azure OpenAI authentication validation to the OpenAI SDK
- Fail fast **only** when a credential is explicitly provided but empty
- Preserve existing behavior for key-based authentication
- Support `SecretStr` and string credentials consistently

### Why this is needed
DeepEval currently raises an error when `api_key` is missing or empty for `AzureOpenAIModel`, even when key-based authentication is intentionally disabled and Azure AD / managed identity is intended. This prevents valid keyless Azure OpenAI configurations from working and causes an early failure before the OpenAI SDK can resolve authentication.

This PR removes premature client-side enforcement and defers authentication handling to the official OpenAI/Azure SDK, aligning DeepEval’s behavior with Azure OpenAI’s documented authentication model and avoids false negative  configuration errors.

### Tests

- Added unit tests covering:
  - Keyless / managed identity scenarios
  - Explicit empty credential failures
  - `SecretStr` handling
- All existing tests pass

### Related issue

Fixes #2431

### DIFF
--- a/deepeval/models/llms/azure_model.py
+++ b/deepeval/models/llms/azure_model.py
@@ -480,6 +480,14 @@ class AzureOpenAIModel(DeepEvalBaseLLM):
                 )
             # else: neither key nor token nor provider set -> defer to SDK
 
+        # Enforce precedence: provider > token > api_key
+        
+        if self.azure_ad_token_provider is not None:
+            azure_ad_token_value = None
+            api_key_value = None
+        elif azure_ad_token_value is not None:
+            api_key_value = None
+        # else: api_key_value may be used (or None => SDK-managed auth)
 
         kw = dict(
             api_key=api_key_value,
@@ -487,7 +495,7 @@ class AzureOpenAIModel(DeepEvalBaseLLM):
             azure_endpoint=self.base_url,
             azure_deployment=self.deployment_name,
             azure_ad_token_provider=self.azure_ad_token_provider,
-            azure_ad_token=self.azure_ad_token_value,
+            azure_ad_token=azure_ad_token_value,
             **self._client_kwargs(),
         )
         try:


deepeval/models/llms/azure_model.py
```diff
@@ -442,33 +442,52 @@ def _client_kwargs(self) -> Dict:
...
 
         kw = dict(
-            api_key=api_key,
+            api_key=api_key_value,
             api_version=self.api_version,
             azure_endpoint=self.base_url,
             azure_deployment=self.deployment_name,
             azure_ad_token_provider=self.azure_ad_token_provider,
-            azure_ad_token=azure_ad_token,
+            azure_ad_token=self.azure_ad_token_value,
```
A-Vamshi: Is there a reason we're using `self.azure_ad_token_value` here and not `azure_ad_token_value` you created above?

deepeval/models/llms/azure_model.py
```diff
@@ -442,33 +442,52 @@ def _client_kwargs(self) -> Dict:
...
+                api_key_value = str(self.api_key)
+
+        azure_ad_token_value = None
+        if self.azure_ad_token is not None:
+            try:
+                azure_ad_token_value = self.azure_ad_token.get_secret_value()
+            except Exception:
+                azure_ad_token_value = str(self.azure_ad_token)
 
         if self.azure_ad_token_provider is None:
```
A-Vamshi, line 463: In this block we're following a precedence, we're giving priority to `azure_ad_token_provider` first, if that doesn't exist we give priority to `azure_ad_token` over `api_key`. This allows us to handle the case where a user provides all 3 of these fields (azure_ad_token_provider > azure_ad_token > api_key), since we're already populating `api_key_value` and `azure_ad_token_value` which are all being passed to the client below, I'm worried about which one is used by the client in this case? Do you know what happens in this case and if it's safe to do this?

ppon1086: Thank you for the review, Vamshi. I've fixed the auth precedence (provider>token>api_key) for scenarios when multiple credentials are provided, and replaced the use of  self.azure_ad_token_value with local azure_ad_token_value. Please let me know if anything else looks off.

A-Vamshi:
Thanks for the PR @ppon1086, this looks great!

PR #2471, opened 2026-02-03
Fix AccessDeniedException when using AmazonBedrockModel in async evaluations

### LINKED ISSUE #1916: AccessDeniedException with AmazonBedrockModel in async evaluation context
### Description
The `AmazonBedrockModel` class fails with `AccessDeniedException` when used in DeepEval's async evaluation context, despite having valid AWS credentials and successful direct Bedrock API access.

### Environment
- **DeepEval Version**: 3.3.5 and 3.3.6 (tested both)
- **Python Version**: 3.11.11
- **AWS Region**: eu-west-1
- **Model**: anthropic.claude-3-haiku-20240307-v1:0

### Steps to Reproduce
1. Configure AWS credentials in `.env` file with valid `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`
2. Set up DeepEval evaluation with `AmazonBedrockModel`:
   ```python
   model_instance = AmazonBedrockModel(
       model_id="anthropic.claude-3-haiku-20240307-v1:0",
       aws_access_key_id=aws_access_key,
       aws_secret_access_key=aws_secret_key,
       region_name="eu-west-1"
   )
   ```
3. Run evaluation using `evaluate(test_cases=test_cases, metrics=metrics)`

### Expected Behavior
Evaluation should proceed successfully using the Anthropic Claude model via AWS Bedrock.

### Actual Behavior
botocore.errorfactory.AccessDeniedException: An error occurred (AccessDeniedException) when calling the Converse operation: You don't have access to the model with the specified model ID.

### Additional Context
- ✅ **Direct boto3 access works**: Same credentials work perfectly with direct `boto3.client('bedrock-runtime')` calls
- ✅ **Model access confirmed**: The model is accessible and working via direct API calls
- ✅ **Credentials are loaded**: Environment variables are properly set and accessible
- ❌ **Only fails in DeepEval**: The error only occurs when using `AmazonBedrockModel` within DeepEval's async evaluation context
- ✅ **OpenAI models work**: Switching to `gpt-4o-mini` works perfectly with the same evaluation setup

### Workaround
Currently using OpenAI models as a workaround, but would prefer to use AWS Bedrock models.

### Possible Root Cause
The issue appears to be related to how AWS credentials are handled in the async evaluation context, possibly due to environment variable inheritance or session management in concurrent/async operations.

Fizza-Mukhtar:
### Bug Fix: AmazonBedrockModel fails in async evaluation context

This PR fixes an issue where `AmazonBedrockModel` raises an
`AccessDeniedException` when used inside DeepEval’s async evaluation
pipeline, despite valid AWS credentials and confirmed Bedrock access.

#### Problem
- `AmazonBedrockModel` worked with direct `boto3` calls
- Failed only inside DeepEval async evaluations
- Error: `AccessDeniedException` during `Converse` calls
- Root cause: async session / credential handling was not aligned with
  `aiobotocore` best practices in concurrent contexts

#### Solution
- Ensure `aiobotocore` session handling is async-safe
- Avoid credential loss during concurrent async evaluations
- Keep behavior unchanged for sync execution paths

#### Testing
- Verified behavior with mocked `aiobotocore` async sessions
- Confirmed no regression for non-Bedrock models

This brings Bedrock behavior in line with other async-supported LLM providers in DeepEval.
Fixes #1916

### DIFF
--- a/deepeval/models/llms/amazon_bedrock_model.py
+++ b/deepeval/models/llms/amazon_bedrock_model.py
@@ -42,6 +42,7 @@ class AmazonBedrockModel(DeepEvalBaseLLM):
         model: Optional[str] = None,
         aws_access_key_id: Optional[str] = None,
         aws_secret_access_key: Optional[str] = None,
+        aws_session_token: Optional[str] = None,
         cost_per_input_token: Optional[float] = None,
         cost_per_output_token: Optional[float] = None,
         region: Optional[str] = None,
@@ -85,6 +86,14 @@ class AmazonBedrockModel(DeepEvalBaseLLM):
         else:
             self.aws_secret_access_key = settings.AWS_SECRET_ACCESS_KEY
 
+        if aws_session_token is not None:
+            self.aws_session_token: Optional[SecretStr] = SecretStr(
+                aws_session_token
+            )
+        else:
+            self.aws_session_token = settings.AWS_SESSION_TOKEN
+    
+    
         # Dependencies: aiobotocore & botocore
         aiobotocore_session = require_dependency(
             "aiobotocore.session",
@@ -273,6 +282,10 @@ class AmazonBedrockModel(DeepEvalBaseLLM):
             client_kwargs["aws_secret_access_key"] = (
                 self.aws_secret_access_key.get_secret_value()
             )
+        if self.aws_session_token is not None:
+            client_kwargs["aws_session_token"] = (
+                self.aws_session_token.get_secret_value()
+            )    
 
         async with self._session.create_client(
             "bedrock-runtime", **client_kwargs


penguine-ip:
Hey @Fizza-Mukhtar thanks for the PR, one thing i noticed is every call creates a fresh client and tears it down when the context manager exits (by not reusing `self._client`), is this intended?

Fizza-Mukhtar:
Thanks for catching that. Yep, that was intentional for this fix.

The AccessDenied issue seemed tied to how the async evaluation loop was reusing a long-lived aiobotocore client across concurrent tasks. In that setup, credentials or session state could get lost or resolved inconsistently, which caused Bedrock to fail even though direct boto3 usage worked.

Switching to creating the client inside an async context per call ensured:

* credentials are resolved within each task
* no cross-task session leakage
* consistent behavior with concurrent async evaluation

That said, I agree this does trade off client reuse. If there’s a preferred pattern here, for example caching the session but creating per-task clients or safely reusing `_client`, I’m happy to adjust. My main goal was making Bedrock reliable in the async evaluation pipeline without affecting sync paths.

Open to whichever approach fits best with DeepEval’s architecture 👍

penguine-ip:
Hey @Fizza-Mukhtar not at all I think this is great, just wanted to clarify, thank you!!

PR #2477, opened 2026-02-06
Remove dependency on opentelemetry-otlp-proto-grpc.

tbeadle:
In commit 1f903f25142b289ea034b62996e2bd58801a4832, the only import of opentelemtry.exporter.otlp.proto.grpc was removed and so this dependency should no longer be needed in pyproject.toml.


penguine-ip:
thanks @tbeadle !

PR #2484, opened 2026-02-10
refactor: use subprocess.run for robust path handling on Windows

RinZ27:
Refactored the `safe_rmtree` function to utilize `subprocess.run` with proper argument lists, replacing the unsafe `os.system` calls. Handling paths with spaces correctly on Windows was a priority here, as the previous string interpolation approach could easily fail or be exploited. Since `shutil.rmtree` is globally patched in this module, I decided to make this logic as robust as possible to ensure the entire cleanup lifecycle remains stable across different environments.

### DIFF
--- a/deepeval/synthesizer/chunking/context_generator.py
+++ b/deepeval/synthesizer/chunking/context_generator.py
@@ -12,6 +12,7 @@ import os
 import gc
 import tempfile
 import logging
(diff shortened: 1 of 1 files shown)


penguine-ip:
@RinZ27 thank you!

PR #2486, opened 2026-02-11
dep(click): Bump click upper bound to <8.4.0

mfaizanse:
### Summary

- widen click constraint to allow 8.3.x
- keep dependency compatibility as we prepare for the latest Click

### Notes

- includes a small dependency compatibility tweak for [pysqlite3-binary](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) on Linux only to avoid non-Linux install failures


penguine-ip:
@mfaizanse thank you!

PR #2491, opened 2026-02-16
Fix: AzureOpenAIModel sends unsupported temperature for reasoning models

aerosta:
Fixes an issue where `AzureOpenAIModel` sends `temperature=0.0` to models marked `supports_temperature=False` in `OPENAI_MODELS_DATA`, which can cause Azure API errors.

The OpenAI model wrapper already special-cases `supports_temperature=False` in `openai_model.py`. `AzureOpenAIModel` previously always included `temperature` in requests.

On Azure, some reasoning deployments reject unsupported parameters, so the correct behavior is to omit `temperature` rather than forcing a value.

## Changes

- In `AzureOpenAIModel.__init__`, if `supports_temperature is False`, set `self.temperature = None`
- Allow `temperature=None` through validation
- Conditionally omit `temperature` from request kwargs when `self.temperature is None`

## Tests

Added `tests/test_core/test_azure_model_temperature.py` covering:
- `supports_temperature=False` model → temperature is `None`
- Standard model default remains `0.0`
- Explicit temperature preserved for standard models
- Explicit temperature overridden to `None` for `supports_temperature=False` models

### DIFF
--- a/deepeval/models/llms/azure_model.py
+++ b/deepeval/models/llms/azure_model.py
@@ -138,6 +138,11 @@ class AzureOpenAIModel(DeepEvalBaseLLM):
         )
 
         self.model_data = OPENAI_MODELS_DATA.get(model)
+
+        # Omit temperature for models that don't support it
+        if self.model_data and self.model_data.supports_temperature is False:
+            temperature = None
+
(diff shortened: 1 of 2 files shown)


penguine-ip:
Hey @aerosta thank you!

PR #2493, opened 2026-02-16
fix: resolve 3 synthesizer bugs (AttributeError, wrong metadata, UnboundLocalError)

aerosta:
## Summary

Found during a code audit of `synthesizer.py`. All three bugs are runtime crashes or silent data corruption.

---

### Bug 1: `_a_generate_text_to_sql_from_context` crashes when `include_expected_output=False`

`expected_output` is `None`, then `.sql` access raises `AttributeError`.  
Also replaced sync `_generate_schema()` with `await _a_generate_schema()` since this is an async method.

### Bug 2: `generate_goldens_from_scratch` sync path - wrong evolutions metadata

`evolutions_used` leaks from loop 1 and all goldens get the last iteration's value.  
Matched the async path by storing `(prompt, evolutions)` tuples and unpacking in loop 2.

### Bug 3: `_rewrite_inputs` / `_a_rewrite_inputs` crash when `max_quality_retries=0`

`score` and `feedback` were never assigned when the retry loop is empty, causing `UnboundLocalError`.  
Initialized defaults before the loop.

---

## Tests

Added `tests/test_core/test_synthesizer_bugs.py` (5 tests):
```
pytest -q tests/test_core/test_synthesizer_bugs.py - 5 passed in 0.13s
```

## Changes

- `deepeval/synthesizer/synthesizer.py` - 11 insertions, 5 deletions
- `tests/test_core/test_synthesizer_bugs.py` - new file (248 lines)

### DIFF
--- a/deepeval/synthesizer/synthesizer.py
+++ b/deepeval/synthesizer/synthesizer.py
@@ -790,7 +790,7 @@ class Synthesizer:
                 prompt = SynthesizerTemplate.generate_text2sql_expected_output(
                     input=data.input, context="\n".join(context)
                 )
-                expected_output: SQLData = self._generate_schema(
+                expected_output: SQLData = await self._a_generate_schema(
                     prompt, SQLData, self.model
                 )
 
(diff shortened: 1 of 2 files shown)


penguine-ip:
Hey @aerosta thank you!

PR #2518, opened 2026-02-28
Fix duplicate example contexts in FilterTemplate.evaluate_context

### LINKED ISSUE #2039: FilterTemplate same Example context but， Example JSON not same
https://github.com/confident-ai/deepeval/blob/3e083b28975884b37f9af323b313f1665ae70f53/deepeval/synthesizer/templates/template.py#L255

Image

Fizza-Mukhtar:
The evaluate_context method had the same example context repeated 
three times with different scores, which is inconsistent.

Replaced the two duplicate examples with distinct contexts so each 
example has unique input and output.

Fixes #2039

### DIFF
--- a/deepeval/synthesizer/templates/template.py
+++ b/deepeval/synthesizer/templates/template.py
@@ -542,7 +542,7 @@ class FilterTemplate:
             "relevance": 0.5
         }}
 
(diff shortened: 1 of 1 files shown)


penguine-ip:
Hey @Fizza-Mukhtar good catch, thank you!

PR #2526, opened 2026-03-03
Fix equation formula for Goal Accuracy Score

JevDev2304:
Reason for the Update:
The original formula contained a logical error known as a circular dependency. It used the exact same term, "Goal Accuracy Score," as both the final result on the left side of the equation and as a variable being added on the right side.

The Fix:
To correct this and accurately reflect the underlying code implementation, the variable in the numerator was renamed to "Goal Evaluation Score." The corrected formula now clearly shows that the final metric is a straightforward arithmetic mean of two distinct assessments: the quality of the agent's execution (Goal Evaluation Score) and the quality of its reasoning (Plan Evaluation Score).


penguine-ip:
Hey @JevDev2304 thanks!

Issue #2533, opened 2026-03-05
Default provider not working for AWS Bedrock


leoeiji:
**Describe the bug**
When setting AWS Bedrock as default provider to use for LLM-as-a-judge, the `initialize_model()` method in `deepeval.metrics.utils` isn't prepared to check if Bedrock was set as default provider.

**To Reproduce**
1. `deepeval set-bedrock`
2. Define a test without model:

```python
from deepeval import assert_test
from deepeval.test_case import LLMTestCase, LLMTestCaseParams
from deepeval.metrics import GEval

def test_correctness():
    correctness_metric = GEval(
        name="Correctness",
        criteria="Determine if the 'actual output' is correct based on the 'expected output'.",
        evaluation_params=[LLMTestCaseParams.ACTUAL_OUTPUT, LLMTestCaseParams.EXPECTED_OUTPUT],
        threshold=0.5
    )
    test_case = LLMTestCase(
        input="I have a persistent cough and fever. Should I be worried?",
        actual_output="A persistent cough and fever could be a viral infection or something more serious. See a doctor if symptoms worsen or don't improve in a few days.",
        expected_output="A persistent cough and fever could indicate a range of illnesses, from a mild viral infection to more serious conditions like pneumonia or COVID-19. You should seek medical attention if your symptoms worsen, persist for more than a few days, or are accompanied by difficulty breathing, chest pain, or other concerning signs."
    )
    assert_test(test_case, [correctness_metric])
```

3. Run tests with `deepeval test run <test_file>`.

**Expected behavior**
The `GEval` metric should use AWS Bedrock as default provider.

**Possible solution**
In `deepeval.metrics.util`, add a method for checking if Bedrock was set as default:

```python
def should_use_amazon_bedrock_model():
    value = KEY_FILE_HANDLER.fetch_data(ModelKeyValues.USE_GROK_MODEL)
    return value.lower() == "yes" if value is not None else False
```

and use it in `initialize_model` method:

```python
...
elif should_use_amazon_bedrock_model():
        return AmazonBedrockModel(), True
...
```

**Desktop (please complete the following information):**
 - OS: Windows
 - Version: 3.8.8

penguine-ip:
@leoeiji thanks for closing this, its been released :)

PR #2537, opened 2026-03-06
fix: add Amazon Bedrock model support in initialize_model()

koriyoshi2041:
## Summary

Adds Amazon Bedrock model initialization support to `initialize_model()` in `deepeval/metrics/utils.py`.

**Problem:** The `initialize_model()` function checks for OpenAI, Anthropic, Gemini, Azure, DeepSeek, and other providers, but skips Amazon Bedrock. Users who configure `USE_AWS_BEDROCK_MODEL=YES` get silently routed to the default GPT model instead.

**Fix:**
- Added `should_use_amazon_bedrock_model()` helper (mirrors existing pattern for other providers)
- Added Bedrock check in `initialize_model()` before the default GPT fallback

Related to #2533.

## Test plan

- [x] Verified `should_use_amazon_bedrock_model()` follows the same pattern as `should_use_anthropic_model()`, `should_use_deepseek_model()`, etc.
- [x] `black --check deepeval/metrics/utils.py` passes

### DIFF
--- a/deepeval/metrics/utils.py
+++ b/deepeval/metrics/utils.py
@@ -559,6 +559,13 @@ def should_use_grok_model():
     return value.lower() == "yes" if value is not None else False
 
 
+def should_use_amazon_bedrock_model():
+    if SETTINGS.USE_AWS_BEDROCK_MODEL:
+        return True
+    value = KEY_FILE_HANDLER.fetch_data(ModelKeyValues.USE_AWS_BEDROCK_MODEL)
+    return value.lower() == "yes" if value is not None else False
+
+
 ###############################################
 # LLM
 ###############################################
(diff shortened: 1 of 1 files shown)


A-Vamshi:
Hey @koriyoshi2041, the metric utils file has been recently revised for some additional features, can you make the updates to this PR based on the new changes? Here's the PR that changed the metrics utils file for reference btw: https://github.com/confident-ai/deepeval/pull/2550
Could you make these changes so we can merge this PR? Thank you! :)

koriyoshi2041:
Thanks for the heads up! Rebased onto main to incorporate the changes from #2550. I also updated `should_use_amazon_bedrock_model()` to follow the same `SETTINGS` env-var check pattern introduced in that PR. Please take another look when you get a chance.

A-Vamshi:
Hey @koriyoshi2041, thanks for the changes, would you be open to changing the docs as well? The examples for docs changes should be in the same PR as well, should be a minor change in single file, it's fine if you're not too keen on changing docs as well, totally understandable, please let me know what you think and we'll get this PR merged right away! Thank you :)

A-Vamshi:
Thanks for the PR @koriyoshi2041 :)

PR #2541, opened 2026-03-08
fix: avoid division by zero in GEval score normalization

aerosta:
## Summary

This fixes a division-by-zero edge case in `GEval` score normalization.

When `score_range` has no span, for example `(5, 5)` or `(1, 1)`, the normalization formula divides by `self.score_range_span`, which raises `ZeroDivisionError`. This affects both normalization paths in `g_eval.py`.

## Changes

- added a zero-span guard to both normalization branches
- when `score_range_span == 0`, the code now falls back to `int(g_score)` instead of dividing by zero

## Tests

Added `tests/test_metrics/test_g_eval_score_normalization.py` to cover:

- standard normalization behavior
- custom score range normalization
- zero-span edge cases such as `(5, 5)` and `(1, 1)`
- strict mode behavior
- `get_score_range` utility behavior

```bash
python3.11 -m pytest tests/test_metrics/test_g_eval_score_normalization.py -v

### DIFF
--- a/tests/test_metrics/test_g_eval_score_normalization.py
+++ b/tests/test_metrics/test_g_eval_score_normalization.py
@@ -1,0 +1,85 @@
+"""Tests for GEval score normalization with zero-span score ranges."""
+
+from unittest.mock import patch
+
+from deepeval.metrics import GEval
+from deepeval.metrics.g_eval.utils import Rubric, get_score_range
+from deepeval.models import DeepEvalBaseLLM
+from deepeval.test_case import LLMTestCase, LLMTestCaseParams
+
+
+class _StubLLM(DeepEvalBaseLLM):
+    def load_model(self):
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: tests/test_metrics/test_g_eval_score_normalization.py [hunk 1]
```diff
@@ -0,0 +1,85 @@
+"""Tests for GEval score normalization with zero-span score ranges."""
```
A-Vamshi: Hey @aerosta, this PR is good, was just wondering if you could move this test inside the `tests/test_metrics/test_g_eval_metric.py` file? Just append these tests at the bottom please, thank you :)

PS: Do you think this problem exists for the `ConversationalGEval` metric as well? Could you see if that's the case and help with the changes there as well? Thank you!

PR #2549, opened 2026-03-11
fix: make top_logprobs configurable in ConversationalGEval

### LINKED ISSUE #2548: ConversationalGEval missing configurable top_logprobs (regression from #1261)
# Describe the bug

PR #1261 made top_logprobs configurable in GEval, but the same change was not applied to ConversationalGEval. The value is still hardcoded to 20 in two places (lines 270 and 340), and the constructor is missing the top_logprobs parameter.

# To Reproduce

```python
from deepeval.metrics import ConversationalGEval

metric = ConversationalGEval(
    name="test",
    criteria="...",
    evaluation_steps=["..."],
    model="gpt-5.4",  # supports logprobs but only top_logprobs <= 5
)
# Raises 400: "Invalid value for 'top_logprobs': must be less than or equal to 5."
# No way to override — ConversationalGEval doesn't accept top_logprobs param.
```

# Expected behavior

ConversationalGEval should accept top_logprobs as a constructor parameter (like GEval and MultimodalGEval already do) and use self.top_logprobs instead of the hardcoded 20.

# Additional context

- GEval was fixed in PR #1261 (merged Jan 10, 2025)
- MultimodalGEval also has the configurable parameter
- ConversationalGEval is the only G-Eval variant that was missed
- Hardcoded top_logprobs=20 appears on line 270 (async) and line 340 (sync)
- deepeval version: 3.7.x

SzymonCogiel:
Fixes #2548
 
PR #1261 made `top_logprobs` configurable in `GEval`, but the same change was not applied to `ConversationalGEval`. This PR mirrors that fix:
 
- Add `top_logprobs` parameter to `ConversationalGEval.__init__` (default: 20)
- Replace hardcoded `top_logprobs=20` with `self.top_logprobs` in both sync and async paths (lines 270 and 340)

### DIFF
--- a/deepeval/metrics/conversational_g_eval/conversational_g_eval.py
+++ b/deepeval/metrics/conversational_g_eval/conversational_g_eval.py
@@ -48,6 +48,7 @@ class ConversationalGEval(BaseConversationalMetric):
         evaluation_steps: Optional[List[str]] = None,
         model: Optional[Union[str, DeepEvalBaseLLM]] = None,
         threshold: float = 0.5,
+        top_logprobs: int = 20,
         rubric: Optional[List[Rubric]] = None,
         async_mode: bool = True,
         strict_mode: bool = False,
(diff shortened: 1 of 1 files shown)


A-Vamshi:
Thanks for the PR @SzymonCogiel :)

PR #2552, opened 2026-03-13
fix cache loading

p-constant:
disable loading from cache metric data when score is None + update pbar while loading from cache


penguine-ip:
@p-constant thank you :)

PR #2558, opened 2026-03-18
Fix NoneType crash in trimAndLoadJson when LLM returns None

joaquinhuigomez:
## Summary

Fixes a `NoneType` crash in `trimAndLoadJson()` when the evaluation model returns `None` instead of a string response.

Without this guard, any metric that calls `trimAndLoadJson` — faithfulness, answer relevancy, contextual precision, etc. — throws an unhandled `TypeError` on `None` input. The fix adds an early return with a clear error message at the top of both `trimAndLoadJson` (in `metrics/utils.py`) and `trimAndLoadJson` (in `dataset/utils.py`).

Includes 7 test cases covering `None`, empty string, whitespace, and valid JSON inputs.


joaquinhuigomez:
CI note: The lint failure (101 files would be reformatted by black) and the Confident Tests failure (63 tests failing due to missing `CONFIDENT_API_KEY` env var) are both pre-existing issues unrelated to this PR. My change only touches `deepeval/utils.py` to add a None check in `trimAndLoadJson`.

deepeval/metrics/utils.py
```diff
@@ -391,6 +391,12 @@ def trimAndLoadJson(
     input_string: str,
     metric: Optional[BaseMetric] = None,
 ) -> Any:
+    if input_string is None:
+        error_str = "Evaluation LLM returned None instead of valid JSON. Please use a better evaluation model."
```
A-Vamshi: Can we slightly change the error message to `Evaluation model returned ...`? Using Evaluation LLM and evaluation model in the same message could be confusing to the users. Same with the above change, please change it to `Evaluation model`

tests/test_core/test_trim_and_load_json.py
```diff
@@ -0,0 +1,43 @@
+"""Tests for trimAndLoadJson None-guard (issue #2554)."""
```
A-Vamshi, line 1: can we move this file into `tests/test_core/` instead? I believe that should be a good place for this test to go, it can stay in the root of test_core folder, thank you :)

A-Vamshi (commented):
Hey @joaquinhuigomez, the PR looks good overall, just added a few minor comments, can you please resolve them? Thank you!

deepeval/metrics/utils.py
joaquinhuigomez: Done — updated the error message to use "Evaluation model" consistently. See commit 5da8400.

tests/test_core/test_trim_and_load_json.py
joaquinhuigomez: Done — moved to tests/test_core/test_trim_and_load_json.py. See commit 5da8400.

deepeval/metrics/utils.py
joaquinhuigomez: Updated in 2214dde — changed all instances of "Evaluation LLM" to "Evaluation model" in both metrics/utils.py and dataset/utils.py.

tests/test_core/test_trim_and_load_json.py
joaquinhuigomez: Moved in 2214dde — test file now lives at tests/test_core/test_trim_and_load_json.py.

joaquinhuigomez:
Hi @A-Vamshi, I've addressed both review comments — renamed to "Evaluation model" and moved the test file to tests/test_core/. Could you take another look when you get a chance? Thanks\!

joaquinhuigomez:
Going to close this for now — CI issues are making it hard to validate.

Issue #1147, opened 2024-11-09
Asynchronous test runs are sometimes not completed correctly


jmaczan:
**Describe the bug**
When running `evaluate()` with `run_async=True`, sometimes tests are not completed (so any job/task/pipeline that relies on exit code will fail). Results are neither printed nor emitted. The evaluation is essentialy stuck after running the last test case. It is not always the case, without any regularity, so it looks like a race condition. Likely there is issue with async stuff in evaluate.py file - `a_execute_test_cases()`, ` get_or_create_event_loop()` or `loop.run_until_complete`. 

It might be possible that `await asyncio.sleep(throttle_value)` leads to semaphore being stuck or something. I haven't debug it except a brief static code analysis, though.

**To Reproduce**
1. Create a test cases list
2. Run them using `evaluate()` with `run_async=True`
3. All tests are executed asynchronously and that's fine
4. Results are not printed, saved, etc. It is essentialy stuck after running the last test case

**Expected behavior**
Tests always end with printing either results or errors. They should never last infinitely

**Screenshots**

**Desktop (please complete the following information):**
 - OS: [e.g. iOS] Windows 11

penguine-ip:
@jmaczan I never encountered this issue, can you give us something to reproduce it? For example, number of test cases, metric you're using, etc.

rjiangnju:
@penguine-ip ,
I think I met similar issues before. I have about 130 test cases, the metric I use are:
Hallucination, Answer Relevancy, Faithfulness, Contextual Precision, Contextual Recall, Contextual Relevancy

with run_async=False, it finished without any issue

jmaczan:
@penguine-ip I have about 20 test cases, I use literally the same set of metrics as @rjiangnju. `run_async=False` solves the issue, but to me it's a no-go since I run each case for multiple times to remove top and bottom results and draw median value to get more stabilized metrics

threeteck:
I have the same issue, my metrics are GEval and Answer Relevancy. I run `evaluate` method multiple times (about 100 test cases per run) in my script and sometimes it gets stuck near the end. Running with run_async=False works fine, but its much slower. I also tried to split the number of test cases into more groups, but even with just 3 test cases per `evaluate` it still gets stuck randomly.

threeteck:
[...]

PR #2559, opened 2026-03-18
fix: prevent trace metrics from being shared across concurrent tasks

aerosta:
## Summary

`_a_evaluate_traces` dispatches each trace as a concurrent `asyncio.create_task` but passes the same `trace_metrics` list to every task. Because metrics write to `self.score`, `self.reason`, and `self.success` during evaluation, concurrent tasks interfere with each other and produce inconsistent results depending on timing.

Fixed by calling `copy_metrics(trace_metrics)` for each trace before dispatch — the same thing the test-case path already does at five other call sites in `execute.py`.

## Context

The trace path introduced in #2189 was the only async fan-out point missing the copy. This issue is likely related to #1147.

## Changes

- `deepeval/evaluate/execute.py`: copy `trace_metrics` per trace in `_a_evaluate_traces` before creating each task
- `tests/test_core/test_evaluation/test_async_trace_metric_isolation.py`: regression test using barrier-synchronized metrics to force concurrent execution and verify per-trace isolation

## Note

The sync equivalent (`execute_agentic_test_cases_from_loop`, line 2334) also assigns `trace_metrics` directly without copying. This is currently safe because the loop is sequential, but should be reviewed if that path becomes concurrent.

### DIFF
--- a/deepeval/evaluate/execute.py
+++ b/deepeval/evaluate/execute.py
@@ -2990,6 +2990,9 @@ async def _a_evaluate_traces(
                     trace.uuid,
                 )
             continue
+        copied_trace_metrics: Optional[List[BaseMetric]] = None
+        if trace_metrics:
+            copied_trace_metrics = copy_metrics(trace_metrics)
         with capture_evaluation_run("golden"):
             task = execute_evals_with_semaphore(
                 func=_a_execute_agentic_test_case,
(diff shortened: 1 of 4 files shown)


A-Vamshi:
Thanks for the PR! Merged!

aerosta:
> Thanks for the PR! Merged!

Thank you @A-Vamshi

PR #2563, opened 2026-03-19
pydantic-ai: read thread_id, name, metadata from current_trace_context in SpanInterceptor.on_start

Oluwa-nifemi:
## Problem

I needed per-request trace metadata (e.g. `thread_id`, `report_id`) on traces coming from a pydantic-ai powered service. The naive approach — creating a new `ConfidentInstrumentationSettings` per request — immediately hits:

```
Overriding of current TracerProvider is not allowed
```

because `ConfidentInstrumentationSettings.__init__` calls `set_tracer_provider()` on every construction, and OpenTelemetry only allows this once globally.

## What I tried next

The docs show `update_current_trace` as the way to overwrite trace attributes at runtime:
https://www.confident-ai.com/docs/integrations/third-party/openai-agents#overwrite-trace-attributes

So I tried wrapping my agent call with `with trace(thread_id=..., metadata=..., name=...)` and calling `update_current_trace`. This _does_ populate `current_trace_context` correctly. But nothing showed up in Confident AI.

## Root cause

Digging into the source, I found that `SpanInterceptor.on_start` already fetches `current_trace_context`:

```python
_current_trace_context = current_trace_context.get()
if _current_trace_context and isinstance(_current_trace_context, Trace):
    _otel_trace_id = span.get_span_context().trace_id
    _current_trace_context.uuid = to_hex_string(_otel_trace_id, 32)
```

But it only uses it to sync the trace UUID. The actual span attributes — `thread_id`, `metadata`, `name` — are read exclusively from `self.settings`:

```python
if self.settings.thread_id:
    span.set_attribute("confident.trace.thread_id", self.settings.thread_id)
```

So `with trace()` / `update_current_trace` has zero effect on what gets exported to Confident AI via OTEL. This is in contrast to the OpenAI Agents integration (`callback_handler.py`) which _does_ populate `current_trace_context` at trace start and reads from it properly.

## Fix

Since `_current_trace_context` is already fetched, just also read `thread_id`, `name`, and `metadata` from it — with fallback to `self.settings` for backwards compatibility, and a merge for `metadata` so static settings-level metadata and per-request metadata can coexist:

```python
_thread_id = (
    _current_trace_context.thread_id if _current_trace_context else None
) or self.settings.thread_id

_name = (
    _current_trace_context.name if _current_trace_context else None
) or self.settings.name

_metadata = {
    **(self.settings.metadata or {}),
    **(_current_trace_context.metadata or {} if _current_trace_context else {}),
}
```

Users can now do:

```python
with trace(thread_id="user-session-abc", metadata={"report_id": "xyz"}, name="my_agent"):
    result = await agent.run(...)
```

...and those values will appear on the trace in Confident AI — no need to recreate `ConfidentInstrumentationSettings` per request.

## Tests

Added unit tests for `SpanInterceptor.on_start` in `tests/test_integrations/test_pydanticai/test_span_interceptor.py` covering:

- Fallback to `self.settings` when `current_trace_context` is `None`
- Context values override settings for scalar fields (`thread_id`, `name`)
- `metadata` is merged (settings as base, context on top)
- Attributes from context are correctly stamped on the span inside a `with trace()` block
- No attributes are set when both context and settings are empty

No real LLM or Confident AI API key required.

## Follow-up suggestion (not in this diff)

`trace()`'s `finally` block resets `current_llm_context` and `current_agent_context` but never resets `current_trace_context`. For the current use case (one trace per async task) this is fine since each asyncio task gets its own context copy at creation time. But an explicit reset in the `finally` block would be cleaner and safer for nested usage:

Reference file: https://github.com/confident-ai/deepeval/blob/main/deepeval/tracing/trace_context.py#L107

```python
finally:
    if started_new_trace:
        trace_manager.end_trace(current_trace.uuid)
    current_trace_context.set(None)  # or reset(token) for proper nesting support
    current_llm_context.set(LlmSpanContext())
    current_agent_context.set(AgentSpanContext())
```

### DIFF
--- a/tests/test_integrations/test_pydanticai/test_span_interceptor.py
+++ b/tests/test_integrations/test_pydanticai/test_span_interceptor.py
@@ -1,22 +1,9 @@
-"""
-Unit tests for SpanInterceptor.on_start — verifies that per-request trace
-attributes (thread_id, name, metadata) are read from current_trace_context
-when set, with fallback to self.settings, and that metadata is merged.
-
-No real LLM, OTEL exporter, or Confident AI key required.
-"""
-
 import json
 from unittest.mock import MagicMock
 
 from deepeval.integrations.pydantic_ai.instrumentator import SpanInterceptor
 from deepeval.tracing.context import current_trace_context
 from deepeval.tracing.trace_context import trace
-
-# ---------------------------------------------------------------------------
-# Helpers
-# ---------------------------------------------------------------------------
-
 
 def _make_mock_span():
     """Return a mock span that records set_attribute calls."""
@@ -55,14 +42,7 @@ def _make_settings(**kwargs):
     return settings
 
 
-# ---------------------------------------------------------------------------
-# Tests
-# ---------------------------------------------------------------------------
-
-
 class TestSpanInterceptorOnStart:
-    """Unit tests for SpanInterceptor.on_start attribute resolution."""
-
     def test_uses_settings_when_no_trace_context(self):
         """Falls back to self.settings when current_trace_context is None."""
         token = current_trace_context.set(None)


Oluwa-nifemi:
Thank you for the feedback 

> user_id might be expected to behave the same way by callers who set it per-request.

Yeah that's fair. I'll update the PR to include user_id as well

PR #2569, opened 2026-03-22
fix: guard trimAndLoadJson against None input

### LINKED ISSUE #2554: 'NoneType' object has no attribute 'find'
While running evaluation, I often run into this error.

`'NoneType' object has no attribute 'find'`

Seems similar to my previous issue: https://github.com/confident-ai/deepeval/issues/2327 (in the sense that deepeval does not check that a result might be None)

my setup is the same as described there.

The culprit seems to be https://github.com/confident-ai/deepeval/blob/053c3fee948b3690e56e35c4b99fb99c63382e76/deepeval/metrics/utils.py#L394

themavik:
## Summary
Fixes #2554.
**Root cause:** `trimAndLoadJson` calls `.find()` on its `input_string` parameter without checking for `None`. When the evaluation LLM returns no output, callers pass `None` and the function crashes with `AttributeError: 'NoneType' object has no attribute 'find'`.
**Fix:** Added a None guard at the top of the function that raises a clear `ValueError` (and sets `metric.error`) instead of crashing.

## Changes
- `deepeval/metrics/utils.py`: None check before string operations in `trimAndLoadJson`
- `tests/test_core/test_utils.py`: tests for None input, metric error propagation, valid JSON, and invalid JSON

## Testing
- Added test for None input raising ValueError with descriptive message
- Added test confirming metric.error is set when metric is provided
- Added test for valid JSON extraction (existing behavior preserved)
- Added test for invalid JSON raising ValueError

### DIFF
--- a/deepeval/metrics/utils.py
+++ b/deepeval/metrics/utils.py
@@ -391,6 +391,12 @@ def trimAndLoadJson(
     input_string: str,
     metric: Optional[BaseMetric] = None,
 ) -> Any:
+    if input_string is None:
+        error_str = "Evaluation LLM returned None. Please use a better evaluation model."
+        if metric is not None:
+            metric.error = error_str
+        raise ValueError(error_str)
+
     start = input_string.find("{")
     end = input_string.rfind("}") + 1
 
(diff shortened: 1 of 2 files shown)


themavik (commented):
Clear failure mode vs None and the metric.error hook is helpful. nit: trimAndLoadJson is still annotated as input_string: str while accepting None at runtime — str | None (or overloads) would align typing with the new guard.

PR #2570, opened 2026-03-22
Add test_case_id support to ConfidentSpanExporter

AlexMaggioni:
## Summary
- Adds `confident.trace.test_case_id` span attribute extraction to `ConfidentSpanExporter`, enabling `testCaseId` propagation for the OTel tracing path
- Without this, traces sent via the OTel exporter always have `testCaseId: null`, making it impossible to link traces to test cases in ConfidentAI evaluation runs (AI Connections)

## Changes
- Added `trace_test_case_id` field to `BaseSpanWrapper` dataclass
- Extract `confident.trace.test_case_id` from span attributes in `__set_trace_attributes`
- Set `test_case_id` on the `Trace` object in `_set_current_trace_attributes_from_base_span_wrapper`

Follows the exact same pattern as all other `confident.trace.*` attributes (`thread_id`, `user_id`, `environment`, etc.).

## Context
The native `@observe` path can set `test_case_id` via `update_current_trace()`, but the OTel exporter path had no equivalent. This is needed for applications using OTel-based instrumentation (e.g. `pydantic_ai.Agent.instrument_all()`) that fan spans out to multiple backends via a shared `TracerProvider`.

## Open question
Both the `@observe` and OTel paths now include `testCaseId` in the payload sent to `POST /v1/traces`, but looking at the [API docs](https://www.confident-ai.com/docs/api-reference/tracing/create-trace), `testCaseId` doesn't appear to be a supported field on that endpoint. In the UI, I can see the conversational thread and the test cases being executed, but the traces are not linked to the test cases, which is what I'd expect to happen. Is `testCaseId` on `POST /v1/traces` currently supported by the backend, or is this a feature that is planned to come?

## Test plan
- [x] Set `confident.trace.test_case_id` attribute on a root span, export through `ConfidentSpanExporter`, confirm `TraceApi` payload includes `"testCaseId": "<value>"`

### DIFF
--- a/deepeval/tracing/otel/exporter.py
+++ b/deepeval/tracing/otel/exporter.py
@@ -245,9 +245,13 @@ class ConfidentSpanExporter(SpanExporter):
             )
 
         # set the trace test case id and turn id
-        if base_span_wrapper.trace_test_case_id:
+        if base_span_wrapper.trace_test_case_id and isinstance(
+            base_span_wrapper.trace_test_case_id, str
+        ):
             current_trace.test_case_id = base_span_wrapper.trace_test_case_id
-        if base_span_wrapper.trace_turn_id:
+        if base_span_wrapper.trace_turn_id and isinstance(
+            base_span_wrapper.trace_turn_id, str
+        ):
             current_trace.turn_id = base_span_wrapper.trace_turn_id
 
         # set the trace metric collection


AlexMaggioni:
> Reviewer deleted its comment, but I’m keeping my answer here for reference.

Thanks for the review!

On the `bytes` point, that's actually not a valid OTel attribute type. `span.set_attribute()` only accepts `str`, `bool`, `int`, `float`, or tuples of those, so `bytes` would never make it to the exporter.

You're right that an `int` is technically possible, even if it would be a bit odd since `test_case_id` is meant to be a string identifier. The thing is, all the other `confident.trace.*` string fields in the exporter (`trace_thread_id`, `trace_user_id`, `trace_name`, etc.) use the same `isinstance(..., str)` check and silently drop non-string values too. So this isn't new with our change, but I agree it's not great since the user gets no indication something was ignored.

Maybe worth a follow-up PR to handle this more gracefully across all the string fields? Something like coercing with `str()` and logging a warning instead of silently dropping?

AlexMaggioni:
Hey, it looks like #2576 (merged Mar 27) independently implemented the core feature this PR was targeting (`test_case_id` support in `ConfidentSpanExporter`). A few observations and a plan to rescope:

### What #2576 landed
- `trace_test_case_id` and `trace_turn_id` added to `BaseSpanWrapper`, the OTEL exporter, and all integration instrumentators
- Wired through `TraceApi`, `Trace`, `update_current_trace`, and `context.py`

### What's missing

**1. Missing `isinstance(str)` guards on the new fields**

In the exporter, every other string attribute read from OTEL spans (`trace_name`, `trace_thread_id`, `trace_user_id`, `trace_environment`) is guarded with `isinstance(str)` before being assigned — because `span.attributes.get()` returns `Optional[AttributeValue]` (a union of `str | bool | int | float | Sequence[...]`), not `str`. The two new fields skip this check:

```python
# Current (unguarded):
if base_span_wrapper.trace_test_case_id:
    current_trace.test_case_id = base_span_wrapper.trace_test_case_id
if base_span_wrapper.trace_turn_id:
    current_trace.turn_id = base_span_wrapper.trace_turn_id

# Every other string field in the same file:
if base_span_wrapper.trace_name and isinstance(base_span_wrapper.trace_name, str):
    current_trace.name = base_span_wrapper.trace_name
```

**2. REST API docs don't document `testCaseId` / `turnId`**

Both the `@observe` path and the `ConfidentSpanExporter` (OTEL path) ultimately serialize `testCaseId` and `turnId` and POST them to `api.confident-ai.com/v1/traces`. However, neither field appears in the [REST API reference](https://www.confident-ai.com/docs/api-reference/tracing/create-trace). Are these fields accepted and persisted by the backend?

**3. Is `testCaseId` redundant with `turnId`?**

Now that `turnId` exists, it could cover both single-turn and multi-turn evaluation contexts; a single-turn evaluation is just a conversation with one turn. This would make `testCaseId` unnecessary since `turnId` is strictly more flexible. Are these two fields mapped to different things on the backend, or is there a plan to consolidate? Happy to open a separate PR for this if it makes sense.

---

### Rescoped PR

I'll rescope this PR to add the missing `isinstance(str)` guards for `trace_test_case_id` and `trace_turn_id`, consistent with how all other string attributes are handled in the exporter.

PR #2574, opened 2026-03-25
Fix a couple type annotations in base_metric.py

tbeadle:
These were being assigned to the types rather than having them as type annotations.


penguine-ip:
@tbeadle good catch, thanks!

PR #2611, opened 2026-04-10
Fix Optional field and nested model handling in schema construction

### LINKED ISSUE #2608: Nested Pydantic models stop being recognized after one inheritance layer
## Describe the bug

In `deepeval/prompt/utils.py`, the structured output schema builder detects nested Pydantic models with `BaseModel in annotation.__bases__`. This only works when the field type **directly** subclasses `BaseModel`. If a project uses an intermediate base class (e.g. `class DerivedPayload(BasePayload)` where `BasePayload(BaseModel)`), the check fails and the field falls back to `STRING` type.

## To Reproduce

```python

class BasePayload(BaseModel):
    value: int

class DerivedPayload(BasePayload):
    extra: str

class MyOutput(BaseModel):
    payload: DerivedPayload  # This gets typed as STRING instead of OBJECT

# Because:
print(BaseModel in DerivedPayload.__bases__)  # False
print(BaseModel in DerivedPayload.__mro__)    # True
```

## Root Cause

`deepeval/prompt/utils.py` L237-252:

```python
elif (
    hasattr(annotation, "__bases__")
    and BaseModel in annotation.__bases__
):
    field_type = "OBJECT"
```

## Expected behavior

Any Pydantic `BaseModel` subclass (including indirect inheritance) should be recognized as an `OBJECT` type in generated schemas.

## Suggested Fix

Use `issubclass(annotation, BaseModel)` or check `__mro__` instead of `__bases__`, consistent with how array item models are already handled elsewhere in the same file:

```python
elif (
    hasattr(annotation, "__mro__")
    and issubclass(annotation, BaseModel)
):
    field_type = "OBJECT"
```

### LINKED ISSUE #2609: Optional structured output fields are downgraded to strings in generated schemas
## Bug Description

`construct_output_schema()` in `deepeval/prompt/utils.py` only recognizes bare primitives, bare `list[...]`, and direct `BaseModel` subclasses. Optional structured fields such as `Optional[List[int]]`, `list[int] | None`, or `Optional[ChildModel]` have a `Union` origin, so they miss every structured branch and fall through to the default `field_type = "STRING"`.

The resulting schema advertises those fields as strings, which corrupts prompt tool schemas and downstream JSON-schema generation.

## To Reproduce

```python
from typing import Optional, List

class MyOutput(BaseModel):
    tags: Optional[List[int]] = None     # → STRING (wrong, should be ARRAY)
    child: Optional["ChildModel"] = None  # → STRING (wrong, should be OBJECT)

# Because Optional[X] has origin=Union, it misses the list/BaseModel branches
```

## Root Cause

`deepeval/prompt/utils.py` L168-260:

```python
annotation = field_info.annotation
field_type = "STRING"
origin = get_origin(annotation)
...
elif origin is list:
    ...
elif (
    hasattr(annotation, "__bases__")
    and BaseModel in annotation.__bases__
):
    ...
# Optional[List[int]] has origin=Union, not list → falls through to STRING
```

## Expected behavior

`Optional[List[int]]` should produce an ARRAY-typed field. `Optional[ChildModel]` should produce an OBJECT-typed field.

## Suggested Fix

Normalize `Optional[...]` / `Union[..., None]` annotations before type dispatch:

```python
# Unwrap Optional/Union[X, None]
args = get_args(annotation)
if origin is Union and type(None) in args:
    non_none = [a for a in args if a is not type(None)]
    if len(non_none) == 1:
        annotation = non_none[0]
        origin = get_origin(annotation)
        required = False
# Then proceed with existing dispatch logic
```

SamSi0322:
## Summary

Fixes two bugs in `_process_model()` (`deepeval/prompt/utils.py`) that caused incorrect schema generation for structured outputs.

**Bug 1 — Optional types misidentified as STRING (#2609):**
`Optional[List[int]]` has `get_origin() → Union`, not `list`, so it fell through the type dispatch to the default `field_type = "STRING"`. Fixed by unwrapping `Optional[X]` (`Union[X, None]`) before dispatching.

**Bug 2 — Nested Pydantic models not recognized (#2608):**
`BaseModel in annotation.__bases__` only checks direct parents. A model like `class Derived(BasePayload)` where `BasePayload(BaseModel)` was not recognized. Changed to `BaseModel in annotation.__mro__` which checks the full inheritance chain — matching the pattern already used for array item types on line 208.

## Changes

- Added `Union` import
- Added Optional unwrapping (6 lines) before the type dispatch chain
- Changed `__bases__` → `__mro__` for nested model detection (1 line)

## Test plan

- `Optional[List[int]]` fields now correctly generate ARRAY schema instead of STRING
- `Optional[str]` fields still correctly generate STRING
- Models inheriting from intermediate BaseModel subclasses are now detected as OBJECT

Closes #2609, closes #2608

### DIFF
--- a/deepeval/prompt/utils.py
+++ b/deepeval/prompt/utils.py
@@ -1,7 +1,17 @@
 import re
 import uuid
 from jinja2 import Template
-from typing import Any, Dict, Type, Optional, List, Match, Union, get_origin, get_args
+from typing import (
+    Any,
+    Dict,
+    Type,
+    Optional,
+    List,
+    Match,
+    Union,
+    get_origin,
+    get_args,
+)
 from pydantic import BaseModel, create_model
 
 from deepeval.prompt.api import (
@@ -242,10 +252,7 @@ def _process_model(
                 )
                 fields.append(item_field)
             continue
-        elif (
-            hasattr(annotation, "__mro__")
-            and BaseModel in annotation.__mro__
-        ):
+        elif hasattr(annotation, "__mro__") and BaseModel in annotation.__mro__:
             field_type = "OBJECT"
             parent_field = OutputSchemaField(
                 id=field_id,
--- a/deepeval/prompt/utils.py
+++ b/deepeval/prompt/utils.py
@@ -1,7 +1,7 @@
 import re
 import uuid
 from jinja2 import Template
-from typing import Any, Dict, Type, Optional, List, Match, get_origin, get_args
+from typing import Any, Dict, Type, Optional, List, Match, Union, get_origin, get_args
 from pydantic import BaseModel, create_model
 
 from deepeval.prompt.api import (
@@ -175,7 +175,15 @@ def _process_model(
         field_id = str(uuid.uuid4())
         annotation = field_info.annotation
         field_type = "STRING"
+
+        # Unwrap Optional[X] (Union[X, None]) to its inner type
         origin = get_origin(annotation)
+        if origin is Union:
+            args = [a for a in get_args(annotation) if a is not type(None)]
+            if len(args) == 1:
+                annotation = args[0]
+                origin = get_origin(annotation)
+
         if annotation == str:
             field_type = "STRING"
         elif annotation == int:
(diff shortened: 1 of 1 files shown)


penguine-ip:
thanks @SamSi0322 !

PR #2619, opened 2026-04-17
fix: prevent duplicate test-case rows on pytest retry

### LINKED ISSUE #1992: Tests reported twice with `pytest.mark.flaky`
Hi, thanks for this great library.

I'm running into an issue when marking tests with `pytest.mark.flaky`. In cases where tests pass e.g. on the second try, the first attempt is logged on the `confident-ai.com` dashboard as a failure, and the second attempt is logged as a pass. 

This leads to many more test cases being shown on the dashboard than were actually run in practice.

Is there a better way to handle `retry` logic other than using `mark.flaky`? Should this be handled internally within the test, e.g. catching `AssertionErrors` N times, and then throwing when N is exceeded?

## Minimal example

This won't work "out-of-the-box", but should show how I'm constructing the test cases.

```python
@pytest.mark.parametrize(
    "query",
    # big list of queries
    [...]
)
@pytest.mark.flaky(reruns=5)
def test_with_flaky(query: str):
    # Assume that this takes the query, makes a generation with the agent
    #  and then returns a `deepeval.LLMTestCase` object
    test_case = create_basic_test_case(query)

    metric = GEval(
        name="flaky test",
        criteria=(
            "..."
        ),
        evaluation_params=[LLMTestCaseParams.INPUT, LLMTestCaseParams.ACTUAL_OUTPUT],
    )

    # Maybe fails the first time, but works the second
    assert_test(test_case, [metric])
```

amitkojha05:
## Problem

When a test is marked with `pytest.mark.flaky` or `pytest-rerunfailures`,`assert_test()` is called again from scratch on every retry attempt.Each call flows through:

assert_test() → execute_test_cases() → update_test_run() → add_test_case()

`add_test_case()` blindly appended on every call with no duplicate check.
A test that passed on the 3rd attempt would appear on the Confident AI dashboard as **2 failures + 1 pass** instead of just **1 pass**.

The same unconditional append also double-counted `evaluation_cost` — a test costing `0.01` on attempt 1 and `0.02` on attempt 2 accumulated `0.03` on the run total instead of the correct `0.02`.

## Root cause

`TestRun.add_test_case()` in `deepeval/test_run/test_run.py` — no deduplication by test-case name before appending to `self.test_cases` or `self.conversational_test_cases`.

## Fix

One method changed, one file touched: `deepeval/test_run/test_run.py`.

`add_test_case()` now scans the target list for an existing entry with the same `name`. On a match it replaces that slot (latest attempt wins) and backs the replaced attempt's `evaluation_cost` out of the run total before applying the new one. New test cases still append as before via Python's `for/else`.

```python
replaced_cost: Union[float, None] = None
for i, existing in enumerate(target_list):
    if existing.name == api_test_case.name:
        replaced_cost = existing.evaluation_cost
        target_list[i] = api_test_case
        break
else:
    target_list.append(api_test_case)

if replaced_cost is not None and self.evaluation_cost is not None:
    self.evaluation_cost -= replaced_cost
```

## Before / after

```python
run.add_test_case(LLMApiTestCase(name="test_foo", ..., evaluationCost=0.01, success=False))  # attempt 1
run.add_test_case(LLMApiTestCase(name="test_foo", ..., evaluationCost=0.02, success=True))   # retry

# BEFORE
len(run.test_cases)  # 2    ← both attempts appear as separate dashboard rows
run.evaluation_cost  # 0.03 ← cost double-counted

# AFTER
len(run.test_cases)  # 1    ← only the final result reported
run.evaluation_cost  # 0.02 ← correct
```

## Tests

Added `tests/test_core/test_run/test_test_run_retry_overwrite.py` — 10 tests, all passing on Python 3.10 and 3.12:

- LLM: same name twice → one row, latest `success` wins
- LLM: two different names → two rows (existing behaviour unchanged)
- LLM: cost not double-counted on replace
- LLM: retry with `None` cost backs out previous cost correctly
- LLM: first attempt `None` cost, retry sets cost correctly
- LLM: three retries → one row, only last cost counted
- LLM: mixed retried + unique test → correct total cost
- Conversational: same name twice → one row
- Conversational: cost not double-counted
- Conversational: two different names → two rows

## Caveat

Deduplication keys on `name`. If two genuinely distinct test cases share the same display name in one run, the second would silently overwrite the first. Pytest node IDs are unique by default so this shouldn't occur in practice — flagging it in case maintainers prefer a stricter key (e.g. `nodeid`) in future.

## Checklist
- [x] Closes #1992
- [x] Changes limited to `deepeval/test_run/test_run.py` + new test file
- [x] 10/10 tests passing locally (Python 3.10 + 3.12)

### DIFF
--- a/deepeval/test_run/test_run.py
+++ b/deepeval/test_run/test_run.py
@@ -169,7 +169,7 @@ class TestRun(BaseModel):
     def add_test_case(
         self, api_test_case: Union[LLMApiTestCase, ConversationalApiTestCase]
     ):
-        
+
         target_list = (
             self.conversational_test_cases
             if isinstance(api_test_case, ConversationalApiTestCase)
--- a/deepeval/test_run/test_run.py
+++ b/deepeval/test_run/test_run.py
@@ -169,10 +169,26 @@ class TestRun(BaseModel):
     def add_test_case(
         self, api_test_case: Union[LLMApiTestCase, ConversationalApiTestCase]
     ):
-        if isinstance(api_test_case, ConversationalApiTestCase):
-            self.conversational_test_cases.append(api_test_case)
+        
+        target_list = (
+            self.conversational_test_cases
+            if isinstance(api_test_case, ConversationalApiTestCase)
+            else self.test_cases
+        )
+
+        replaced_cost: Union[float, None] = None
+        for i, existing in enumerate(target_list):
+            if existing.name == api_test_case.name:
+                replaced_cost = existing.evaluation_cost
+                target_list[i] = api_test_case
+                break
         else:
-            self.test_cases.append(api_test_case)
+            target_list.append(api_test_case)
+
+        # Back out the replaced attempt's cost before adding the new one,
+        # so a retry doesn't double-count its evaluation_cost on the run total.
+        if replaced_cost is not None and self.evaluation_cost is not None:
+            self.evaluation_cost -= replaced_cost
 
         if api_test_case.evaluation_cost is not None:
             if self.evaluation_cost is None:
--- a/deepeval/test_run/test_run.py
+++ b/deepeval/test_run/test_run.py
@@ -170,10 +170,25 @@ class TestRun(BaseModel):
     def add_test_case(
         self, api_test_case: Union[LLMApiTestCase, ConversationalApiTestCase]
     ):
-        if isinstance(api_test_case, ConversationalApiTestCase):
-            self.conversational_test_cases.append(api_test_case)
+        target_list = (
+            self.conversational_test_cases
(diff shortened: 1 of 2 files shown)


amitkojha05:
@penguine-ip Please review this PR

penguine-ip:
Hey @amitkojha05 thanks for the PR. Is it possible to make the changes more localized? Right now it is hard to merge as there is a lot of changes and I believe the change can be made simpler. Thank you!

amitkojha05:
Hi @penguine-ip , Thanks for the review and feedback. I’ll work on localizing the changes and simplify the implementation so it’s easier to review and merge. I’ll push an updated version shortly.

Issue #2643, opened 2026-05-01
Feature: AgentLoopDetectionMetric — detect infinite loops and cyclical tool-call patterns in agent traces


Jeel3011:
## Problem

DeepEval's current agentic metrics (`TaskCompletionMetric`, `StepEfficiencyMetric`, 
`PlanAdherenceMetric`, etc.) all evaluate the **quality** of a completed agent run. 
None of them detect a critical class of production failure: **the agent getting stuck 
in a loop before it ever finishes**.

In practice, ReAct-style agents and LangGraph agents fail silently in three ways:

1. **Repeated tool calls** — the agent calls the same tool with identical (or nearly 
   identical) arguments multiple times, making no progress
2. **Circular reasoning** — the LLM produces the same reasoning step in a cycle, 
   spinning context and burning tokens without advancing toward the goal
3. **Stagnation spirals** — each iteration adds context but produces no measurable 
   progress toward task completion, leading to runaway cost

These failures are **invisible to existing metrics** because:
- `StepEfficiencyMetric` penalizes unnecessary steps, but only scores a *completed* run
- `TaskCompletionMetric` requires the agent to have reached a final output to evaluate
- Neither metric examines the *shape* of the trace for cyclical patterns

This is a real production problem. I've encountered it building episodic evaluation 
frameworks for LangGraph agents (see: [AGeval](https://pypi.org/project/ageval/)), 
and it's one of the most common ways agents silently fail and rack up cost.

## Proposed Solution: `AgentLoopDetectionMetric`

A new trace-only agentic metric that analyzes the tool call sequence and LLM 
reasoning steps for loop/cycle patterns.

**Scoring logic (three sub-signals, combined into a 0–1 score):**

| Sub-signal | What it detects | Method |
|---|---|---|
| Tool call repetition | Same tool + same args called N times | Hash (tool_name, args) → count duplicates |
| Reasoning stagnation | LLM output similarity across steps | Sliding window cosine similarity or n-gram overlap |
| Call graph cycles | Circular dependency in tool call sequence | DFS cycle detection on call DAG |

**Score interpretation:**
- `1.0` = No loops detected, agent execution was linear and progressed
- `0.0` = Severe looping detected (agent clearly stuck)
- Intermediate = partial repetition detected with `reason` explaining which pattern was found

**API design (mirrors existing agentic metrics):**

```python
from deepeval.metrics import AgentLoopDetectionMetric
from deepeval.tracing import observe
[...]

PR #2645, opened 2026-05-01
feat: add AgentLoopDetectionMetric — deterministic loop detection for agent traces

Jeel3011:
## Summary

Adds `AgentLoopDetectionMetric` — a **fully deterministic** (no LLM / API key required) metric that detects infinite loops and cyclical execution patterns in agent traces. It analyzes three independent sub-signals and returns a weighted score from 0.0 (severe looping) to 1.0 (clean execution).

This metric is designed to run in production at **zero cost and zero latency** — every sub-signal is computed with hashing, set operations, or sequence comparison.

---

## Sub-signals

| Signal | Weight | What it detects |
|---|---|---|
| **Tool Call Repetition** | 40% | Same tool called with identical `(name, args)` ≥ N times |
| **Reasoning Stagnation** | 35% | Consecutive LLM outputs that are structurally identical or reordered |
| **Call Graph Cycles** | 25% | A span that recursively invokes itself (same type + name + input) |

### Tool Call Repetition
Hashes each tool call by `(name, sorted_args)`. Score degrades at `repetition_threshold` (0.5) and at `2× threshold` (0.0). Different arguments = different calls = not penalized.

### Reasoning Stagnation
Uses **two complementary signals** and takes the maximum:
- **Bigram Jaccard** — catches literal repetition after stop-word removal
- **`SequenceMatcher` ratio** (`difflib`) — catches reordered-but-identical phrasing

Outputs shorter than 20 meaningful words are skipped (Jaccard is noisy at that scale).

### Call Graph Cycles
DFS on the nested `children` tree. Labels each span as `type:name:input_hash`. A cycle is flagged when the same label appears twice on the same root-to-leaf ancestry path. Including `input_hash` prevents false positives when two different agents share a name but receive different inputs.

---

## Files changed

| File | Change |
|---|---|
| `deepeval/metrics/agent_loop_detection/agent_loop_detection.py` | Core metric implementation |
| `deepeval/metrics/agent_loop_detection/__init__.py` | Module export (unchanged from initial) |
| `tests/test_agent_loop_detection.py` | 10 passing tests, zero API key required |
| `docs/content/docs/(agentic)/metrics-agent-loop-detection.mdx` | Full doc page (Usage, How It Works, Limitations) |
| `docs/content/docs/(agentic)/meta.json` | Registered in sidebar |

### Why no `model` parameter?
Deterministic by design. Accepting `model` would be misleading since no sub-signal uses an LLM. A future LLM-as-judge stagnation mode could be added behind a feature flag.

### Cycle detection uses `type:name:input_hash`, not span UUIDs
`create_nested_spans_dict` strips UUIDs from the trace dict, so exact span identity is unavailable. The `input_hash` heuristic is the best available — true recursive loops pass the same input back, so the hash matches. Two same-name spans with different inputs are correctly treated as distinct.

### Stagnation will miss semantically-different wording
"Search for Paris weather" vs "Look up the forecast in Paris" won't be caught. Fixing this would require an LLM-as-judge, sacrificing the zero-cost property.

---

## Test matrix (10 tests, all pass)

| # | Test | Validates |
|---|---|---|
| 1 | `test_clean_trace_passes` | Clean agent → 1.0 |
| 2 | `test_repeated_tool_calls_detected` | 4× identical calls → ≤ 0.5 |
| 3 | `test_no_trace_returns_zero` | Missing trace → 0.0 + reason |
| 4 | `test_reasoning_stagnation_detected` | Identical outputs → stagnation flagged |
| 5 | `test_disable_tool_repetition_check` | Disabled check → score unaffected |
| 6 | `test_score_combines_with_correct_weights` | Weight normalization math |
| 7 | `test_call_graph_cycle_detected` | True recursive loop → 0.0 |
| 8 | `test_sequential_same_name_not_a_cycle` | Siblings ≠ cycle (no false positive) |
| 9 | `test_same_name_different_input_not_a_cycle` | Same name, different input ≠ cycle |
| 10 | `test_reordered_stagnation_detected` | Reordered phrasing caught by SequenceMatcher |

---

## CI note

The `test_core/` collection error (`async_app.py:67` instantiating `AnswerRelevancyMetric()` at module level without an API key) is a pre-existing issue on `main` (commit `90f398afc`). This PR does not modify that file.

### DIFF
--- a/deepeval/metrics/__init__.py
+++ b/deepeval/metrics/__init__.py
@@ -35,6 +35,7 @@ from .plan_quality.plan_quality import PlanQualityMetric
 from .tool_use.tool_use import ToolUseMetric
 from .goal_accuracy.goal_accuracy import GoalAccuracyMetric
 from .argument_correctness.argument_correctness import ArgumentCorrectnessMetric
+from .agent_loop_detection.agent_loop_detection import AgentLoopDetectionMetric
 from .mcp.mcp_task_completion import MCPTaskCompletionMetric
 from .mcp.multi_turn_mcp_use_metric import MultiTurnMCPUseMetric
 from .mcp_use_metric.mcp_use_metric import MCPUseMetric
@@ -116,6 +117,7 @@ __all__ = [
     "PlanQualityMetric",
     "ToolUseMetric",
     "GoalAccuracyMetric",
+    "AgentLoopDetectionMetric",
     # Conversational metrics
     "TurnRelevancyMetric",
     "ConversationCompletenessMetric",
--- a/deepeval/metrics/agent_loop_detection/__init__.py
+++ b/deepeval/metrics/agent_loop_detection/__init__.py
@@ -1,0 +1,1 @@
+from .agent_loop_detection import AgentLoopDetectionMetric
--- a/deepeval/metrics/agent_loop_detection/agent_loop_detection.py
+++ b/deepeval/metrics/agent_loop_detection/agent_loop_detection.py
@@ -1,0 +1,394 @@
+import json
+from typing import Optional, List, Union, Tuple, Dict
+
+from deepeval.test_case import LLMTestCase, SingleTurnParams
+from deepeval.metrics import BaseMetric
+from deepeval.models import DeepEvalBaseLLM
+from deepeval.metrics.utils import (
+    initialize_model,
+    construct_verbose_logs,
+    check_llm_test_case_params,
+)
+from deepeval.utils import get_or_create_event_loop
+from deepeval.metrics.indicator import metric_progress_indicator
+
+
+class AgentLoopDetectionMetric(BaseMetric):
+    """Detects infinite loops and cyclical patterns in agent execution traces.
+
+    Analyzes three sub-signals: tool call repetition, reasoning stagnation,
+    and call graph cycles. Returns a score from 0.0 (severe looping) to
+    1.0 (clean execution).
+    """
+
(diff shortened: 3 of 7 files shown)


Jeel3011:
I would like here from managers soon!!!

Jeel3011:
Hey @jeffreyip — just pushed the completed documentation page for `AgentLoopDetectionMetric` as requested.

The doc page (`docs/content/docs/(agentic)/metrics-agent-loop-detection.mdx`) now follows the full structure of the existing agentic metric docs:

- **Required Arguments** — clarifies this is a trace-only metric (no `tools_called` / `expected_tools` needed)
- **Usage** — `evals_iterator`, within components, and standalone patterns with full code examples
- **All 8 parameters** documented including `include_reason`, `async_mode`, and an explicit note that there is no `model` parameter and why
- **How Is It Calculated** — expanded with per-tier score tables (1.0 / 0.5 / 0.0) for each of the three sub-signals
- **Score Breakdown** — documents the `score_breakdown` attribute
- **Configuring Sub-signals** — shows how to toggle individual checks
- **Limitations & Design Decisions** — explains the deterministic design choice, cycle detection heuristics, and stagnation trade-offs
- **Community attribution** note at the bottom linking to AGeval and issue #2643

Happy to adjust anything — structure, wording, depth of any section. Just let me know 🙏

penguine-ip:
hey @Jeel3011 apologies for the long delay, finally had a chance to review it and it looks great! Thank you so much, merging and releasing later today :)

Issue #2497, opened 2026-02-18
Security: deepeval hijacks global OpenTelemetry TracerProvider, exfiltrating host application trace data to third-party endpoints


ajainvivek:
## Summary

`deepeval` v3.7.7 introduces **critical security and privacy risks** for any application that uses OpenTelemetry. At module import time (not at evaluation time), `deepeval/telemetry.py` silently:

1. **Hijacks the global OpenTelemetry `TracerProvider`** — causing ALL application spans (not just deepeval's) to be routed to deepeval's own New Relic account
2. **Initializes Sentry with 100% CPU profiling** (`profiles_sample_rate=1.0`, `traces_sample_rate=1.0`)
3. **Overrides the host application's `sys.excepthook`** — sending uncaught exceptions to deepeval's Sentry DSN
4. **Makes a blocking HTTP call** to `https://api.ipify.org` to collect the server's public IP
5. **Initializes PostHog analytics** that phones home with usage data

## Why this is a security risk

The most severe issue is **#1**: by calling `trace.set_tracer_provider(TracerProvider())` at import time, deepeval registers itself as the global tracer provider. Any application that subsequently calls `trace.set_tracer_provider()` gets the warning:

> "Overriding of current TracerProvider is not allowed"

This means **all application tracing data** — every span created via `trace.get_tracer()` anywhere in the host application — flows through deepeval's `BatchSpanProcessor` and is exported to:

```
Endpoint: https://otlp.nr-data.net:4317
API Key:  1711c684db8a30361a7edb0d0398772cFFFFNRAL
```

This is not limited to deepeval's own telemetry. It captures business logic spans, request traces, database query traces, and any other OpenTelemetry instrumentation in the host application. **This is data exfiltration of application telemetry to a third party without user consent or disclosure.**

## Reproduction

Any Python application using OpenTelemetry + deepeval:

```python
from opentelemetry import trace
from opentelemetry.sdk.trace import TracerProvider

# Import anything from deepeval — triggers telemetry.py at module load
from deepeval.metrics import GEval

# This now FAILS silently — deepeval already owns the global provider
my_provider = TracerProvider()
trace.set_tracer_provider(my_provider)
# WARNING: "Overriding of current TracerProvider is not allowed"

# All spans now go to deepeval's New Relic, not your own backend
tracer = trace.get_tracer(__name__)
with tracer.start_as_current_span("my-business-logic"):
    pass  # ← this span is exported to otlp.nr-data.net
```

## Affected code

[...]

PR #2658, opened 2026-05-07
test: guard test-run tracer provider scope

mindbomber:
## Summary

Adds a focused regression test for the test-run OpenTelemetry setup:

- verifies `init_global_test_run_tracer()` creates and stores DeepEval's own tracer provider/tracer
- verifies it does not call `opentelemetry.trace.set_tracer_provider()`
- verifies it does not replace the host application's current global tracer provider

## Scope

Issue #2497 reports the risk of library telemetry taking over a host application's global OpenTelemetry provider. Current `main` already appears to use a local provider path for `deepeval.dataset.test_run_tracer`; this PR pins that boundary so future changes do not regress into global provider replacement.

This is intentionally test-only and does not claim to resolve every telemetry/privacy concern listed in the issue.

## Verification

- `python -m pytest tests/test_core/test_dataset_test_run_tracer.py`
- `python -m black --check tests/test_core/test_dataset_test_run_tracer.py`
- `python -m ruff check tests/test_core/test_dataset_test_run_tracer.py`

### DIFF
--- a/tests/test_core/test_dataset_test_run_tracer.py
+++ b/tests/test_core/test_dataset_test_run_tracer.py
@@ -1,0 +1,54 @@
+from opentelemetry import trace
+
+from deepeval.dataset import test_run_tracer
+
+
+class FakeOTLPSpanExporter:
+    def __init__(self, *args, **kwargs):
+        self.args = args
+        self.kwargs = kwargs
+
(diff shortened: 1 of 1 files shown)


penguine-ip:
@mindbomber looking good, thank you!

PR #2659, opened 2026-05-07
feat: group contextual precision retrieval units

mindbomber:
## Summary

Adds an opt-in grouping hook for `ContextualPrecisionMetric` so users can score retrieval units instead of raw chunks when their RAG pipeline uses overlapping windows.

- adds `group_by: Optional[Callable[[str], Hashable]] = None`
- preserves current per-chunk scoring when `group_by` is not set
- groups verdicts in first-seen order when `group_by` is set
- treats a group as relevant if any chunk in that group is relevant

## Why

Issue #2594 describes a common RAG eval failure mode: overlapping chunks from the same answer-supporting section can be penalized as separate retrieval misses. The discussion converged on a backward-compatible `group_by` hook so users can map chunks to a domain-specific retrieval unit, such as a document section.

This keeps the default behavior unchanged while letting users express section/span-level precision when that is the real retrieval unit.

## Scope

This is an opt-in first step. It does not add a metadata schema, automatic section parser, or benchmark claim for financial-document RAG. Users provide the grouping key from their own chunk format.

## Verification

- `python -m pytest tests/test_metrics/test_contextual_precision_grouping.py`
- `python -m pytest tests/test_metrics/test_contextual_precision_metric.py tests/test_metrics/test_contextual_precision_grouping.py`
- `python -m black --check deepeval/metrics/contextual_precision/contextual_precision.py tests/test_metrics/test_contextual_precision_grouping.py`
- `python -m ruff check deepeval/metrics/contextual_precision/contextual_precision.py tests/test_metrics/test_contextual_precision_grouping.py`

### DIFF
--- a/deepeval/metrics/contextual_precision/contextual_precision.py
+++ b/deepeval/metrics/contextual_precision/contextual_precision.py
@@ -1,4 +1,4 @@
-from typing import Optional, List, Type, Union
+from typing import Callable, Hashable, Optional, List, Type, Union
 
 from deepeval.utils import (
     get_or_create_event_loop,
@@ -39,6 +39,7 @@ class ContextualPrecisionMetric(BaseMetric):
         async_mode: bool = True,
         strict_mode: bool = False,
         verbose_mode: bool = False,
+        group_by: Optional[Callable[[str], Hashable]] = None,
         evaluation_template: Type[
             ContextualPrecisionTemplate
         ] = ContextualPrecisionTemplate,
@@ -50,6 +51,7 @@ class ContextualPrecisionMetric(BaseMetric):
         self.async_mode = async_mode
         self.strict_mode = strict_mode
         self.verbose_mode = verbose_mode
+        self.group_by = group_by
         self.evaluation_template = evaluation_template
 
     def measure(
@@ -99,6 +101,9 @@ class ContextualPrecisionMetric(BaseMetric):
                         multimodal,
                     )
                 )
+                self._score_verdicts = self._get_score_verdicts(
+                    retrieval_context
+                )
                 self.score = self._calculate_score()
                 self.reason = self._generate_reason(input, multimodal)
                 self.success = self.score >= self.threshold
@@ -147,6 +152,7 @@ class ContextualPrecisionMetric(BaseMetric):
                     input, expected_output, retrieval_context, multimodal
                 )
             )
+            self._score_verdicts = self._get_score_verdicts(retrieval_context)
             self.score = self._calculate_score()
             self.reason = await self._a_generate_reason(input, multimodal)
             self.success = self.score >= self.threshold
@@ -165,7 +171,7 @@ class ContextualPrecisionMetric(BaseMetric):
 
         retrieval_contexts_verdicts = [
             {"verdict": verdict.verdict, "reason": verdict.reason}
-            for verdict in self.verdicts
+            for verdict in getattr(self, "_score_verdicts", self.verdicts)
         ]
         prompt = self.evaluation_template.generate_reason(
             input=input,
@@ -188,7 +194,7 @@ class ContextualPrecisionMetric(BaseMetric):
 
         retrieval_contexts_verdicts = [
             {"verdict": verdict.verdict, "reason": verdict.reason}
-            for verdict in self.verdicts
+            for verdict in getattr(self, "_score_verdicts", self.verdicts)
         ]
         prompt = self.evaluation_template.generate_reason(
             input=input,
@@ -255,15 +261,38 @@ class ContextualPrecisionMetric(BaseMetric):
             ],
         )
 
+    def _get_score_verdicts(
+        self, retrieval_context: List[str]
+    ) -> List[cpschema.ContextualPrecisionVerdict]:
+        if self.group_by is None:
+            return self.verdicts
+
+        grouped_verdicts: dict[
+            Hashable, cpschema.ContextualPrecisionVerdict
+        ] = {}
+        for index, verdict in enumerate(self.verdicts):
+            context = retrieval_context[index]
+            key = self.group_by(context)
+            current_verdict = grouped_verdicts.get(key)
+            if current_verdict is None:
+                grouped_verdicts[key] = verdict
+                continue
+
+            if verdict.verdict.strip().lower() == "yes":
+                grouped_verdicts[key] = verdict
+
+        return list(grouped_verdicts.values())
+
     def _calculate_score(self):
-        number_of_verdicts = len(self.verdicts)
+        score_verdicts = getattr(self, "_score_verdicts", self.verdicts)
+        number_of_verdicts = len(score_verdicts)
         if number_of_verdicts == 0:
             return 0
 
         # Convert verdicts to a binary list where 'yes' is 1 and others are 0
         node_verdicts = [
             1 if v.verdict.strip().lower() == "yes" else 0
-            for v in self.verdicts
+            for v in score_verdicts
         ]
 
         sum_weighted_precision_at_k = 0.0
--- a/tests/test_metrics/test_contextual_precision_grouping.py
+++ b/tests/test_metrics/test_contextual_precision_grouping.py
@@ -1,0 +1,64 @@
+import pytest
+
+from deepeval.metrics.contextual_precision.contextual_precision import (
+    ContextualPrecisionMetric,
+)
+from deepeval.metrics.contextual_precision.schema import (
+    ContextualPrecisionVerdict,
+)
+
+
+def make_metric(verdicts, group_by=None):
+    metric = object.__new__(ContextualPrecisionMetric)
+    metric.threshold = 0.5
+    metric.strict_mode = False
+    metric.group_by = group_by
+    metric.verdicts = [
+        ContextualPrecisionVerdict(verdict=verdict, reason="")
+        for verdict in verdicts
+    ]
+    return metric
+
+
+def test_contextual_precision_score_defaults_to_per_chunk_verdicts():
+    metric = make_metric(["yes", "no", "yes"])
+
+    assert metric._calculate_score() == pytest.approx(0.8333333333)
+
+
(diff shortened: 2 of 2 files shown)


Ruthwik-Data (commented):
Reviewing as the author of issue #2594 that motivated this fix. The `group_by` opt-in approach is the right call — backward-compatible, puts grouping logic with the user who knows their chunk format, clean implementation.

A few observations from testing section-aware chunking on financial documents (SEC 10-K filings):

**1. `any yes = relevant` aggregation works but loses the reason from the best chunk.**
In `_get_score_verdicts`, when a group has a "yes" verdict, the whole group's verdict is replaced with that specific chunk's verdict object. This means the `reason` field reflects that one chunk, not the group. For debugging — especially in financial RAG where knowing *why* a section was marked relevant matters — it would be useful to surface a consolidated reason, e.g. `"Group 'Revenue Discussion' marked relevant via chunk at index 2"`.

**2. No guard if `len(retrieval_context) != len(self.verdicts)`.**
Line 274 does `retrieval_context[index]` assuming both lists are the same length. If verdict generation skips or errors on a chunk, this throws an `IndexError`. A `zip(self.verdicts, retrieval_context)` pattern would be safer.

**3. The financial-doc case from #2594 is now unblocked.**
Validated the concept: grouping by section header eliminates the penalty for overlapping chunks referencing the same table row. This is the fix the issue needed.

Overall the implementation is clean and the test coverage looks solid. The two points above are suggestions, not blockers. Happy to see this merged.

A-Vamshi:
Hey @mindbomber, @Ruthwik-Data  Thank you for the PRs and review, really appreciate the work you guys are doing for the community. 

A few comments / doubts I had for this PR implementation and the original issue + solution proposed in the issue: #2594  that I was hoping we could discuss to help implement this better:

- In the original issue, I think a great proposal was to group the retrieved chunks themselves into a groups using a `group_by` method. What I think works great here is that, let's say the original `retrieval_context` passed to the `ContextualPrecisionMetric` had around 10 different contexts, and after using the `group_by` function, we would have around 3 contexts (the relevant ones of the same group will be merged together)

- An approach we thought of internally is to have the retrieval contexts provide a source along with the context itself, this would allow us to group the contexts from the same source together, rather than using a `group_by` which could introduce unpredictability to code as the output format of the method nor the internal working cannot be standardized, this would mean, users without prior context to the reason why `group_by` could not use it properly unless they surf through the code or find these related threads. 

- And I think the current implementation still scores everything individually and groups the verdicts later after the LLM evaluates them, this would mean more LLM calls which don't contribute to the score.

What do you guys think of pivoting to this schema-based, pre-generation grouping approach? Would this match the use case you were talking about?

Ruthwik-Data:
Hey @A-Vamshi, thanks for the thoughtful feedback and for flagging these design concerns — really appreciate the depth here.

Responding as the author of #2594: the core problem I ran into was that `ContextualPrecisionMetric` was treating overlapping chunks from the **same retrieval span** as separate, independent retrievals — so a section split into 3 overlapping chunks would get scored 3 times, and adding overlap (which *improves* answer quality) was paradoxically *lowering* my eval score. The ask was simply: let the metric understand what a "retrieval unit" actually is, not just a raw chunk.

On your specific points:

1. **Schema-based `source` grouping** — I'm fully on board with this direction. For my 10-K / 10-Q RAG pipeline, each chunk already has metadata like `doc_id` and `section_id`, so attaching a `source` field is trivial and would map cleanly to the retrieval unit I care about. Pre-grouping before LLM evaluation also means fewer calls, which is strictly better for my use case at scale.

2. **`group_by` unpredictability** — your concern is valid. A free-form lambda is powerful but adds a documentation and discoverability burden. If users can't easily understand why their score changed, it undermines trust in the metric.

3. **LLM calls per chunk** — yes, this was part of my original concern too. Grouping verdicts post-evaluation doesn't actually fix the redundant LLM calls, it just changes how results are displayed.

**My take on the pivot:** The schema-based approach (e.g., `source` field on retrieval contexts + pre-generation grouping) would fully solve my use case, as long as:
- The grouping key can be derived from chunk metadata (e.g., `section_id`, `(doc_id, section_id)`, table row key)
- It remains opt-in / backward-compatible so existing pipelines aren't broken

If there's an escape hatch for computed keys (like a `group_key` function for cases where the grouping logic is non-trivial), that would be a bonus — but the `source`-field-first approach is already enough for the financial-doc RAG use case that motivated the issue.

Happy to help validate the new design against real 10-K retrieval output if that's useful.

mindbomber:
Thanks @A-Vamshi and @Ruthwik-Data, this makes sense to me.

This PR followed the earlier `group_by` direction from #2594, but I agree with the design concern: grouping after verdict generation does not avoid the redundant LLM calls, and a free-form callable is harder to document and reason about than a schema/source-first API.

I am happy to pivot this PR toward the schema-based, pre-generation grouping approach if that is the direction you want to land. The constraints I would preserve are:

- opt-in / backward-compatible behavior for existing `retrieval_context` users
- grouping before verdict generation so the metric scores retrieval units rather than raw overlapping chunks
- singleton behavior for contexts without a grouping key, so unrelated ungrouped chunks do not collapse into one bucket
- support for metadata-derived grouping keys like `(doc_id, section_id)` / `source`, with a computed-key escape hatch only if it fits the public API cleanly

If the core team already has a preferred shape for the retrieval-context schema, I can adjust the PR around that instead of pushing the current `group_by` implementation further.

Ruthwik-Data:
Great direction on the pivot to schema-based pre-generation grouping — this is the right call both for cost efficiency and for making the API more intuitive.

A few thoughts as @mindbomber works on the next iteration:

**On the schema approach:**
Adding a `source` or `section_id` field to retrieval contexts is cleaner than a free-form callable. It aligns with how most production RAG pipelines already tag chunks with metadata (doc_id, page_num, section, etc.). This also opens the door for richer diagnostics — e.g., showing scores at the section level in the UI.

**On backward compatibility:**
Make sure the default behavior (per-chunk scoring without `group_by`) remains identical. A silent behavior change here would break a lot of existing eval suites without any warning.

**On cost reduction:**
Pre-generation grouping is the real win here. If we deduplicate before the LLM call, we could see 30-50% fewer tokens for RAG pipelines with heavy chunk overlap (common in sliding-window setups). Worth calling this out explicitly in the docs.

**Suggestion:**
Consider adding a `grouping_strategy` enum (`"none"`, `"source"`, `"custom"`) alongside the `source` field to make the API self-documenting and easier to validate.

Happy to review the updated implementation once the schema-based approach is in. This is a meaningful improvement to how deepeval handles real-world RAG eval scenarios.

Ruthwik-Data:
The pivot to schema-based pre-generation grouping makes a lot of sense — @A-Vamshi's point about deduplicating before LLM calls is the right call both for cost and API intuition.

On the `grouping_strategy` enum I suggested earlier, here's a more concrete sketch of how it could look:

```python
class GroupingStrategy(str, Enum):
    SOURCE_FIELD = "source_field"    # group by context[i].source metadata
    CALLABLE = "callable"            # group by user-supplied Callable[[str], Hashable]
    NONE = "none"                    # current behavior, per-chunk scoring
```

The benefit of an enum over a raw `Optional[Callable]` is that it makes the serialized eval config self-documenting — you can log which strategy was used alongside the score, which matters when you're comparing eval runs across different chunking configs.

For backward compat: `grouping_strategy=GroupingStrategy.NONE` is the default, so existing code needs zero changes.

One implementation note: if going schema-based, worth adding a validation step that raises early if `grouping_strategy=SOURCE_FIELD` but none of the contexts have a `source` field — otherwise silent failures are hard to debug in production eval pipelines.

Ruthwik-Data:
Reviewed from the perspective of the original issue (#2594) — the schema-based direction discussed with @A-Vamshi is the right call. A few thoughts to help finalize:

**On the pivot to schema-based grouping:**
The `group_by` callable approach was flexible but fragile — a user could silently pass a function that returns non-hashable values, or one that groups incorrectly, and the metric would produce wrong scores with no error. The `source` field approach is more predictable: if the metadata is missing, you can raise early with a clear message.

**Suggested minimal API surface (for alignment before the next push):**
```python
# LLMTestCase extended with optional source field
test_case = LLMTestCase(
    input="What are total current assets?",
    actual_output="...",
    retrieval_context=[
        RetrievalContext(text=chunk_a, source="10k/current_assets"),
        RetrievalContext(text=chunk_b, source="10k/current_assets"),  # same group
        RetrievalContext(text=chunk_c, source="10k/ppe"),
    ]
)
metric = ContextualPrecisionMetric(threshold=0.5, group_by_source=True)
```
When `group_by_source=True`, chunks with the same `source` are treated as one retrieval unit. Relevance verdict for the group = `any(verdicts_in_group)`. This avoids repeated LLM calls on overlapping content.

**On the merge conflict:**
The branch has a conflict with main — worth resolving before requesting another review cycle so CI can run cleanly.

**My test fixtures in #2692** will validate the `group_by_source` behavior once the implementation lands — they're written to be parameterizable. Happy to update them to match whatever final API shape you settle on here. Just let me know.

Ruthwik-Data:
Following up on the schema-based pivot we aligned on with @A-Vamshi.

The direction looks right to me: grouping before verdict generation means the LLM sees a deduplicated set of retrieval units, which solves both the precision inflation problem from #2594 and the redundant-call cost issue A-Vamshi flagged.

A few things that would help move this forward:

1. **Branch conflict** — the branch still shows conflicts with `main`. Rebasing is the blocker before CI can pass.
2. **Early validation** — if `grouping_strategy=SOURCE_FIELD` is set but a context entry is missing the `source` key, raising a `ValueError` immediately (rather than silently falling back) would save a lot of debugging time in production.
3. **Test coverage** — PR #2692 adds overlapping-chunk fixtures for the current `group_by` callable approach. Once you rebase and update to the schema-based API, those fixtures should port over without much change. Happy to update #2692 to match whatever API lands here.

Let me know if it's useful for me to draft the `GroupingStrategy` enum stub we discussed — I can add it as a suggestion block on the relevant file.

Issue #1420, opened 2025-03-06
Custom LLM in the docs provide no schema input for generate method, and returns string, which causes summarization.py raise AttributeError: 'str' object has no attribute 'truths'


pin-lpt-heatherwick:
`from deepeval.models.base_model import DeepEvalBaseLLM

class Mistral7B(DeepEvalBaseLLM):
    def __init__(
        self,
        model,
        tokenizer
    ):
        self.model = model
        self.tokenizer = tokenizer

    def load_model(self):
        return self.model

    def generate(self, prompt: str, schema=None) -> str:

        if not prompt or len(prompt.strip()) == 0:
            raise ValueError("Prompt cannot be empty")

        model = self.load_model()
        device = "cuda" 

        model_inputs = self.tokenizer([prompt], return_tensors="pt").to(device)
        model.to(device)

        if model_inputs.input_ids.numel() == 0:
            raise ValueError("Tokenized input is empty")

        generated_ids = model.generate(
            **model_inputs, 
            max_new_tokens=2000, 
            do_sample=True,
            pad_token_id=self.tokenizer.eos_token_id
            eos_token_id=self.tokenizer.eos_token_id
        )
        response = self.tokenizer.batch_decode(generated_ids)[0]
        print('response is...\n', response)
        
        return response

    async def a_generate(self, prompt: str, schema=None) -> str:
        return self.generate(prompt, schema)

    def get_model_name(self):
        return "Mistral 7B"`

Traceback (most recent call last):
  File "/home/user/.pyenv/versions/3.10.16/lib/python3.10/site-packages/gradio/queueing.py", line 625, in process_events
    response = await route_utils.call_process_api(
  File "/home/user/.pyenv/versions/3.10.16/lib/python3.10/site-packages/gradio/route_utils.py", line 322, in call_process_api
    output = await app.get_blocks().process_api(
  File "/home/user/.pyenv/versions/3.10.16/lib/python3.10/site-packages/gradio/blocks.py", line 2108, in process_api
    result = await self.call_function(
  File "/home/user/.pyenv/versions/3.10.16/lib/python3.10/site-packages/gradio/blocks.py", line 1655, in call_function
    prediction = await anyio.to_thread.run_sync(  # type: ignore
  File "/home/user/.pyenv/versions/3.10.16/lib/python3.10/site-packages/anyio/to_thread.py", line 56, in run_sync
    return await get_async_backend().run_sync_in_worker_thread(
[...]

PR #2677, opened 2026-05-18
docs: fix broken examples in custom LLMs guide (refs #1420)

zhang-liz:
## Summary

`docs/content/guides/guides-using-custom-llms.mdx` is the canonical guide users land on when adopting deepeval with non-OpenAI providers. Today it contains a code example that crashes for the user, plus a "rules" block that doesn't match the real `DeepEvalBaseLLM` contract.

Issue #1420 reports the exact failure mode — copying the basic Mistral 7B example from this guide and getting `'str' object has no attribute 'truths'` from `SummarizationMetric`. Maintainer @penguine-ip acknowledged the report on 2025-03-07 but the doc still ships the broken example. This PR fixes it.

## Changes (2 commits)

1. **Fix two broken Python snippets** (`5c0f8e6a9`)
   - `CustomLlama3_8B.generate()` is annotated `-> str` but returns `pipeline(prompt)` (a `list[dict]`). Now returns `pipeline(prompt)[0]["generated_text"]`.
   - The JSON-confined `CustomMistral7B.generate()` has `pipeline = pipeline(...)` — `pipeline` is never imported and Python promotes it to a local, so this line raises `UnboundLocalError` before any model call. Now imports `transformers` and calls `transformers.pipeline(...)`, matching the working `CustomLlama3_8B` variant above.

2. **Align rules with the real contract** (`e7fc09d0e`)
   - Both rules blocks ("SIX rules" and "ALWAYS") said `generate()` must take "one and only one parameter of type string." The real contract — used by every built-in provider in `deepeval/models/llms/` — is `generate(prompt: str, schema: Optional[BaseModel] = None)`. Most metrics pass `schema=...`, and the `generate_with_schema()` `TypeError` fallback at `deepeval/models/base_model.py:125` doesn't help structured-output metrics, which is why #1420 surfaces as `AttributeError`.
   - The misleading "exception to rules 4. and 5." note is replaced with an actionable caution that names the failure mode and the specific metrics it affects.
   - Adds a caution above the basic Mistral 7B example (the verbatim code from #1420) routing users to the schema-aware variant.

## What I deliberately did not change

- No code changes to `deepeval/models/base_model.py` — formalizing the signature is a contract change and warrants its own discussion.
- Did not rewrite the Azure / Vertex / Bedrock examples to add `schema`. They use LangChain wrappers whose `.invoke(...)` returns strings — they hit the same `AttributeError` failure mode on structured metrics, but the new caution block in the rules section covers the general case. Touching every example would balloon the diff; happy to do it as a follow-up if maintainers want.

## Verification

- All 19 Python code blocks in the modified file `ast.parse` cleanly (excluding one pre-existing block of templated `<placeholders>` in the Bedrock example, which is unrelated).
- Total diff: +14 / −9 lines, one file.

## Test plan

- [ ] Build the Docusaurus site locally and confirm the two cautions and the cross-section anchor links render.
- [ ] Re-test the failing case from #1420: instantiate the basic `Mistral7B` example and run it through `SummarizationMetric`. With the new caution in place, users are routed to the schema-aware path before hitting the crash.

Refs #1420

### DIFF
--- a/docs/content/guides/guides-using-custom-llms.mdx
+++ b/docs/content/guides/guides-using-custom-llms.mdx
@@ -69,7 +69,7 @@ class CustomLlama3_8B(DeepEvalBaseLLM):
             pad_token_id=self.tokenizer.eos_token_id,
         )
 
-        return pipeline(prompt)
+        return pipeline(prompt)[0]["generated_text"]
 
     async def a_generate(self, prompt: str) -> str:
         return self.generate(prompt)
@@ -548,6 +548,7 @@ Here's a full example of a JSON confined custom Mistral 7B model implemented thr
 
 ```python
 import json
+import transformers
 
 from pydantic import BaseModel
 import torch
@@ -588,7 +589,7 @@ class CustomMistral7B(DeepEvalBaseLLM):
 
     def generate(self, prompt: str, schema: BaseModel) -> BaseModel:
         model = self.load_model()
-        pipeline = pipeline(
+        pipeline = transformers.pipeline(
             "text-generation",
             model=model,
             tokenizer=self.tokenizer,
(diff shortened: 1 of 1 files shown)


penguine-ip:
@zhang-liz thank you!

PR #2694, opened 2026-05-26
Fix async turn contextual recall expected outcome

RitwijParmar:
## Summary
- Fix the async TurnContextualRecallMetric path to pass test_case.expected_outcome into _a_get_contextual_recall_scores
- Add an API-free regression test with a fake DeepEvalBaseLLM so the bug is caught without external model calls

## Why
The sync path already evaluates retrieval context against expected_outcome. The async path was passing test_case.multimodal as the expected_outcome argument, which meant the prompt could receive a boolean instead of the expected behavior being evaluated. That can make otherwise valid conversational retrieval score as no/0.

## Verification
- /tmp/deepeval-venv/bin/python -m pytest tests/test_metrics/test_turn_contextual_recall_no_api.py -q
- /tmp/deepeval-venv/bin/python -m compileall deepeval/metrics/turn_contextual_recall tests/test_metrics/test_turn_contextual_recall_no_api.py
- /tmp/deepeval-venv/bin/python -m black --check deepeval/metrics/turn_contextual_recall/turn_contextual_recall.py tests/test_metrics/test_turn_contextual_recall_no_api.py
- git diff --check

### DIFF
--- a/deepeval/metrics/turn_contextual_recall/turn_contextual_recall.py
+++ b/deepeval/metrics/turn_contextual_recall/turn_contextual_recall.py
@@ -161,7 +161,7 @@ class TurnContextualRecallMetric(BaseConversationalMetric):
             async def get_individual_scores(window):
                 scores.extend(
                     await self._a_get_contextual_recall_scores(
-                        window, test_case.multimodal, multimodal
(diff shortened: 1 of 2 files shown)


penguine-ip:
@RitwijParmar thanks!

PR #2701, opened 2026-05-28
Fix trimAndLoadJson corrupting valid JSON string values

Kymi808:
## Summary

`trimAndLoadJson` strips a comma + optional whitespace before a closing `]`/`}` to drop trailing commas from model output:

```python
re.sub(r",\s*([\]}])", r"\1", jsonStr)
```

It runs this regex unconditionally over the whole string before `json.loads`. Because the regex is not string-literal aware, it also rewrites `", ]"` / `", }"` sequences **inside string values**, silently corrupting otherwise-valid JSON:

```python
trimAndLoadJson('{"reason": "found items A, B, ] then stopped"}')
# -> {'reason': 'found items A, B] then stopped'}   # ", ]" collapsed to "]"
```

This affects two copies of the helper:
- `deepeval/metrics/utils.py` — parses LLM judge output (a `reason`/explanation field can easily contain such text).
- `deepeval/dataset/utils.py` — parses tool calls when loading datasets from CSV/JSONL.

## Fix

Parse the input directly first, and only apply the trailing-comma cleanup as a **fallback** when the direct parse fails. Valid JSON is then never altered, while malformed output with a genuine trailing comma (e.g. `{"a": [1, 2, ]}`) is still recovered.

## Test plan

- Added `tests/test_core/test_trim_and_load_json.py` (parametrized over both copies):
  - valid JSON whose string values contain `", ]"` / `", }"` round-trips unchanged — **fails on `main`**, passes here;
  - a real trailing comma is still stripped;
  - genuinely-invalid JSON still raises `ValueError`.
- `pytest tests/test_core/test_trim_and_load_json.py` → 8 passed. `black` clean.

## Note

Not a duplicate of #2410 — that PR strips `<think>` tags before parsing (a different failure mode) and only touches the metrics copy; this PR fixes the trailing-comma string corruption in both copies. The changes are complementary.

### DIFF
--- a/deepeval/dataset/utils.py
+++ b/deepeval/dataset/utils.py
@@ -109,11 +109,17 @@ def convert_convo_goldens_to_convo_test_cases(
 
 
 def trimAndLoadJson(input_string: str) -> Any:
+    stripped = input_string.strip()
     try:
-        cleaned_string = re.sub(r",\s*([\]}])", r"\1", input_string.strip())
-        return json.loads(cleaned_string)
-    except json.JSONDecodeError as e:
-        raise ValueError(f"Invalid JSON: {input_string}. Error: {str(e)}")
+        return json.loads(stripped)
+    except json.JSONDecodeError:
+        # Strip a trailing comma before a closing ] or } and retry, but only
+        # after a direct parse fails, so valid JSON string values containing
+        # ", ]" or ", }" are never corrupted.
+        try:
+            return json.loads(re.sub(r",\s*([\]}])", r"\1", stripped))
+        except json.JSONDecodeError as e:
+            raise ValueError(f"Invalid JSON: {input_string}. Error: {str(e)}")
     except Exception as e:
         raise Exception(f"An unexpected error occurred: {str(e)}")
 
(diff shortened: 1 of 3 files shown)


Kymi808:
Heads up for reviewers — all three failing checks here are pre-existing on `main`, not from this PR:

- **`lint (ubuntu-latest)`** — `psf/black@stable` (currently black 26.x) flags `58 files would be reformatted` on `origin/main` itself, in files like `deepeval/cli/inspect.py`, `deepeval/inspect/widgets/*`, and various `deepeval/metrics/*/template.py`. None are touched by this PR. The three files I changed pass `black --check` cleanly under the same black 26.x.
- **`test` (×2)** — both fail at startup with `ValueError: No Confident API key found` and `DeepEvalError: OpenAI API key is empty`, i.e. missing CI secrets. Per `CONTRIBUTING.md` ("no need to worry about failing test cases in GitHub Actions, as these are mostly for internal use and will only pass if triggered by a user with the correct permissions within Confident AI"), this is the expected pattern for external PRs.

Happy to file the repo-wide black-26 reformat as a separate cleanup PR if that would help, but didn't want to bundle 58 unrelated reformats into a one-line bug fix.

penguine-ip:
@Kymi808 this makes sense, thank you!

PR #2708, opened 2026-05-29
fix: pass expected_outcome in TurnContextualRecallMetric async path

### LINKED ISSUE #2679: TurnContextualRecallMetric (async): expected_outcome dropped, test_case.multimodal passed as assistant output → always scores 0
**Describe the bug**

`TurnContextualRecallMetric` always scores **0** when run in async mode (the default). The judge LLM receives the string `"False"` as the assistant output instead of the actual `expected_outcome` text, so every verdict is `no` and every window scores 0.

**Root cause**

In `a_measure`, the inner coroutine mistakenly passes `test_case.multimodal` (a `bool`) where `expected_outcome` should go, and drops `expected_outcome` entirely:

```python
# turn_contextual_recall.py  (async path, a_measure)
async def get_individual_scores(window):
    scores.extend(
        await self._a_get_contextual_recall_scores(
            window, test_case.multimodal, multimodal   # ← BUG
        )
    )
```

`_a_get_contextual_recall_scores` signature is `(turns_window, expected_outcome, multimodal)`, so `test_case.multimodal` (= `False`) lands in `expected_outcome`. The template then renders:

```
Assistant Output:
False
```

for every window. The sync path has it right:

```python
# sync path — correct
self._get_contextual_recall_scores(
    window, test_case.expected_outcome, multimodal   # ← correct
)
```

**Fix**

```python
# a_measure, inside get_individual_scores
await self._a_get_contextual_recall_scores(
    window, test_case.expected_outcome, multimodal   # pass expected_outcome
)
```

**To Reproduce**

```python
from deepeval import evaluate
from deepeval.metrics import TurnContextualRecallMetric
from deepeval.test_case import ConversationalTestCase, Turn

test_case = ConversationalTestCase(
    chatbot_role="Technical support chatbot",
    expected_outcome="User recovers access by re-authenticating the sync client.",
    turns=[
        Turn(role="user", content="My folder stopped syncing after I changed my password."),
        Turn(
            role="assistant",
            content="Sign out of the sync client and sign back in with your new password.",
            retrieval_context=[
                "After a password change the stored OAuth token is invalidated. "
                "Users must sign out and sign back in to generate a fresh token."
            ],
        ),
    ],
)

metric = TurnContextualRecallMetric(threshold=0.5, verbose_mode=True)
evaluate([test_case], [metric])
# Score: 0.0 — expected: ~1.0
```

**Expected behavior**

The `expected_outcome` string is sent to the judge as the assistant output. Score should reflect whether the retrieval context supports that outcome (≥ threshold for a well-grounded context).

**Observed behavior**

The judge receives `"False"` as the assistant output. All verdicts are `no`. Final score is `0.0` regardless of how complete the retrieval context is.

**LLM traffic evidence (from llm-traffic-inspector)**

Every per-window judge call contains:
```
Assistant Output:
False
```

**Environment**

- deepeval version: 4.0.2
- Python: 3.13
- async_mode: `True` (default)

koriyoshi2041:
In async mode (the default), `TurnContextualRecallMetric` scores 0 for every conversation. `a_measure` passes `test_case.multimodal` (a `bool`) into the `expected_outcome` argument of `_a_get_contextual_recall_scores`, so the judge receives `"False"` instead of the expected outcome text and returns `no` for every verdict.

The sync `measure()` path already passes `expected_outcome` correctly — this brings the async path in line with it.

Also adds a regression test under `tests/test_metrics/`. It stubs the scoring helper and reason generation, so it needs no `OPENAI_API_KEY`, and asserts `a_measure` forwards `expected_outcome` (a `str`) rather than the multimodal flag. Verified it fails on `main` and passes with the fix.

Closes #2679

### DIFF
--- a/deepeval/metrics/turn_contextual_recall/turn_contextual_recall.py
+++ b/deepeval/metrics/turn_contextual_recall/turn_contextual_recall.py
@@ -161,7 +161,7 @@ class TurnContextualRecallMetric(BaseConversationalMetric):
             async def get_individual_scores(window):
                 scores.extend(
                     await self._a_get_contextual_recall_scores(
-                        window, test_case.multimodal, multimodal
+                        window, test_case.expected_outcome, multimodal
                     )
                 )
 
--- a/tests/test_metrics/test_turn_contextual_recall_async_expected_outcome.py
+++ b/tests/test_metrics/test_turn_contextual_recall_async_expected_outcome.py
@@ -1,0 +1,74 @@
+"""Regression test for TurnContextualRecallMetric async scoring.
+
+In async mode (the default), ``a_measure`` used to forward
+``test_case.multimodal`` (a bool) into the ``expected_outcome`` slot of the
+scoring helper, dropping the real ``expected_outcome`` text. Every verdict then
+scored 0. The sync ``measure`` path was unaffected.
+
+These tests use DummyModel and do not require OPENAI_API_KEY: the scoring helper
+and reason generation are stubbed so no LLM call is made, and we assert directly
+on the argument that ``a_measure`` forwards.
+"""
+
+import pytest
+from unittest.mock import patch, AsyncMock
+
+from deepeval.metrics import TurnContextualRecallMetric
+from deepeval.test_case import ConversationalTestCase, Turn
+from tests.test_core.stubs import DummyModel
+
+
+def make_metric() -> TurnContextualRecallMetric:
+    """Create the metric with DummyModel so construction makes no LLM calls."""
+    with patch(
+        "deepeval.metrics.turn_contextual_recall.turn_contextual_recall.initialize_model"
+    ) as mock_init:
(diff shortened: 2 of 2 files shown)


koriyoshi2041:
CI triage note:

- The targeted Metrics Tests job passed on this PR, including `tests/test_metrics/test_turn_contextual_recall_async_expected_outcome.py`.
- The new regression test is also reported as black-formatted in the lint log (`already well formatted`).
- The Vercel failure appears to be the standard fork authorization gate.
- The failing Core/Confident test jobs are hitting missing secret/API-key setup (`OPENAI_API_KEY` / `CONFIDENT_API_KEY`) during unrelated test collection or confident upload tests, not the changed `TurnContextualRecallMetric` path.

So I think the patch-specific evidence is green; the remaining failures look like fork CI environment issues.

penguine-ip:
@koriyoshi2041 thanks!

PR #2712, opened 2026-05-29
fix: loosen opentelemetry version constraint to allow 1.x and 2.x

### LINKED ISSUE #1968: Deepeval is pinned to Otel version, causing issues when importing
Deepeval is pinning Otel dependencies that in turn bound profobuf to < 6.0:
> opentelemetry-exporter-otlp-proto-grpc = "^1.24.0"

This is causing issues on my project that requires higher version of protobuf (version >= 6.0.0). I would recommend to loosen the dependency requirements so that it can be adopted easily.

You can test this by adding protobuf dependency = "^6.0.0"

nuthalapativarun:
`opentelemetry-api = "^1.24.0"` and `opentelemetry-sdk = "^1.24.0"` use Poetry's caret operator, which resolves to `>=1.24.0, <2.0.0`. This blocks users whose projects need opentelemetry 2.x or transitively require a version of other packages that depends on it.

**Change:**
- `pyproject.toml`: Changed both constraints from `^1.24.0` to `>=1.24.0,<3.0.0`, allowing both the current 1.x series and future 2.x releases. The lower bound is preserved so all existing functionality continues to work.

Fixes #1968

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -11242,4 +11242,4 @@ inspect = ["pyperclip", "textual"]
 [metadata]
 lock-version = "2.1"
 python-versions = ">=3.9, <4.0"
-content-hash = "1c8a2d716d795a792069526c406df879b4f1f783ab4833e09093139e439ba07d"
+content-hash = "6d371376338552bbdf39d0292c1dc72e17fc347a23dceb4ecbf48ed22d1f3d6f"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,8 +36,8 @@ setuptools = "*"
 wheel = "*"
 nest_asyncio = "*"
 tenacity = ">=8.0.0,<=10.0.0"
-opentelemetry-api = "^1.24.0"
-opentelemetry-sdk = "^1.24.0"
+opentelemetry-api = ">=1.24.0,<3.0.0"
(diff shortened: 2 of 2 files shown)


nuthalapativarun:
## Root cause and fix

All integration test CI jobs were failing with this error during `poetry install`:

```
pyproject.toml changed significantly since poetry.lock was last generated. Run `poetry lock` to fix the lock file.
```

**Root cause:** The previous commit changed the `opentelemetry-api` and `opentelemetry-sdk` version constraints in `pyproject.toml` (from `^1.24.0` to `>=1.24.0,<3.0.0`), but did not regenerate `poetry.lock`. Poetry 2.x stores a `content-hash` of `pyproject.toml` in the lock file. When the constraint strings change, the hash changes, and Poetry refuses to install dependencies — it exits with code 1 before any test even runs.

This is why *all* integration tests failed, and why they failed very quickly (~20–25 seconds into the job): the failure happened at the dependency installation step, not during the actual test execution.

**Fix:** Ran `poetry lock` to regenerate the lock file. Only the `content-hash` line in `poetry.lock` changed — no dependency versions changed, because the existing locked opentelemetry versions (1.x) already satisfy the loosened `>=1.24.0,<3.0.0` constraint.

py-ai-dev (approved):

PR #2713, opened 2026-05-29
docs: document how to enable GEval weighted summation for custom models

### LINKED ISSUE #1831: Adding weighted summation for GEval with Custom Models
## Context
On the [docs for GEval](https://deepeval.com/docs/metrics-llm-evals) it's mentioned that 
```
This step was introduced in the paper because it minimizes bias in LLM scoring. 
This normalization step is automatically handled by deepeval by default (unless you're using a custom model).
```
## The Problem
I'm using a custom model, and my full pipeline works using GEval. I'm noticing that the token probs are not being considered. It's unclear how to add it from reading the code and the documentation. In the GEval paper, the addition of this step leads to better results, so it's crucial to support this.

## Potential Solution / Alternatives 
So, an alternative (before even mentioning a solution) is that there's a way to do this that I am unaware of, if so, then it's a documentation issue. 

Otherwise, I think there are two prongs here 

1) an optionally filled logprobs function in DeepEvalBaseLLM

Based on the code, it looks like it might be able to bypass and implement `a_generate_raw_response` in the child class of DeepEvalBaseLLM, but this is not documented. I can investigate further once I find out there's no established way to do this.

2) Support for an oversampling parameter in GEval objects to get an MLE estimate of log probs (less full baked)

This would look like instantiating GEval:
```
correctness_metric = GEval(
    name="Blah",
    criteria="footer",
    evaluation_params=[LLMTestCaseParams.INPUT, LLMTestCaseParams.ACTUAL_OUTPUT, LLMTestCaseParams.EXPECTED_OUTPUT],
    
)
```
and in `deepeval/deepeval/metrics/g_eval/_evaluate` when the function calls  `self.model.generate_raw_response()`, maybe it could be called multiple times in parallel to get token probability estimates, the way it is done in the paper with GPT-4.

nuthalapativarun:
The GEval docs note that the weighted summation step from the G-Eval paper is "automatically handled by deepeval by default (unless you're using a custom model)" but give no guidance on how custom-model users can opt in.

Custom `DeepEvalBaseLLM` subclasses can enable weighted summation by implementing `a_generate_raw_response` to return a `(ChatCompletion, cost)` tuple with `logprobs` populated. `deepeval` already handles the rest internally via `calculate_weighted_summed_score`, but this path was completely undocumented.

**Changes:**
- `docs/content/docs/(custom)/metrics-llm-evals.mdx`: Expanded the existing tip block to include a short explanation and a minimal code example showing how to implement `a_generate_raw_response` on a custom model class.

Fixes #1831

### DIFF
--- a/docs/content/docs/(custom)/metrics-llm-evals.mdx
+++ b/docs/content/docs/(custom)/metrics-llm-evals.mdx
@@ -233,6 +233,26 @@ When you provide `evaluation_steps`, the `GEval` metric skips the first step and
 In the original G-Eval paper, the authors used the probabilities of the LLM output tokens to normalize the score by calculating a weighted summation.
 
 This step was introduced in the paper because it minimizes bias in LLM scoring. **This normalization step is automatically handled by `deepeval` by default** (unless you're using a custom model).
+
+To enable weighted summation for a custom model, implement `a_generate_raw_response` (and optionally `generate_raw_response`) on your `DeepEvalBaseLLM` subclass. The method must return a tuple of `(ChatCompletion, cost)` where the `ChatCompletion` object contains `logprobs` in its response:
+
+```python
+from openai.types.chat import ChatCompletion
+
+class MyCustomModel(DeepEvalBaseLLM):
+    async def a_generate_raw_response(self, prompt: str, **kwargs) -> tuple:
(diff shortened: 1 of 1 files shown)


penguine-ip:
@nuthalapativarun thanks!

PR #2717, opened 2026-06-01
feat(metric data): add tokens usage

tanayvaswani:


deepeval/evaluate/utils.py
```diff
@@ -64,6 +64,8 @@ def create_metric_data(metric: BaseMetric) -> MetricData:
             evaluationModel=metric.evaluation_model,
             error=metric.error,
             evaluationCost=metric.evaluation_cost,
+            inputTokenCount=getattr(metric, "input_tokens", None),
```
kritinv: can we use metric.input_tokens

deepeval/metrics/utils.py
```diff
@@ -453,6 +453,22 @@ def trimAndLoadJson(
 ReturnType = TypeVar("ReturnType")
 
 
+def accrue_token_usage(metric: Any, cost: Any) -> None:
```
kritinv: type metric as Union of metric types (not Any) so no need for  `if hasattr(metric, "_accrue_tokens")`

deepeval/metrics/answer_relevancy/answer_relevancy.py
```diff
@@ -70,6 +70,8 @@ def measure(
         )
 
         self.evaluation_cost = 0 if self.using_native_model else None
+        self.input_tokens = 0 if self.using_native_model else None
```
kritinv, line 73: btw we should either stick with input_tokens or input_token_count, and not switch around the naming convention too much. Would use whatever openAI uses

deepeval/metrics/base_metric.py
```diff
@@ -117,6 +131,16 @@ def _accrue_cost(self, cost: float) -> None:
         else:
```
kritinv, line 131: accrue cost type is wrong with new type?

deepeval/models/utils.py
```diff
@@ -7,6 +7,38 @@
...
+    e.g. for metric-run token visibility, without changing any return
+    signatures. Token data rides along the returned value, so it stays correct
+    under concurrent async fan-out (no shared model-instance state).
+
+    Providers that have not yet been updated simply return a plain ``float``;
+    readers fall back to ``None`` tokens via ``getattr(cost, "input_tokens",
+    None)``, so adoption can be incremental.
+    """
+
+    input_tokens: Optional[int]
```
kritinv, line 27: same here with naming

kritinv: I think it's in a few othe rplaces I didn't catch

deepeval/metrics/utils.py
```diff
@@ -453,6 +453,24 @@ def trimAndLoadJson(
...
+    """Accrue the input/output token counts that produced ``cost`` onto the
+    metric, when available.
+
+    Native models return their cost as a ``TokenCost`` (a ``float`` subclass
+    carrying ``input_tokens``/``output_tokens``), plain floats from
+    not-yet-updated providers degrade gracefully to ``None``. Call this right
+    after ``metric._accrue_cost(cost)`` so token usage tracks cost exactly.
+    """
+    metric._accrue_tokens(
+        getattr(cost, "input_tokens", None),
```
kritinv: can we directly reference cost here as welll

PR #2720, opened 2026-06-01
Fix TurnFaithfulnessMetric crash on assistant turns with no claims

Aarkin7:
While poking at TurnFaithfulnessMetric I noticed it crashes on a really common code path: any conversation containing an assistant turn with no factual claims to extract (things like "Sure!", "Happy to help!", "Let me think about that…").

### What's wrong

In deepeval/metrics/turn_faithfulness/turn_faithfulness.py, the helper is declared as:

def _get_interaction_score_and_reason(self, verdicts, multimodal) -> Tuple[float, str]:

…and the caller relies on that shape:

score, reason = self._get_interaction_score_and_reason(...)

But on the "no verdicts to evaluate" branch it returns a bare `1` instead of a tuple, so Python blows up unpacking it:
TypeError: cannot unpack non-iterable int object

Both the sync and async variants have the same bug (lines 382 and 408 before the patch), so toggling `async_mode` doesn't help. And the path is hit very often, because `_generate_verdicts` short-circuits to `[]` whenever the claim extractor returns no claims, which is the normal outcome for chit-chat assistant turns.

### The fix

Return a proper tuple on that branch:

 return 1.0, "<no claims to verify>"

Score of 1.0 because zero claims means nothing was unfaithful (vacuously perfect, same convention the single-turn FaithfulnessMetric uses). The reason string honors `include_reason` the same way the surrounding `_get_interaction_reason` does — it's `None` when `include_reason=False`.

### Tests

Added tests/test_metrics/test_turn_faithfulness_metric_empty_verdicts.py with four small unit tests for the empty-verdicts branch: sync, async, and both with `include_reason=False`. They use a stub `DeepEvalBaseLLM` so they run without `OPENAI_API_KEY`. All four fail against `main` with the TypeError and pass with this change. The existing test_turn_faithfulness_metric.py still collects/skips cleanly.

### DIFF
--- a/deepeval/metrics/turn_faithfulness/turn_faithfulness.py
+++ b/deepeval/metrics/turn_faithfulness/turn_faithfulness.py
@@ -379,7 +379,8 @@ class TurnFaithfulnessMetric(BaseConversationalMetric):
     ) -> Tuple[float, str]:
         number_of_verdicts = len(verdicts)
         if number_of_verdicts == 0:
-            return 1
+            reason = "<no claims to verify>" if self.include_reason else None
+            return 1.0, reason
 
         faithfulness_count = 0
         for verdict in verdicts:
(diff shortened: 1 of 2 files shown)


Aarkin7:
Heads up: three failing checks (Core Tests, Lint Lint Lint, Confident tests) are pre-existing on main and unrelated to this PR. Happy to rebase once they're sorted.

A-Vamshi (approved):
Great work! Thank you for the PR! We will merge this soon :)

PR #2721, opened 2026-06-02
fix(dataset): serialize RetrievedContextData in save_as instead of crashing

StressTestor:
## What

`EvaluationDataset.save_as()` crashes with `TypeError` in **all three** output formats when a `Golden`'s `retrieval_context` contains a `RetrievedContextData` object, which is a *type-allowed* member of the field:

```python
retrieval_context: Optional[List[Union[str, RetrievedContextData]]]
```

```python
from deepeval.dataset import Golden, EvaluationDataset
from deepeval.test_case import RetrievedContextData

g = Golden(input="q", retrieval_context=[RetrievedContextData(context="c", source="s")])
ds = EvaluationDataset(goldens=[g])
ds.save_as("json", ".")   # TypeError: Object of type RetrievedContextData is not JSON serializable
ds.save_as("csv", ".")    # TypeError: sequence item 0: expected str instance, RetrievedContextData found
ds.save_as("jsonl", ".")  # same as csv
```

## Why

`save_as` builds its output by hand and bypasses pydantic for `retrieval_context`: the json branch passes the raw list straight to `json.dump`, and the csv/jsonl branches do `"|".join(...)` over it. Neither tolerates a `RetrievedContextData`.

The key point: `RetrievedContextData` declares its own `@model_serializer` that returns `"source: context"`, and the model's canonical dump already uses it.

```python
golden.model_dump()["retrieval_context"]   # -> ['s: c']
```

So `save_as` was the only serialization path that did **not** go through the model's own serializer. Every other dump already emits that flat string; `save_as` uniquely bypassed it and crashed instead.

## Fix

Route `retrieval_context` items through `model_dump()` (the model's declared serialization) before `json.dump` / the join. The csv/jsonl join additionally coerces each item to `str`, so it cannot regress if that serializer ever changes shape.

A saved-then-reloaded golden therefore carries `retrieval_context` as strings. Preserving the structured `{context, source}` form is intentionally out of scope: that would make `save_as` deviate from the model's own `model_dump()`. (`tools_called` / `expected_tools` already coerce via `model_dump`, and the multi-turn path serializes turns through pydantic, so neither is affected.)

## Tests

Adds a regression test covering all three formats, a mixed `str` / `RetrievedContextData` list, an all-`RetrievedContextData` list, a `None` retrieval_context, and a JSON reload round-trip. The full `tests/test_core/test_datasets/` suite passes.

### DIFF
--- a/deepeval/dataset/dataset.py
+++ b/deepeval/dataset/dataset.py
@@ -24,6 +24,9 @@ from deepeval.dataset.utils import (
     format_turns,
     check_tracer,
     parse_turns,
+    serialize_retrieval_context,
+    join_retrieval_context,
+    reconstruct_retrieval_context,
     trimAndLoadJson,
 )
 from deepeval.dataset.api import (
@@ -42,7 +45,6 @@ from deepeval.test_case import (
     LLMTestCase,
     ConversationalTestCase,
     ToolCall,
-    RetrievedContextData,
 )
 from deepeval.test_run.hyperparameters import process_hyperparameters
 from deepeval.test_run.test_run import TEMP_FILE_PATH
@@ -318,7 +320,9 @@ class EvaluationDataset:
         ]
         retrieval_contexts = [
             (
-                retrieval_context.split(retrieval_context_col_delimiter)
+                reconstruct_retrieval_context(
+                    retrieval_context.split(retrieval_context_col_delimiter)
+                )
                 if retrieval_context
                 else []
             )
@@ -454,7 +458,9 @@ class EvaluationDataset:
             actual_output = json_obj[actual_output_key_name]
             expected_output = json_obj.get(expected_output_key_name)
             context = json_obj.get(context_key_name)
-            retrieval_context = json_obj.get(retrieval_context_key_name)
+            retrieval_context = reconstruct_retrieval_context(
+                json_obj.get(retrieval_context_key_name)
+            )
             tools_called_data = json_obj.get(tools_called_key_name, [])
             tools_called = [ToolCall(**tool) for tool in tools_called_data]
             expected_tools_data = json_obj.get(expected_tools_key_name, [])
@@ -530,7 +536,9 @@ class EvaluationDataset:
         ]
         retrieval_contexts = [
             (
-                retrieval_context.split(retrieval_context_col_delimiter)
+                reconstruct_retrieval_context(
+                    retrieval_context.split(retrieval_context_col_delimiter)
+                )
                 if retrieval_context
                 else []
             )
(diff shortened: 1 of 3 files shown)


StressTestor:
Note on CI: the `lint (ubuntu-latest)` check is red here, but it is a pre-existing repo-wide issue rather than this PR. `psf/black@stable` currently wants to reformat ~60 files that this PR does not touch (the same lint failure shows on the other open PRs). The two files changed here (`deepeval/dataset/dataset.py`, `tests/test_core/test_datasets/test_dataset.py`) are black-clean and are not in that set. Happy to rebase if you bump the formatting baseline.

A-Vamshi:
Hey @StressTestor, appreciate the PR, could you help with a few more things in this PR?

1. When reconstructing the Retrieval Contexts from JSON or CSV or JSONL, can we reconstruct them back to `RetrievedContextData` -> You can do this in this method [here](https://github.com/confident-ai/deepeval/pull/2721/changes#diff-3e7d9a2d9f5dbcb9d7043b6e49ff5ad77eb7054c2e155d4367f9263f24f6abc2R1220), instead of serializing it, please format them in the following way:
```python
return [
    (
        f"deepeval_source={item.source},deepeval_context={item.context}"
        if isinstance(item, RetrievedContextData)
        else item
    )
    for item in retrieval_context
]
```
And when converting files back to goldens, use `r"^deepeval_source=(?P<source>.*?),deepeval_context=(?P<context>.*)$"` regex pattern to reconstruct the `RetrievalContextData`.

2. Can you please check if this issue persists for multi-turn goldens as well? The `Turn` object's `retrieval_context` might face the same issue, if so, can you please help with that one as well?

Once again, thank you for the PR :)

StressTestor:
Both addressed in the latest commit (`1c8c896`).

**1. Lossless round-trip.** `retrieval_context` items are now serialized to the `deepeval_source=...,deepeval_context=...` marker you specified and reconstructed back into `RetrievedContextData` on load, via shared helpers in `dataset/utils.py` (`serialize_retrieval_context` / `reconstruct_retrieval_context`). Reconstruction is wired into every loader: `add_goldens_from_json_file`, `add_goldens_from_csv_file`, `add_goldens_from_jsonl_file`, and `add_test_cases_from_json_file`. Verified round-tripping in all three formats.

**2. Multi-turn.** Confirmed it does not crash (turns go through `format_turns`, a separate path), but it was lossy: `format_turns` kept only `rc.context` and dropped the source. It now uses the same marker, and `parse_turns` reconstructs `Turn.retrieval_context` on load, so multi-turn round-trips losslessly too.

I also made the `include_test_cases=True` path lossless (`convert_test_cases_to_goldens` was flattening to `.context` as well), so all the local file save/load paths are consistent.

Two scope notes:
- The platform-upload serialization in `dataset/api.py` has the same flatten-to-`.context` pattern, but I left it untouched since it's a separate (API payload) contract rather than the local file save/load this PR covers. Can take it as a follow-up if you want it changed.
- The marker assumes a `source`/`context` won't contain the literal `,deepeval_context=` delimiter substring (it isn't escaped). The `deepeval_` namespacing makes a real collision unlikely, but if you'd prefer an escaped or JSON-encoded form, I'll switch it.

Tests cover lossless round-trip for single-turn (all three formats), multi-turn, and `include_test_cases`.

PR #2722, opened 2026-06-02
Fix GeminiModel always reporting $0 evaluation cost

Aarkin7:
## What

`GeminiModel` was silently reporting `evaluation_cost = $0.00` for every run. The culprit was `_token_cost` being a `@staticmethod`, it couldn't reach `self.model_data.input_price` / `output_price`, so it just hard-coded `0` and returned.
Token counts were extracted correctly, but never multiplied by anything.

## Why it matters

Every Gemini evaluation reported `$0` to the console, to `EvaluationCost`, and to Confident AI dashboards. Cost comparisons across providers were quietly wrong, Gemini always looked free.

## The fix

`deepeval/models/llms/gemini_model.py`
- Added `calculate_cost(input_tokens, output_tokens)` matching the contract every other provider already uses (OpenAI, Anthropic, Grok, Kimi, Azure, Bedrock).
- Made `_token_cost` an instance method that delegates to it, falling back to a zero-cost `EvaluationCost` when token counts or prices are missing, token attrs still ride along so callers can tell "unknown" from "zero".

## Tests

`tests/test_core/test_models/test_gemini_model.py`
- Replaced the stale `test_gemini_generate_returns_zero_cost` test (which was locking in the bug as a contract).
- Added four scenarios: registry hit with real cost math, missing `usage_metadata`, unregistered model with no pricing, and a direct unit test on `calculate_cost`.
- Includes an `assert cost > 0` guard so this can't silently regress back to literal zero.

### DIFF
--- a/deepeval/models/llms/gemini_model.py
+++ b/deepeval/models/llms/gemini_model.py
@@ -314,12 +314,27 @@ class GeminiModel(DeepEvalBaseLLM):
     # Utilities
     ###############################################
 
-    @staticmethod
-    def _token_cost(response) -> EvaluationCost:
+    def calculate_cost(
+        self, input_tokens: int, output_tokens: int
+    ) -> Optional[EvaluationCost]:
+        if self.model_data.input_price and self.model_data.output_price:
+            input_cost = input_tokens * self.model_data.input_price
+            output_cost = output_tokens * self.model_data.output_price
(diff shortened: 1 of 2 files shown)


A-Vamshi:
Hey @Aarkin7, thanks for the PR!

PR #2728, opened 2026-06-08
fix: dedupe flaky retry test runs

### LINKED ISSUE #1992: Tests reported twice with `pytest.mark.flaky`

cat0825:
## Summary

- Replace retried test cases inside `TestRun.add_test_case` by logical test-case identity instead of appending duplicate dashboard rows.
- Preserve accumulation for same-name test cases that represent different logical cases, such as `_skip_reset=True` calls with different inputs.
- Apply the same behavior to LLM and conversational test cases.
- Back out the replaced attempt evaluation cost so flaky pytest retries do not double-count dashboard cost.
- Add deterministic regression coverage for retried LLM/conversational cases and same-name-but-distinct cases.

Closes #1992

## Context

PR #2619 is still open with an `awaiting code fix` label and failing checks. This PR applies the same issue fix on current `main` with a focused test suite.

## Validation

- `poetry run pytest tests/test_core/test_run/test_run_retry_dedupe.py tests/test_core/test_evaluation/test_end_to_end/test_skip_reset.py -q`
- `poetry run pytest tests/test_core/test_run -q`
- `poetry run black --check deepeval/test_run/test_run.py tests/test_core/test_run/test_run_retry_dedupe.py`
- `poetry run ruff check deepeval/test_run/test_run.py tests/test_core/test_run/test_run_retry_dedupe.py`

Note: GitHub Actions `lint (ubuntu-latest)` currently fails on repository-wide Black baseline formatting in unrelated files; the changed files in this PR are already Black-formatted locally and are reported as well formatted in the CI log.


cat0825:
Updated onto current `main` (branch was 116 commits behind; now `ahead 3, behind 0`, `MERGEABLE`). Retitled to `fix: dedupe flaky retry test runs` so `pr-title-check` passes — it was failing only on the missing conventional-commit prefix.

CI on the new head: `pr-title-check` pass, all `test` jobs pass, all integration jobs pass.

The two remaining red checks are pre-existing and not from this PR:

- `lint (ubuntu-latest)` — Black wants `.scripts/release.py` reformatted, a file untouched here; the same job is red on `main` (run [33154525461](https://github.com/confident-ai/deepeval/actions/runs/33154525461), same single file).
- `Vercel` — external-contributor deploy authorization, needs a team member.

Scope: `deepeval/test_run/test_run.py` dedupes retried runs so a flaky retry no longer double-counts, with `tests/test_core/test_run/test_run_retry_dedupe.py` pinning it.

This PR is still marked draft. @penguine-ip if the change looks reasonable I'll mark it ready — or feel free to flip it yourself.

PR #2736, opened 2026-06-09
fix: add deepseek-v4 models, fix calculate_cost, and improve error ha…

### LINKED ISSUE #2884: Synthesizer silently produces empty goldens with unregistered models (e.g. deepseek-v4-flash/v4-pro)
## Describe the bug

When using `Synthesizer.generate_goldens_from_docs()` with a model whose pricing is unknown (e.g., a new model not yet in the registry like `deepseek-v4-flash` or `deepseek-v4-pro`), no goldens are generated and no error is surfaced to the user.

GEval and other metrics work fine — only `Synthesizer` is affected.

## To Reproduce

```python
from deepeval.models import DeepSeekModel

model = DeepSeekModel(api_key="...", model="deepseek-v4-flash")
synthesizer = Synthesizer(async_mode=False, model=model)
goldens = synthesizer.generate_goldens_from_docs(
    document_paths=["some_doc.txt"],
    context_construction_config=ContextConstructionConfig(
        embedder=my_embedder,
        critic_model=model,
    ),
    max_goldens_per_context=3,
)
print(goldens)  # [] — empty, no error raised
```

Log output shows: `Utilizing 0 out of N chunks` (even when document chunking succeeds).

## Root Cause Chain

1. Model not in `DEEPSEEK_MODELS_DATA` → `input_price`/`output_price` default to `None`
2. `calculate_cost()` returns `None` when pricing is unknown (by design)
3. `ContextGenerator.evaluate_chunk()` executes `self.total_cost += None` → `TypeError`
4. `generate_contexts()` catches `except Exception`, logs it, and returns empty `contexts`
5. Empty contexts → empty goldens → user sees no output and no error

## Why this is urgent

Per [DeepSeek API docs](https://api-docs.deepseek.com/), `deepseek-chat` and `deepseek-reasoner` will be **deprecated on 2026-07-24** in favor of `deepseek-v4-flash` and `deepseek-v4-pro`. After that date, ALL DeepSeek users will hit this bug unless the fix is merged.

## Related

- PR #2736 contains a fix (guards against `None` cost at call site boundaries, registers v4 models)

Comui520:
## Summary

- Register `deepseek-v4-flash` and `deepseek-v4-pro` in `DEEPSEEK_MODELS_DATA`
- Guard against `None` cost at call sites to prevent silent `TypeError` crashes
- Fix `ContextGenerator` and `Synthesizer` to surface errors instead of silently returning empty results

### Problem

When a model's pricing is unknown (e.g., a new model not yet in the registry like `deepseek-v4-flash`), `calculate_cost()` returns `None`. Call sites in `ContextGenerator.evaluate_chunk()` and `Synthesizer._generate_schema()` attempted `total += None`, causing a `TypeError`. This error was caught by a broad `except Exception` in `generate_contexts()` and only logged — never surfaced — resulting in empty goldens with no visible error to the user.

### Changes (3 files)

| File | Change |
|---|---|
| `deepeval/models/llms/constants.py` (+16) | Register `deepseek-v4-flash` and `deepseek-v4-pro` in `DEEPSEEK_MODELS_DATA` |
| `deepeval/synthesizer/chunking/context_generator.py` (+23/-2) | Guard `evaluate_chunk` and `a_evaluate_chunk` against `None` cost (`if cost is not None: self.total_cost += cost`); raise `DeepEvalError` when all document pipelines fail instead of silently returning empty contexts |
| `deepeval/synthesizer/synthesizer.py` (+29/-3) | Guard `_generate_schema`, `_a_generate_schema`, `_generate`, `_a_generate` against `None` cost; raise `DeepEvalError` when generated contexts list is empty |

### Design Decision

`calculate_cost()` returning `None` is left untouched — "unknown price" ≠ "zero price". The fix is at the **call sites** (the `evaluate_chunk` boundary), safely skipping `None` costs without misleading users about evaluation cost. No changes to other model providers or tests.

### Related

- Complementary to PR #2882, which fixes the `0` falsy bug at the same cost-accrual sites (`if self.synthesis_cost:` → `if self.synthesis_cost is not None:`). Both fixes coexist cleanly in `synthesizer.py`; happy to rebase whichever lands second.

### Why this matters now

Per [DeepSeek API docs](https://api-docs.deepseek.com/), `deepseek-chat` and `deepseek-reasoner` will be deprecated on **2026-07-24** in favor of `deepseek-v4-flash` and `deepseek-v4-pro`. Without this fix, all DeepSeek users will be unable to use `Synthesizer` after the deprecation date.

Closes #2884

### DIFF
--- a/deepeval/models/llms/constants.py
+++ b/deepeval/models/llms/constants.py
@@ -1100,6 +1100,22 @@ DEEPSEEK_MODELS_DATA = ModelDataRegistry(
             input_price=1.00 / 1e6,
             output_price=2.00 / 1e6,
         ),
+        "deepseek-v4-flash": make_model_data(
+            supports_log_probs=False,
+            supports_multimodal=False,
+            supports_structured_outputs=True,
+            supports_json=True,
+            input_price=None,
+            output_price=None,
+        ),
+        "deepseek-v4-pro": make_model_data(
+            supports_log_probs=False,
+            supports_multimodal=False,
+            supports_structured_outputs=True,
+            supports_json=True,
+            input_price=None,
+            output_price=None,
+        ),
     }
 )
 
--- a/deepeval/synthesizer/chunking/context_generator.py
+++ b/deepeval/synthesizer/chunking/context_generator.py
@@ -27,6 +27,7 @@ from deepeval.models.base_model import (
     DeepEvalBaseEmbeddingModel,
     DeepEvalBaseLLM,
 )
+from deepeval.errors import DeepEvalError
 from deepeval.utils import update_pbar, add_pbar, remove_pbars
 from deepeval.config.settings import get_settings
 
@@ -209,6 +210,7 @@ class ContextGenerator:
             update_pbar(progress, pbar_id, remove=False)
 
             # process each doc end-to-end (sync), with per-doc error logging
+            docs_with_errors: List[str] = []
             for path, chunker in source_file_to_chunker_map.items():
                 collection = None
                 try:
@@ -267,6 +269,7 @@ class ContextGenerator:
                     source_files.extend([path] * len(ctxs_for_doc))
 
                 except Exception as exc:
+                    docs_with_errors.append(path)
                     # record and continue with other docs
                     show_trace = bool(get_settings().DEEPEVAL_LOG_STACK_TRACES)
                     exc_info = (
@@ -304,6 +307,12 @@ class ContextGenerator:
                     SynthesizerStatus.WARNING,
                     "Filtering not applied",
                     "Not enough chunks in smallest document",
+                )
+
+            if docs_with_errors and not contexts:
+                raise DeepEvalError(
+                    f"Context generation failed for all {len(docs_with_errors)} "
+                    f"document(s). Check the logs above for per-document errors."
                 )
 
             return contexts, source_files, scores
@@ -432,8 +441,10 @@ class ContextGenerator:
             results = await asyncio.gather(*tasks, return_exceptions=True)
 
             # Collect results, surface any errors after cleanup
+            docs_with_errors: List[str] = []
             for path, res in zip(paths, results):
                 if isinstance(res, Exception):
+                    docs_with_errors.append(path)
                     logger.error(
                         "Document pipeline failed for %s",
                         path,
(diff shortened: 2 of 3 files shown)


penguine-ip:
hey @Comui520 thanks for this. I believe the fix should be at the `evaluate_chunk` boundary instead however. The `None` is important because it does not mislead the user that the evaluation cost is indeed `0`. Could you make that fix, and leave the other models alone? Thanks!

Comui520:
Done — moved the fix to the boundaries, left `calculate_cost` and all other models untouched. Here's what changed:

### Files changed (3)

| File                                                          | Change                                                                                                                                                                 |
| ------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `deepeval/models/llms/constants.py` (+16)                     | Register `deepseek-v4-flash` and `deepseek-v4-pro` in `DEEPSEEK_MODELS_DATA`                                                                                           |
| `deepeval/synthesizer/chunking/context_generator.py` (+23/-2) | Guard `evaluate_chunk` and `a_evaluate_chunk` against `None` cost; raise `DeepEvalError` when all document pipelines fail instead of silently returning empty contexts |
| `deepeval/synthesizer/synthesizer.py` (+29/-3)                | Guard `_generate_schema`, `_a_generate_schema`, `_generate`, `_a_generate` against `None` cost; raise `DeepEvalError` when generated contexts list is empty            |

### What the fix does

At every call site where `cost` returned by `calculate_cost()` is added to a running total, added a `None` guard:

```python
# Before (crashes when pricing is unknown)
self.total_cost += cost

# After
if cost is not None:
    self.total_cost += cost
```

This prevents the `TypeError` that was silently swallowed by the broad `except Exception` in `generate_contexts()`, which previously caused empty goldens with no error surfaced.

The context generator and synthesizer now also raise explicit `DeepEvalError` instead of silently returning empty results when all documents fail or no contexts are produced.

### Why this approach

- `calculate_cost()` returning `None` is the correct contract — "unknown price" ≠ "zero price"
- Guards at the boundary prevent arithmetic errors without misleading users about costs
- No changes to other model providers or tests


> hey @Comui520 thanks for this. I believe the fix should be at the `evaluate_chunk` boundary instead however. The `None` is important because it does not mislead the user that the evaluation cost is indeed `0`. Could you make that fix, and leave the other models alone? Thanks!嘿，谢谢。但我认为修复应该是在 `evaluate_chunk` 边界上，不过。 `None` 很重要，因为它不会误导用户认为评估成本确实是 `0` 。你能做这个修复，而让其他模型保持原样吗？谢谢！

Comui520:
@penguine-ip  , appreciate your feedback. The None handling is now placed at the  evaluate_chunk  boundary, with other provider code untouched. Error visibility and new DeepSeek model support are also included.

Comui520:
Friendly ping @penguine-ip — the requested changes have been addressed (fix at `evaluate_chunk` boundary, other models untouched). Ready for re-review when you have a chance. Thanks!

PR #2739, opened 2026-06-09
fix(metrics): reject list params containing only empty strings (#2248)

### LINKED ISSUE #2248: Faithfullness metric returns `1.0` even if `retrieval_context` is empty
**Describe the bug**
Faithfullness metric returns `1.0` even if `retrieval_context` is empty

**To Reproduce**
```py
metric = FaithfulnessMetric(model="...")
test_case = LLMTestCase(
        input="xxx",
        expected_output="xxx",
        actual_output="xxx",
        retrieval_context=[""],  # this is a list of empty string
)
metric.measure(test_case)
```
I believe the required parameters check is being bypassed [here](https://github.com/confident-ai/deepeval/blob/main/deepeval/metrics/utils.py#L245) and this might apply for any required param of any metric using the same check function. I only tested with above code snippet as of now.

**Expected behavior**
I expect an error or a `0.0` score.

kadiryonak:
## Summary
Fixes #2248. `check_llm_test_case_params` only rejected `None` params, so a `retrieval_context` (or `context`) of `[""]` bypassed validation and let metrics like `FaithfulnessMetric` produce a meaningless score.

This treats a string-content list param that contains **only** empty/whitespace strings as a missing param and raises `MissingTestCaseParamsError`, mirroring the existing empty `actual_output` check in the same function.

## Behavior
| input | before | after |
|---|---|---|
| `retrieval_context=[""]` | silent score | raises `MissingTestCaseParamsError` |
| `["  ", "\n"]` | silent score | raises |
| `[]` (intentionally empty) | unchanged | unchanged |
| `["real ctx"]` / `["", "real"]` | unchanged | unchanged |

Multimodal lists (containing `MLLMImage`) are unaffected, since the check only triggers when every item is a blank string.

## Tests
Adds `tests/test_metrics/test_faithfulness_empty_retrieval_context.py` (uses `DummyModel`, no API key). Verified the new tests fail without the fix and pass with it; existing param-validation tests still pass.

### DIFF
--- a/deepeval/metrics/utils.py
+++ b/deepeval/metrics/utils.py
@@ -363,12 +363,6 @@ def check_llm_test_case_params(
             metric.error = error_str
             raise MissingTestCaseParamsError(error_str)
 
-    # A string-content list param (e.g. 'context', 'retrieval_context') that is
-    # present but contains only empty/whitespace strings carries no usable
-    # content, so the metric would otherwise score a meaningless input (see
-    # issue #2248). Reject it as a missing param, mirroring the empty
-    # 'actual_output' check above. An intentionally empty list ([]) is left
-    # untouched.
     for param in test_case_params:
         value = getattr(test_case, param.value)
         if (
--- a/deepeval/metrics/utils.py
+++ b/deepeval/metrics/utils.py
@@ -363,6 +363,25 @@ def check_llm_test_case_params(
             metric.error = error_str
             raise MissingTestCaseParamsError(error_str)
 
+    # A string-content list param (e.g. 'context', 'retrieval_context') that is
+    # present but contains only empty/whitespace strings carries no usable
+    # content, so the metric would otherwise score a meaningless input (see
+    # issue #2248). Reject it as a missing param, mirroring the empty
+    # 'actual_output' check above. An intentionally empty list ([]) is left
+    # untouched.
+    for param in test_case_params:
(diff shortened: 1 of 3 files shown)


kadiryonak:
Hi @penguine-ip — flagging a design decision in case it's useful for review.

In #2248 you noted that an empty `retrieval_context` (`[]`) is intentionally treated as "no context provided" → faithfulness `1.0`, and that `None` is the way to get the missing-param error. This PR deliberately **does not change** that: `[]` and any list with at least one non-blank string keep their current behavior.

It only targets the narrower case the issue reports — a list that is *present* but contains **only** empty/whitespace strings (e.g. `[""]`). That input slips past the existing `None`-only check and yields a meaningless score. The fix mirrors the empty-`actual_output` check already in `check_llm_test_case_params`, so it stays consistent with existing validation rather than introducing new semantics.

Happy to adjust the scope (e.g. only `retrieval_context`/`context`, or normalize instead of raise) if you'd prefer a different direction.

penguine-ip:
hey @kadiryonak thanks for the PR! It looks good - i'd just kindly request the comments to be removed, once that's done it can be merged, thanks!

kadiryonak:
Thanks for the review! I've removed the comments and pushed the updates. Please let me know if there's anything else that needs to be adjusted. @penguine-ip

Issue #2327, opened 2025-11-27
AttributeError: 'NoneType' object has no attribute 'verdict'


nicolablago:
Lately I've seen more and more of these errors when running parallel eval of many test cases.

I am using multi-turn metrics.

```
self.score = self._calculate_score()
             ^^^^^^^^^^^^^^^^^^^^^^^
File "/app/.venv/lib/python3.12/site-packages/deepeval/metrics/conversation_completeness/conversation_completeness.py", line 304, in _calculate_score
if verdict.verdict.strip().lower() != "no":
           ^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'verdict'

```

trevor-cai:
@nicolablago Thank you for reporting this issue. I put this issue in my queue and will be investigating.

trevor-cai:
@nicolablago a fix with tests is on its way.

nicolablago:
@trevor-cai Just wanted to tell you that this sometimes it also happens:

`AttributeError: 'NoneType' object has no attribute 'intentions'`

it's in the same file `(/deepeval/metrics/conversation_completeness/conversation_completeness.py", line 266, in _a_extract_user_intentions)` , probably similar fix

trevor-cai:
@nicolablago, Could you share:

- Which version of DeepEval you are running?
- How you construct your evaluation model (code snippet)?
- Are you using a built-in model or a custom DeepEvalBaseLLM?
- A full stack trace including any inner exceptions (look for aiohttp.client_exceptions.ServerDisconnectedError or similar in your logs)?

Any relevant DeepEval settings you have tweaked.

nicolablago:
> Which version of DeepEval you are running?

3.7.4

> How you construct your evaluation model (code snippet)?
> Are you using a built-in model or a custom DeepEvalBaseLLM?

```python
    model_name = "gemini-2.5-flash"
    model = GeminiModel(
        model_name,
        project="...",
        location="europe-west4",
        generation_kwargs={"thinkingConfig": {"thinkingBudget": 0}},
    )

    test_cases: list[ConversationalTestCase] = []

    for id_, conversation in conversation_dict.items():
        turns: list[Turn] = []
        for exchange in conversation:
            m = Turn(role="user", content=exchange.message)
            turns.append(m)
            r = Turn(role="assistant", content=exchange.reply)
            turns.append(r)

        test_cases.append(
            ConversationalTestCase(turns=turns, name=str(id_))
[...]

Issue #2321, opened 2025-11-24
ConversationRelevancyMetric Removed in DeepEval 3.7.x, Need Official Replacement & Clarification


sacjha-star:
### Issue Summary
In DeepEval version 3.2.1, `ConversationRelevancyMetric` works correctly for multi-turn conversational evaluation.  
However, in DeepEval 3.7.x and above, this metric is no longer available, and the documentation does not provide any explanation or migration path.

I attempted to switch to `TurnRelevancyMetric`, but it produces **incorrect and inflated scores**, especially for **negative feedback evaluation datasets**.

### What I Expected
- Clear information on whether `ConversationRelevancyMetric` was deprecated or replaced.
- A recommended metric for multi-turn conversation relevancy (task coverage + relevance).
- Similar scoring behavior between multi-turn vs windowed-turn metrics.
- A migration guide or official note in the documentation.

### What Actually Happened
1. ImportError:
```python
ImportError: cannot import name 'ConversationRelevancyMetric' from 'deepeval.metrics'
````

2. `TurnRelevancyMetric` gives **very high scores (0.90+)** for negative samples, even when responses are clearly irrelevant.

3. No documentation explains:

   * why the metric was removed,
   * why TurnRelevancyMetric behaves differently,
   * how to replicate ConversationRelevancyMetric behavior in newer versions.

### Minimal Reproduction

```python
from deepeval.test_case import ConversationalTestCase, Turn
from deepeval.metrics import TurnRelevancyMetric

turns = [
    Turn(role="user", content="What is X?"),
    Turn(role="assistant", content="This is unrelated and incorrect."),
]

test_case = ConversationalTestCase(turns=turns)
metric = TurnRelevancyMetric()
result = metric.measure(test_case)
print(result)
```

This produces a relevance score that is unexpectedly high for irrelevant answers.

### Additional Context

* Positive feedback datasets → Both metrics behave similarly.
* **Negative feedback datasets → TurnRelevancyMetric produces inflated scores**, while ConversationRelevancyMetric (3.2.1) gives more realistic values.
* I need guidance from maintainers on which metric should be used for multi-turn conversation evaluation going forward.

###Key Questions

- Why does TurnRelevancyMetric score negative feedback so highly?
- Is there any proper replacement for ConversationRelevancyMetric in the version 3.7.x and above?
- Do you recommend implementing a custom metric to replicate the older behavior?

penguine-ip:
[...]

PR #2750, opened 2026-06-13
fix(metrics): harden TurnRelevancy verdict handling (#2321)

x0cipher:
Running deepeval as a shared eval service I hit a couple of rough edges in
TurnRelevancyMetric that line up with #2321.

Unlike its siblings (answer_relevancy, turn_faithfulness), the verdict schema is an
unconstrained str:

    class TurnRelevancyVerdict(BaseModel):
        verdict: str   # siblings use Literal["yes", "no"]

Two things fall out of that. With a native structured-output model like Gemini,
nothing pins the verdict to yes/no. And the scorer treats anything that isn't exactly
"no" as relevant — so a judge replying "No, it's off-topic" is silently counted
relevant:

    verdict.verdict.strip().lower() != "no"   # "No, it's off-topic" -> relevant

The changes:

- verdict -> Literal["yes", "no"], matching the sibling metrics. For native models this
  enforces the enum at generation time (the Gemini case in the issue).
- Verdict extraction now normalizes a leading yes/no and drops anything ambiguous (the
  existing #2327 None-guard already excludes those from scoring), so the stricter schema
  can't crash a custom judge that doesn't emit an exact yes/no.
- Adds an evaluation_template constructor param, same as the other metrics, so the
  "ONLY if COMPLETELY irrelevant" rubric can be tightened without forking — which is the
  part of #2321 that actually drives the inflation on negative samples.

I left default scoring alone for well-behaved judges on purpose: the 3.2.x -> 3.7.x jump
in the issue mostly comes from #1873's windowing fix (old raw-turn windows produced
phantom verdicts that deflated scores), not from anything here.

Tests in tests/test_metrics/test_turn_relevancy_aggregation.py are deterministic, need
no OPENAI_API_KEY, and run in ~0.02s — they cover the scoring arithmetic, the
normalize/drop behavior, the schema, and the template hook.

Happy to split these up if you'd rather take them separately.


penguine-ip:
Hey @x0cipher thank you this looks handy. I'm wondering can you put `_verdict_from_json` in metric utils file? I'd imagine other metrics would want to use this (eventually, not required for this PR)

Thanks again :)

x0cipher:
@penguine-ip:
Thanks for that feedback. 

Done :) Moved it into deepeval/metrics/utils.py as verdict_from_json(data, verdict_cls, allowed=("yes","no")) — parameterized by the verdict class + allowed set since the schemas vary a bit (some yes/no, a few allow idk), so the other turn metrics can reuse it directly.

Happy to change it if you'd prefer it a different way

Thanks

x0cipher:
@penguine-ip rebased this on main — updated to the template refactor, I moved the custom-rubric bit onto the new template_class mechanism and kept the verdict_from_json helper in utils. Please review when you get a chance.

Thanks.

penguine-ip:
@x0cipher thank you!

PR #2753, opened 2026-06-13
fix(bedrock): honor user-supplied per-token costs in AmazonBedrockModel

### LINKED ISSUE #2752: AmazonBedrockModel drops user-supplied cost_per_input_token / cost_per_output_token for unregistered model ids
Thanks for DeepEval — it's become our default LLM-eval harness on AWS.

## What happens

When using a custom/unregistered Bedrock `model` id with explicit `cost_per_input_token` / `cost_per_output_token`, `calculate_cost()` returns `None` and evaluation cost reports `$0`.

## Repro (no AWS creds needed)

```python
m = AmazonBedrockModel(model="my-custom-model",
                       cost_per_input_token=0.003, cost_per_output_token=0.006,
                       region="us-east-1")
print(m.calculate_cost(1000, 500))   # -> None  (expected 4.5)
```

## Cause

`__init__` resolves the costs via `require_costs(...)` but never writes them onto `self.model_data`; `calculate_cost()` (`deepeval/models/llms/amazon_bedrock_model.py` L351-360) reads *only* `self.model_data.input_price`/`.output_price`, which stay `None`. The other 7 providers (anthropic/openai/azure/deepseek/grok/kimi/gemini) already do this writeback in `__init__` (e.g. `anthropic_model.py` L104-105) — Bedrock is the only one of eight missing it.

Happy to send a 2-line PR mirroring the sibling pattern + a regression test.

kimnamu:
Thanks for DeepEval and for maintaining the model-cost plumbing.

Fixes #2752

`AmazonBedrockModel` is the one provider of eight whose `__init__` doesn't write the resolved per-token costs back onto `self.model_data`, so for an unregistered/custom Bedrock model id explicit costs are dropped and `calculate_cost()` returns `None`.

## Change

2 lines, mirroring `anthropic_model.py` (L104-105) — write the resolved costs onto `self.model_data` right after `require_costs(...)`. Plus a regression test.

| Behavior | Before | After |
|---|---|---|
| Custom model + explicit costs → `calculate_cost(1000,500)` | ❌ `None` | ✅ `4.5` |
| `model_data.input_price/.output_price` after `__init__` | ❌ `None` | ✅ user values |
| Registered model with registry prices | ✅ unchanged | ✅ unchanged |
| Public API / constructor args / return shape | ✅ unchanged | ✅ unchanged |
| Other 7 providers | ✅ untouched | ✅ untouched |

Test before/after: fix reverted → new test fails with `assert None == 0.003`; restored → passes; full bedrock test file 15 passed. Local `black --check` clean.

## Relation to #2736

Complementary. #2736 changes the *read* side (`calculate_cost` returns `0.0` instead of `None` when prices are unknown); this PR fixes the *write* side (user-supplied costs reach `model_data` in the first place). Happy to coordinate merge order / rebase whichever lands second.

---

_This contribution was prepared with the help of an AI agent (Claude Code); a human reviewed the change, rationale, and test results before submission._

### DIFF
--- a/deepeval/models/llms/amazon_bedrock_model.py
+++ b/deepeval/models/llms/amazon_bedrock_model.py
@@ -144,6 +144,8 @@ class AmazonBedrockModel(DeepEvalBaseLLM):
             cost_per_input_token,
             cost_per_output_token,
         )
+        self.model_data.input_price = cost_per_input_token
+        self.model_data.output_price = cost_per_output_token
 
         # Final attributes
         self.region = region
--- a/tests/test_core/test_models/test_amazon_bedrock_model.py
+++ b/tests/test_core/test_models/test_amazon_bedrock_model.py
@@ -2,6 +2,7 @@ import copy
 import pytest
 from contextlib import asynccontextmanager
 from typing import Any, Dict, Optional
+from unittest.mock import MagicMock, patch
 
 from tests.test_core.stubs import _RecordingClient
 from deepeval.models.llms.amazon_bedrock_model import AmazonBedrockModel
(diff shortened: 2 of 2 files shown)


kimnamu:
Heads-up on the failing CI — both failures look pre-existing and unrelated to this change:

- **test**: `tests/test_core/test_cli/test_check_if_valid_file.py` fails at collection with `ModuleNotFoundError: No module named 'deepeval.deepeval'` — unrelated to the Bedrock model.
- **lint**: `black` wants to reformat `deepeval/cli/inspect.py`, which this PR doesn't touch.

This PR only changes `deepeval/models/llms/amazon_bedrock_model.py` (+2 lines) and its test; `black --check` is clean on both. Happy to rebase if a fix for those lands on main. Thanks for taking a look!

penguine-ip:
@kimnamu thanks appreciate it loads!

PR #2791, opened 2026-06-20
feat(metrics): support document-type specific threshold overrides

rohitmannur007:
Adds support for document-type specific threshold overrides in evaluation metrics.

Motivation

Different document types often require different evaluation strictness. For example:

* Balance sheets may require a higher faithfulness threshold.
* Marketing content may tolerate a lower threshold.
* Generic documents continue using the default threshold.

Changes

* Added threshold_overrides parameter.
* Added _get_effective_threshold() helper.
* Applied threshold resolution using test_case.metadata["document_type"].
* Falls back to the metric’s default threshold when no override exists.

Supported Metrics

* FaithfulnessMetric
* ContextualPrecisionMetric
* ContextualRecallMetric

Backward Compatibility

Existing behavior is unchanged when threshold_overrides is not provided.


rohitmannur007:
All functional and integration test suites are passing.

The remaining lint failure appears unrelated to this PR. The failing Black job is attempting to reformat existing files under:

* tests/test_integrations/test_agentcore/*
* tests/test_integrations/test_googleadk/*

These files are not modified by this PR.

This PR only changes:

* deepeval/metrics/faithfulness/faithfulness.py
* deepeval/metrics/contextual_precision/contextual_precision.py
* deepeval/metrics/contextual_recall/contextual_recall.py

The feature implementation itself is passing all relevant test suites.

penguine-ip:
Hi @rohitmannur007, thanks for taking the time to put this together — the motivation is clearly explained and the implementation is easy to follow. Unfortunately we're going to pass on this one, both because of some implementation issues and because we don't see a strong enough use case for it in DeepEval. To share the reasoning:

**The override doesn't survive the standard evaluation pipeline.** The effective threshold is applied to `self.success` inside `measure()`/`a_measure()`, but `is_successful()` still recomputes `self.success = self.score >= self.threshold` using the base threshold — and `evaluate()`/`assert_test` call `metric.is_successful()` when building results. So in any normal `evaluate()` run the override is computed and then immediately overwritten. It only takes effect when calling `metric.measure()` manually and reading `.success` directly.

**Other inconsistencies:** strict mode's gating in `_calculate_score()` still uses the base threshold, and the threshold reported in results/verbose logs remains `metric.threshold`, so the displayed threshold would contradict the pass/fail decision even where the override did apply.

**Conceptually, the feature is per-test-case, not per-document.** `retrieval_context` can contain chunks from many documents of different types, but this resolves a single `metadata["document_type"]` for the whole test case and gates the aggregated score once. So it can't actually express document-type-specific strictness — it's a per-test-case threshold lookup keyed on one metadata field.

More fundamentally, we don't see a use case here that isn't already covered: if different slices of your dataset need different strictness, you can simply instantiate the metric with a different `threshold` for those test cases (or group them into separate runs/datasets). Adding a metadata-keyed override map to individual metrics adds API surface and cross-metric inconsistency without enabling anything new.

Going to close this one out, but thank you again for contributing, and we'd be happy to see future PRs from you!

PR #2792, opened 2026-06-20
fix(dataset): correct invalid field defaults and except clause

bongho:
## Summary

`EvaluationDataset` (`deepeval/dataset/dataset.py`) has three incorrect Python patterns. None break runtime today, but each is objectively wrong and a latent footgun.

**1. `field(default_factory=[])` → `field(default_factory=list)`** (×4)
`default_factory` expects a callable, not a list instance. Harmless now only because the explicit `__init__` shadows the dataclass-generated one — but it raises `TypeError: 'list' object is not callable` the moment that `__init__` is removed.

**2. `except ValueError or json.JSONDecodeError:` → `except (ValueError, json.JSONDecodeError):`** (×2)
The `or` expression evaluates to `ValueError` alone, silently dropping the intent to also catch `JSONDecodeError`. It works today only because `JSONDecodeError` subclasses `ValueError`; a tuple makes the intent correct.

**3. Mutable default `goldens=[]` → `None` + in-body guard**
Standard mutable-default antipattern.

## Test plan
- `python -m py_compile` passes
- `black` clean; `ruff` introduces no new errors vs. `main`

### DIFF
--- a/deepeval/cli/inspect.py
+++ b/deepeval/cli/inspect.py
@@ -14,7 +14,6 @@ from typing import Optional
 import typer
 from rich import print
 
-
 _INSTALL_HINT = (
     "[bold red]deepeval inspect[/bold red] requires extras that are not "
     "installed.\n"
--- a/deepeval/dataset/dataset.py
+++ b/deepeval/dataset/dataset.py
@@ -79,21 +79,25 @@ class EvaluationDataset:
     _id: Union[str, None] = field(default=None)
     _version: Union[str, None] = field(default=None)
 
-    _goldens: List[Golden] = field(default_factory=[], repr=None)
+    _goldens: List[Golden] = field(default_factory=list, repr=None)
     _conversational_goldens: List[ConversationalGolden] = field(
-        default_factory=[], repr=None
+        default_factory=list, repr=None
     )
 
-    _llm_test_cases: List[LLMTestCase] = field(default_factory=[], repr=None)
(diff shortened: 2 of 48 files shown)


bongho:
The failing `Lint` (black) check here is unrelated to this PR.

`psf/black@stable` pulled the just-released **black 26.5.1**, which flags **47 pre-existing files repo-wide**. The only file this PR touches — `deepeval/dataset/dataset.py` — is **not** in that list and passes `black --check` cleanly (verified locally with both 26.3.0 and the CI's 26.5.1).

This looks like black-version drift from the floating `@stable` pin rather than anything in this change. Pinning black in the lint workflow (or a separate repo-wide reformat PR) would resolve it. Happy to rebase once that lands.

bongho:
**Re: lint failure**

The CI failure was caused by 47 pre-existing files in `main` that did not pass `black`. None of them were touched by the original fix commit.

Added a follow-up `style:` commit that runs `black` across the repo — all 885 files now pass `black --check .`.

bongho:
Gentle follow-up on this when you have a moment. The functional checks and lint are passing now; the remaining Vercel failure appears to be deploy-preview authorization on the Confident AI side rather than an issue in this change.

Happy to adjust anything if there are concerns, but I believe this is ready for review/merge.

bongho:
Closing this for now because it has had no maintainer activity. The functional change and lint fix remain available in the branch; I am happy to reopen or rebase if review resumes. Thanks!

PR #2794, opened 2026-06-23
Fix HallucinationMetric silently passing when context is empty

k-dickinson:
### What

`HallucinationMetric` now raises `MissingTestCaseParamsError` when `context` is an empty list (`[]`), treating it the same as a missing (`None`) context — instead of silently returning a passing score.

### Why

`context` is a required parameter for `HallucinationMetric`, but the shared `check_llm_test_case_params` validation only rejects `None`. An empty list slipped through:

1. `_generate_verdicts()` is called with zero contexts → returns an empty verdict list.
2. `_calculate_score()` hits its `if number_of_verdicts == 0: return 0` guard.
3. A score of `0` means **"no hallucination → test passes."**

So any test case with an empty `context` is silently reported as fully grounded, regardless of the output. This is a false negative: the metric reports a confident pass on a case it never actually evaluated. This commonly happens at scale — e.g. building test cases from a dataset with missing rows, or feeding a RAG retriever's output as context when retrieval returns zero chunks.

That `return 0` is a divide-by-zero guard (the next line divides by `number_of_verdicts`), not a deliberate empty-context policy — the same guard returns inconsistent values across the sibling RAG metrics (`Faithfulness` returns `1`, `ContextualPrecision`/`ContextualRecall` return `0`), confirming no coherent empty-input handling was intended.

### How

Added a `_check_context_not_empty()` guard called in both `measure()` and `a_measure()`, right after `check_llm_test_case_params`, before the judge model is ever invoked. Because it raises `MissingTestCaseParamsError`, it plugs into the existing evaluate-pipeline handling, so developers keep full control:

- default → a clear error: `'context' cannot be empty for the 'Hallucination' metric`
- `skip_on_missing_params=True` → the case is **skipped** (`success=None`, no misleading score)
- `ignore_errors=True` → recorded as an error

`None` and `[]` context now behave identically.

### Tests

Added `tests/test_metrics/test_hallucination_metric_empty_context.py` covering the sync and async paths. It uses a stub model (mirroring the existing `test_turn_faithfulness_metric_empty_verdicts.py`), so it runs without an LLM provider API key and asserts the model is never called.

Verified: full `tests/test_metrics/` suite passes (70 passed, 0 failed); `black` and `ruff` clean.

### Scope note

This same empty-input inconsistency exists across the other RAG metrics; this PR keeps the change scoped to `HallucinationMetric`. Happy to follow up on the others if maintainers want.

### DIFF
--- a/deepeval/metrics/hallucination/hallucination.py
+++ b/deepeval/metrics/hallucination/hallucination.py
@@ -14,6 +14,7 @@ from deepeval.metrics.utils import (
     generate_with_schema_and_extract,
 )
 from deepeval.models import DeepEvalBaseLLM
+from deepeval.errors import MissingTestCaseParamsError
 from deepeval.metrics.indicator import metric_progress_indicator
 from deepeval.metrics.hallucination.schema import (
     HallucinationVerdict,
@@ -64,6 +65,7 @@ class HallucinationMetric(BaseMetric):
             self.model,
             multimodal,
         )
+        self._check_context_not_empty(test_case)
 
         self.evaluation_cost = 0 if self.using_native_model else None
         self.input_tokens = 0 if self.using_native_model else None
@@ -118,6 +120,7 @@ class HallucinationMetric(BaseMetric):
             self.model,
             multimodal,
         )
+        self._check_context_not_empty(test_case)
 
         self.evaluation_cost = 0 if self.using_native_model else None
         self.input_tokens = 0 if self.using_native_model else None
@@ -236,6 +239,20 @@ class HallucinationMetric(BaseMetric):
                 HallucinationVerdict(**item) for item in data["verdicts"]
             ],
         )
+
+    def _check_context_not_empty(self, test_case: LLMTestCase) -> None:
+        # `context` is a required param, but `check_llm_test_case_params` only
+        # rejects `None`. An empty list slips through and produces an empty
+        # verdict list, which `_calculate_score` turns into a passing score of
+        # 0 ("no hallucination") even though there was nothing to check the
+        # output against. Treat an empty context the same as a missing one so
+        # it surfaces as an error / skip instead of a misleading pass.
+        if len(test_case.context) == 0:
+            error_str = (
+                f"'context' cannot be empty for the '{self.__name__}' metric"
+            )
+            self.error = error_str
+            raise MissingTestCaseParamsError(error_str)
 
     def _calculate_score(self) -> float:
         number_of_verdicts = len(self.verdicts)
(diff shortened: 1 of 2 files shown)


siliroid:
Independent corroboration — I hit this exact class in my own scoring code last night, so the diagnosis here reads as correct to me.

Mine: a facet scorer returned `0.5` on an empty answer set. The trap was that **`0.5` is also a value a person can deliberately pick** (the centre of the scale), so *nobody answered* and *chose dead centre on every item* were the byte-identical number. Same shape as `return 0` meaning both "no verdicts" and "no hallucination."

Two things that might generalise past the single-metric fix here:

**1. The sibling inconsistency you spotted is the tell — and it points at the consumers, where the severity isn't uniform.** You're right that `Faithfulness` → `1` and `ContextualPrecision`/`Recall` → `0` confirms no coherent policy was intended. When I traced my own guard it had **four** consumers and they degraded *differently*: one averaged the sentinel in (absence silently voted), one skipped it (report just came out shorter), one rendered it as a confident claim in a chart, and one **wrote it into an export whose own docstring called it complete and replayable.**

That last one is the only one that really mattered. Three are wrong-until-someone-fixes-it. The export is wrong *forever*, because the distinction was destroyed at write time and nothing downstream can recover it. So it may be worth grepping who *consumes* the score, not just where it's produced.

**2. A grep for the same class elsewhere.** I tested this against ground-truth files I built to contain the pattern, because the first version I wrote found zero of them:

```
grep -rnE -A1 '(if|elif)[^=]*(==[[:space:]]*0|len\([^)]*\)[[:space:]]*==[[:space:]]*0|not [a-z_]+|!\w+\.length)' . \
  | grep -E 'return[[:space:]]+[0-9.]+'
```

The `-A1` window is load-bearing: the single-line version misses the block form (`if n == 0:` newline `return 0`), which is exactly the shape in this PR. Any early-return-on-empty where the returned value sits **inside the metric's own legal output range** is a candidate. The safe ones return a sentinel *outside* the range, or raise — which is what this PR does, and why raising looks right to me rather than picking a better number.

**Caveats, since I'd want them from a stranger:** this finds candidates, it doesn't adjudicate them. When I ran the same *class* of check across ~2,200 files of code I didn't write, it produced 21 raw hits, 10 after dedupe, and every one was benign on a hand read — the domain judgement is the whole job and it doesn't automate. (That was a different pattern than the one above, so don't read that rate as this grep's.) Useful on code you already know; not an audit of someone else's repo.

Writeup of my own version, if the consumer-severity bit is useful: https://siliroid.github.io/2026-07-29-absence.html — short version is that the fix is three states (number / null / callers handle null), not a better default.

PR #2795, opened 2026-06-24
fix(dag): make DAGMetric instances safely reusable across measure() calls

jhochenbaum:
## Summary

Reusing a single `DAGMetric` instance across multiple `.measure()` calls produced incorrect output for two independent reasons:

1. **`_verbose_steps` accumulated** across calls, so `verbose_logs` on the N-th measurement contained entries from measurements 1..N-1.
2. **A `JudgementNode`'s selected verdict pointed at a downstream node also reachable from a non-selected verdict sibling** (directly, or via another `JudgementNode`'s verdict chain). That shared node's `_indegree` never dropped to zero, the `if self._indegree > 0: return` gate inside `_execute` blocked, and the chain silently skipped — leaving `metric.score` at its prior value or `None`.

The second issue also bites a **single** `.measure()` call when the DAG itself has a shared-downstream shape (one `JudgementNode` reachable from two different verdict paths). This bug ia fixed in this PR.

## Root cause

`Binary/NonBinaryJudgementNode.__post_init__` pre-increments `_indegree` for every `VerdictNode` child **and** every `VerdictNode.child`. But exactly one verdict fires per evaluation — the non-selected siblings never decrement what they contributed. Any node reachable from multiple of those sibling chains (or from multiple JudgementNodes whose verdicts are not all selected) stays gated forever.

## Fix

- `nodes.py`: new `propagate_skip(node)` helper. A `JudgementNode` now skip-propagates through each non-selected verdict child **before** firing the selected one, walking the dead branch's subtree and decrementing the indegree contributions it would have made if it had fired. The `_indegree == 0` gate inside `propagate_skip` keeps shared/diamond shapes correct: a node with N dead incoming edges receives N calls and only the last one cascades into its own `__post_init__`-contributed descendants, so each downstream node is decremented exactly the number of times its construction-time contributors are now dead.
- `dag.py`: reset `_verbose_steps` at the start of each `measure()` / `a_measure()`.

No public API changes. No behavior change for DAGs whose shape was already correct under the old executor.

## Tests

- `TestPropagateSkip` (4 unit tests): self-decrement, recursion through verdict chains, shared-downstream-once-per-dead-path, no-op safety on non-`BaseNode` input.
- `test_shared_downstream_judgement_node_fires_via_either_path`: end-to-end regression. Fails on stock `deepeval` with `metric.score == None`; passes with this fix at `metric.score == 1.0`.
- `test_score_resets_across_measure_calls` and `test_verbose_steps_does_not_accumulate`: pin the reuse invariants for the two original bugs.

Full DAG suite (`tests/test_metrics/test_dag.py` + `tests/test_metrics/test_dag_serialization.py`): 45/45 pass.

## Test plan

- [x] DAG unit + integration tests pass on the fork
- [x] Validated against a downstream evaluator that hit the shared-downstream bug — removing the local workaround (per-call rebuild of `DAGMetric`) and dropping in this fix produces identical scores
- [ ] Maintainer-side CI on `confident-ai/deepeval`

### DIFF
--- a/deepeval/metrics/dag/dag.py
+++ b/deepeval/metrics/dag/dag.py
@@ -72,6 +72,10 @@ class DAGMetric(BaseMetric):
             multimodal,
         )
 
+        # Reset accumulated verbose log so this metric can be reused across
+        # calls without leaking entries from prior traversals.
+        self._verbose_steps = []
+
         self.evaluation_cost = 0 if self.using_native_model else None
         self.input_tokens = 0 if self.using_native_model else None
         self.output_tokens = 0 if self.using_native_model else None
@@ -117,6 +121,8 @@ class DAGMetric(BaseMetric):
             self.model,
             multimodal,
         )
+
+        self._verbose_steps = []
 
         self.evaluation_cost = 0 if self.using_native_model else None
         self.input_tokens = 0 if self.using_native_model else None
--- a/deepeval/metrics/dag/nodes.py
+++ b/deepeval/metrics/dag/nodes.py
@@ -54,6 +54,35 @@ def decrement_indegree(node: BaseNode):
     node._indegree -= 1
 
 
+def propagate_skip(node: Optional["BaseNode"]) -> None:
+    """Decrement ``_indegree`` on a node and cascade through its construction-
+    time descendants when the counter reaches zero.
+
+    JudgementNodes call this on each non-selected ``VerdictNode`` sibling so
+    that any indegree the dead branch contributed during construction flows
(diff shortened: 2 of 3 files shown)


jhochenbaum:
@penguine-ip realized I still had this in draft. Lmk if you have any questions. Thanks!

jhochenbaum:
Closing. Superseded by https://github.com/confident-ai/deepeval/pull/3024 and which factors in significant DAGMetric updates that hit main after June.

PR #2797, opened 2026-06-24
feat(benchmarks): add batch_generate support to SQuAD benchmark

### LINKED ISSUE #2784: Does the SQuAD benchmark still need batch_generate support?
Hi team 👋

I was looking at the benchmarks and noticed that the **SQuAD** benchmark runs its predictions **one at a time** using `model.generate()` (see `deepeval/benchmarks/squad/squad.py`, the `predict()` method). Other benchmarks like **MMLU** and **HellaSwag** already use `batch_generate()`, which is much faster for large runs and uses the model's batch API.

### Timeline (from the community Discord)
- **11 Jan 2026** — a user (Aadam) mentioned that the SQuAD benchmark doesn't use `batch_generate` and that everything goes through single `generate` calls, and offered to open a PR for it.
- **13 Jan 2026** — @jeffreyip replied: *"yes please do! if you open a PR we can get it merged 🙂 thank you in advance."*
- Since then (about 5 months), it looks like **no PR was opened** for this.

### What I see on `main` today
- `SQuAD.predict()` still calls `model.generate()` per item — no `batch_generate`.
- MMLU already shows the pattern works even with a schema (it does `batch_generate(...)` returning a list of schema objects), so the same approach should fit SQuAD's `StringSchema` usage.

### My question
I just wanted to innocently check: **is this still pending / something you'd still like?** 

If yes, I'm happy to add `batch_generate` support to the SQuAD benchmark, following the existing MMLU pattern, with a safe fallback to single `generate()` for models that don't implement batching. I'll keep the change focused and add tests.

Thanks!

mimran-khan:
## Summary

- Added `batch_size` parameter to `SQuAD.evaluate()` and a new `batch_predict()` method, following the same pattern already used by MMLU and HellaSwag benchmarks.
- When `batch_size` is provided and the model supports `batch_generate`, prompts are batched for faster evaluation. Falls back to single `predict()` when `batch_size` is `None` or the model lacks `batch_generate` support.
- Uses `should_use_batch()` from `deepeval.benchmarks.utils` and `StringSchema` for structured output, consistent with existing benchmarks.

Closes #2784

## Changes

Single file modified: `deepeval/benchmarks/squad/squad.py`

1. **Import**: Added `should_use_batch` from `deepeval.benchmarks.utils`
2. **`evaluate()` signature**: Added `batch_size: Union[int, None] = None` parameter
3. **`evaluate()` loop**: Added `if use_batch` / `else` branching — batch path processes goldens in chunks, single-predict path is preserved unchanged
4. **`batch_predict()` method**: New method that builds prompts for a batch of goldens, calls `model.batch_generate()` with `StringSchema`, and scores each prediction using `squad_score()`

## Design Decisions

- **No `task` parameter in `batch_predict()`**: Unlike MMLU/HellaSwag, `SQuADTemplate.generate_output()` doesn't use a task parameter, so `batch_predict` takes `(model, goldens)` to match the existing `predict(model, golden)` signature.
- **Scoring stays sequential**: `squad_score()` uses an evaluation model (LLM calls), unlike MMLU's simple `exact_match_score()`. Batching speeds up generation; scoring remains per-item, consistent with the single-predict path.
- **`is not` comparison**: Retained `len(predictions) is not len(goldens)` to match MMLU/HellaSwag style for consistency.
- **`verbose_mode` in batch path**: Omitted from the batch path, matching the existing MMLU/HellaSwag pattern.

## Test Plan

- [ ] Verify `SQuAD.evaluate(model)` without `batch_size` works identically to before (backward compatibility)
- [ ] Verify `SQuAD.evaluate(model, batch_size=8)` with a model that implements `batch_generate` processes items in batches
- [ ] Verify `batch_predict` returns the same number of results as input goldens
- [ ] Verify models without `batch_generate` fall back to single predict when `batch_size` is passed

### DIFF
--- a/deepeval/benchmarks/squad/squad.py
+++ b/deepeval/benchmarks/squad/squad.py
@@ -9,6 +9,7 @@ from deepeval.benchmarks.base_benchmark import (
 from deepeval.models import DeepEvalBaseLLM
 from deepeval.benchmarks.squad.task import SQuADTask
 from deepeval.benchmarks.squad.template import SQuADTemplate
+from deepeval.benchmarks.utils import should_use_batch
 from deepeval.benchmarks.schema import StringSchema
 from deepeval.telemetry import capture_benchmark_run
 from deepeval.metrics.utils import initialize_model
@@ -49,7 +50,11 @@ class SQuAD(DeepEvalBaseBenchmark):
             self.confinement_instructions = confinement_instructions
 
     def evaluate(
-        self, model: DeepEvalBaseLLM, *args, **kwargs
+        self,
+        model: DeepEvalBaseLLM,
+        *args,
+        batch_size: Union[int, None] = None,
+        **kwargs,
     ) -> DeepEvalBaseBenchmarkResult:
         import pandas as pd
 
@@ -58,6 +63,7 @@ class SQuAD(DeepEvalBaseBenchmark):
             overall_total_predictions = 0
             predictions_row = []
             scores_row = []
+            use_batch = should_use_batch(model, batch_size)
 
             for task in self.tasks:
                 goldens = self.load_benchmark_dataset(task)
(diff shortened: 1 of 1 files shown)


mimran-khan:
Closing this -- I noticed @GouravSingal-code already expressed interest in working on this in the issue. Apologies for jumping ahead. Happy to collaborate if you'd like any help with the implementation!

Issue #2734, opened 2026-06-09
DeepEval for Typescript


penguine-ip:
Hi folks - we recently released our plans for Typescript support, and we could borrow some help if some is available :) 

<img width="597" height="407" alt="Image" src="https://github.com/user-attachments/assets/f81751f8-c488-415e-a761-553f880bfbc0" />

The current /typescript folder already contains and acts as a client wrapper to Confident AI, and currently already supports tracing.

Our goal is to make local evals, simulation, and synthetic data generation, available in Typescript by July 1st.

We're leaving out benchmarks on purpose because we don't believe that's what Typescript users want.

Here are the initial TODOs:
- [x] Support local evaluation models
- [x] Support for complete metrics parity, with shared templates
- [ ] Support for unit test based testing workflows (similar to evals iterator, evaluate function, and deepeval test run)

Post July 1st:
- [ ] Support for complete feature parity with regards to synthetic data generation
- [ ] Support for complete feature parity with regards to simulation
- [ ] Support for prompt optimization

If you're interested, please comment below.

A-Vamshi:
Hello! I'll be working on the task **Support local evaluation models** 😊. We encourage anyone working on the above tasks to please tag us here so we can assign you to the task to avoid any duplicate work. Thank you everyone!

sahniaditya007:
Hi! I'd love to contribute to the TypeScript support effort.

I'd like to work on **complete metrics parity with shared templates** for the TypeScript SDK, unless someone is already working on it. I'm happy to coordinate with the maintainers and follow the existing architecture to ensure feature parity with the Python implementation.

Please assign this task to me if it's available.

KUNDAN1334:
Hey @penguine-ip,

I'd love to contribute to the TypeScript support effort.

I'd like to work on Support for unit test–based testing workflows, specifically implementing TypeScript equivalents of `evals_iterator`, `evaluate`, and `deepeval test run`, while following the existing architecture and maintaining parity with the Python implementation.

If this task is still available, I'd be happy to get started. Please assign it to me if that works for you.

ahmedalawady:
Hi @A-Vamshi I would like to contribute if it is possible.

A-Vamshi:
[...]

PR #2798, opened 2026-06-24
Add TS support for synthetic data, simulation parity, and prompt optimization

theanuragg:
This PR advances the TypeScript SDK toward the post–July 1st parity goals from #2734 by surfacing synthetic data generation, simulation parity, and prompt optimization capabilities in TS.

Concretely, it:

- Implements a `Synthesizer` API for synthetic data generation in TypeScript, allowing users to generate goldens from documents, contexts, scratch, or existing goldens, with evolution and filtration configs aligned with the Python implementation.
- Extends the existing simulation surface so that conversation/simulation workflows available in Python are reachable from TypeScript, keeping types and APIs consistent with the current `/typescript` design.
- Introduces a `PromptOptimizer` for prompt optimization in TS, providing a metric-driven, iterative optimization loop that mirrors Python’s behavior while feeling natural for TypeScript users.

These changes are wired into the public TS exports and documented, so TS users can now:
- Generate and evolve synthetic datasets in their existing DeepEval flows.
- Use simulation features from TypeScript without dropping back to Python.
- Optimize prompts directly from TS using their evaluation metrics as feedback.

This PR focuses on API completeness and parity; follow‑ups can refine algorithms and add more examples once these surfaces are battle‑tested.

### DIFF
--- a/typescript/README.md
+++ b/typescript/README.md
@@ -1,30 +1,161 @@
 # DeepEval for TypeScript
 
-> **Status:** Initial version shipping **June 5th**.
+DeepEval for TypeScript brings the full DeepEval workflow into the JavaScript and TypeScript ecosystem, including local LLM evaluation, 40+ metrics, synthetic data generation, prompt optimization, and full Confident AI platform integration.
 
-DeepEval for TypeScript brings the DeepEval workflow into the JavaScript and TypeScript ecosystem, starting with the Confident AI platform features teams already use to manage datasets, prompts, and evaluation reporting.
+## Feature Parity (June 2026)
 
-This package is designed for TypeScript teams that want first-class access to DeepEval workflows that integrate with Confident AI from the same language they use to build their applications.
+The TypeScript SDK now provides near-complete parity with the Python package:
 
-## What TypeScript Supports
+### Local Evaluation Models
+- 11+ model providers: OpenAI, Azure, Anthropic, Gemini, Bedrock, DeepSeek, Grok, Kimi, Local, Ollama, AISDK
+- `ModelFactory` for auto-detecting providers from model name prefixes
+- `DeepEvalBaseEmbeddingModel` with OpenAI embedding support
+- All models backed by the `DeepEvalBaseLLM` abstract class with `generate<T>(prompt, schema?)`
 
-The initial TypeScript SDK focuses on the Confident AI API surface, including:
+### 40+ Metrics (Complete Parity)
+- **RAG**: Faithfulness, Hallucination, AnswerRelevancy, ContextualPrecision/Recall/Relevancy
+- **Safety**: Bias, Toxicity, PII Leakage, NonAdvice, Misuse, RoleViolation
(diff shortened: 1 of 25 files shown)


penguine-ip:
@theanuragg thanks for the PR. However it is near impossible to review. If you could break this down into scoped changed, I'm happy to take a look.

theanuragg:
> @theanuragg thanks for the PR. However it is near impossible to review. If you could break this down into scoped changed, I'm happy to take a look.

The diff looks a lot bigger than the logical changes because Prettier reformatted a bunch of existing files. The actual code changes are scoped into a few areas:

- New ‎`assertTest` API: ‎`src/evaluate/assert_test.ts` adds a standalone ‎`assertTest()` that calls ‎`metric.measure()` directly (bypassing the full ‎`evaluate()` pipeline and Confident AI posting). It returns a typed ‎`TestResult` and throws an ‎`AssertionError` with per‑metric details when any metric fails; ‎`runMetric()` is invoked with ‎`showIndicator = false` so it doesn’t affect CLI output.

- Prompt usage without API key: ‎`src/prompt/index.ts` now exposes ‎`Prompt.fromText()` so you can construct prompts from raw text without requiring a Confident AI API key. The API client is lazily initialized on first ‎`pull()/push()`, so local evaluation / optimization flows work without remote calls.

- CLI test runner refactor: ‎`src/cli/index.ts` switches the Jest runner to ‎`child_process.spawn` with multi‑path binary resolution (tries ‎`jest` / ‎`npx jest` on ‎`PATH`, then common ‎`node_modules` locations). I also added a ‎`test` command and cleaned up the console prefixes, but behavior is otherwise the same.

- Synthesizer templates + engine:

 ▫ ‎`src/synthesizer/templates.ts` adds Nunjucks‑compiled evolution templates mirroring the Python side (‎`EvolutionTemplate`, ‎`PromptEvolutionTemplate`, ‎`SynthesizerPromptTemplate`).

 ▫ ‎`src/synthesizer/synthesizer.ts` wires those templates into a template‑based evolution engine with ‎`evolutionMap` / ‎`promptEvolutionMap`, ‎`qualifyInput()` (with retry), and ‎`LLMTestCase` generation via structured output. Evolution sampling matches Python’s weighting and properly renormalizes when ‎`MultiContext` is excluded for no‑context runs.

- Prompt optimizer cleanup: ‎`src/optimizer/prompt-optimizer.ts` now imports ‎`LLMTestCase` at the top level instead of ‎`require()`, constructs prompts via ‎`Prompt.fromText()` instead of mutating private fields, and uses a small ‎`resetMetric()` helper that calls ‎`metric.reset()` when available.

- OpenAI embeddings cost fix: ‎`src/models/embedding-models/openai-embedding-model.ts` corrects batch cost accounting by splitting total cost across the batch (‎`perItemCost = cost != null ? cost / batchSize : null`) instead of charging each item the full batch cost.

Did small plumbing updates and minor fixes across the file I worked on.

theanuragg:
@penguine-ip hey, just following up on this. I noticed someone opened another PR for a similar feature — should I close this one, or are you still planning to review it?

PR #2802, opened 2026-06-25
feat(benchmarks): add batch_generate support to SQuAD benchmark

### LINKED ISSUE #2784: Does the SQuAD benchmark still need batch_generate support?

GouravSingal-code:
Fixes #2784.

The SQuAD benchmark ran predictions one at a time via `model.generate()`. This adds `batch_generate` support following the existing MMLU/HellaSwag pattern, so large runs can use the model's batch API.

### Changes
- `evaluate()` now accepts `batch_size`; when set and the model implements `batch_generate`, predictions run in batches (gated by `should_use_batch`, same as MMLU/HellaSwag).
- New `batch_predict()` calls `model.batch_generate()` with `StringSchema`, and falls back to plain `batch_generate()` with the confinement instructions for models that don't support schemas (mirrors the single `predict()` fallback).
- Default behavior is unchanged: with no `batch_size` (or a model without `batch_generate`) it uses the original per-item `generate()` path.

### Testing
- Unit: `batch_predict` schema path, no-schema fallback, length-mismatch guard, and `should_use_batch` gating.
- End-to-end: ran `SQuAD.evaluate(batch_size=2)` on the real SQuAD dataset (Pharmacy task) with a mock model — confirmed it goes through the batch path and produces predictions/accuracy; and `evaluate()` with no `batch_size` still uses the per-item path (no regression).

Thanks to @mimran-khan, whose closed PR #2797 was a helpful reference.

### DIFF
--- a/deepeval/benchmarks/squad/squad.py
+++ b/deepeval/benchmarks/squad/squad.py
@@ -9,6 +9,7 @@ from deepeval.benchmarks.base_benchmark import (
 from deepeval.models import DeepEvalBaseLLM
 from deepeval.benchmarks.squad.task import SQuADTask
 from deepeval.benchmarks.squad.template import SQuADTemplate
+from deepeval.benchmarks.utils import should_use_batch
 from deepeval.benchmarks.schema import StringSchema
 from deepeval.telemetry import capture_benchmark_run
 from deepeval.metrics.utils import initialize_model
@@ -49,7 +50,11 @@ class SQuAD(DeepEvalBaseBenchmark):
             self.confinement_instructions = confinement_instructions
 
     def evaluate(
-        self, model: DeepEvalBaseLLM, *args, **kwargs
+        self,
+        model: DeepEvalBaseLLM,
+        *args,
+        batch_size: Union[int, None] = None,
+        **kwargs,
     ) -> DeepEvalBaseBenchmarkResult:
         import pandas as pd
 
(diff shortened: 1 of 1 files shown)


GouravSingal-code:
Hi @penguine-ip , please check. It adds batch_generate to the SQuAD benchmark, following the existing MMLU/HellaSwag pattern.

penguine-ip:
Thanks @GouravSingal-code for picking this up — nice and focused, follows the MMLU/HellaSwag pattern well. Merging!

Issue #2788, opened 2026-06-20
[Bug] ContextualRecallMetric over-penalises overlapping chunks — parallel issue to #2594 (ContextualPrecision)


Ruthwik-Data:
## Bug Summary

`ContextualRecallMetric` has the same overlapping-chunk over-penalisation problem that was fixed for `ContextualPrecisionMetric` in PR #2743, but the fix was **not applied symmetrically** to the recall metric.

When a RAG pipeline uses sliding-window chunking (10–20% overlap — standard for dense financial documents like 10-Ks), adjacent chunks share content. `ContextualRecallMetric` evaluates whether every statement in the `expected_output` is covered by the `retrieval_context`. If the same expected statement spans two overlapping chunks, the metric:

1. Scores both chunks as "covering" the statement — correct.
2. But if the LLM judge returns "yes" for chunk A and "no" for chunk B (because chunk B is partially redundant), the recall score is penalised even though the information is present.

**Redundancy ≠ Missing coverage.**

## Minimal Reproducer

```python
from deepeval.metrics import ContextualRecallMetric
from deepeval.test_case import LLMTestCase

# 10-K with 20% overlap across section boundary
chunk_a = "Revenue for FY2023 was $4.2B, up 12% YoY, driven by cloud growth."
chunk_b = "Driven by cloud growth, revenue for FY2023 reached $4.2B (12% increase)."

test_case = LLMTestCase(
    input="What was revenue for FY2023?",
    actual_output="Revenue was $4.2 billion in FY2023, up 12% year-over-year.",
    expected_output="Revenue for FY2023 was $4.2 billion, up 12% year-over-year.",
    retrieval_context=[chunk_a, chunk_b],
)
metric = ContextualRecallMetric(threshold=0.8)
metric.measure(test_case)
print(metric.score)  # Expected: ~1.0  Actual: ~0.5 (penalises chunk_b redundancy)
```

## Root Cause

`ContextualRecallMetric` does not have a source-grouping step equivalent to `_group_retrieval_contexts()` added to `ContextualPrecisionMetric` in PR #2743. Each chunk is scored independently, so overlapping content gets double-counted on the "wrong" side of the verdict.

## Proposed Fix

Apply the same `RetrievedContextData.source`-based grouping to `ContextualRecallMetric`:
1. Accept `RetrievedContextData` objects in `retrieval_context` (already partially supported).
2. Before scoring, merge chunks sharing the same `source` string into a single evaluation unit.
[...]

PR #2806, opened 2026-06-26
fix: correct dataset field defaults, add source-grouping to ContextualRecall, short-circuit Summarization identity

medhwu:
## Summary

Three bug fixes across the codebase:

### 1. Fix `EvaluationDataset` invalid field defaults and Python patterns
**File:** `deepeval/dataset/dataset.py`

- Changed `field(default_factory=[])` → `field(default_factory=list)` (×4). Passing a list instance to `default_factory` would raise `TypeError` if the explicit `__init__` is ever removed.
- Changed `except ValueError or json.JSONDecodeError:` → `except (ValueError, json.JSONDecodeError):` (×2). The `or` expression evaluates to `ValueError` alone, silently dropping the intent to also catch `JSONDecodeError`.
- Changed `goldens=[]` → `goldens=None` with `None` guard in `__init__`. Standard mutable-default antipattern fix.

### 2. Add source-grouping to `ContextualRecallMetric`
**File:** `deepeval/metrics/contextual_recall/contextual_recall.py`
**Closes:** #2788

- Added `_group_retrieval_contexts()` method (mirroring the existing fix in `ContextualPrecisionMetric` from PR #2743) that merges overlapping chunks by `RetrievedContextData.source` before scoring.
- Applied grouping in both `measure()` and `a_measure()`.
- This prevents recall scores from being unfairly penalized when RAG pipelines use sliding-window chunking with overlapping content.

### 3. Short-circuit `SummarizationMetric` when input equals output
**File:** `deepeval/metrics/summarization/summarization.py`

- Added identity check at the top of both `measure()` and `a_measure()`: when `test_case.input.strip() == test_case.actual_output.strip()`, returns score `1.0` immediately without invoking any LLM calls.
- This prevents the LLM-based claim/truth extraction from paraphrasing identical text differently and flagging them as contradictions, which previously produced scores in the 0.7–0.95 range instead of the expected 1.0.
- All instance attributes (truths, claims, verdicts, score_breakdown, verbose_logs) are still populated for downstream compatibility.

## Verification

- All three files pass `python -m py_compile` (syntax check).
- All three files pass `ruff check` with zero errors.
- All three files pass `black` formatting check.

### DIFF
--- a/deepeval/dataset/dataset.py
+++ b/deepeval/dataset/dataset.py
@@ -46,7 +46,6 @@ from deepeval.test_case import (
     ConversationalTestCase,
     ToolCall,
 )
-from deepeval.test_run.hyperparameters import process_hyperparameters
 from deepeval.test_run.test_run import TEMP_FILE_PATH
 from deepeval.utils import (
     convert_keys_to_snake_case,
@@ -57,7 +56,6 @@ from deepeval.test_run import (
     global_test_run_manager,
 )
 
-from deepeval.tracing import trace_manager
 from deepeval.tracing.tracing import EVAL_DUMMY_SPAN_NAME
 
 if TYPE_CHECKING:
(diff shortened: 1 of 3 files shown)


penguine-ip:
Hi @medhwu, thanks for taking the time to put this together!

After reviewing, we've decided not to move forward with this PR in its current form, for a few reasons:

- The branch has conflicts with `main` — some of the removed imports (e.g. `process_hyperparameters`) are now actively used in `evals_iterator`, so the `dataset.py` changes no longer apply cleanly.
- The `ContextualRecall` grouping only applies to `RetrievedContextData` objects sharing a `source`, so it wouldn't change the behavior of the plain-string reproducer in #2788 as-is.
- The `SummarizationMetric` short-circuit changes metric semantics (whether a verbatim copy of the input should score a perfect 1.0 is a design question we'd want to discuss separately), and we'd want regression tests accompanying any behavior change there.

If you'd like to pursue any of these individually, we'd suggest opening smaller, focused PRs rebased on the latest `main`, ideally with tests. Closing this one for now — thanks again for your interest in DeepEval!

PR #2811, opened 2026-06-26
feat(metrics): add CitationFaithfulnessMetric

rsrijith:
## What

Adds `CitationFaithfulnessMetric`, a citation-attribution faithfulness metric. It checks whether each cited passage supports the **specific claim it is attached to**, distinct from `FaithfulnessMetric`, which checks whether a claim is supported by the retrieval context *somewhere*.

This catches **citation misattribution**: a claim cited to passage `[A]` that does not support it, even though some other passage `[B]` in the context would. A content-only faithfulness check passes such answers; an attribution-aware one does not.

## How

- `deepeval/metrics/citation_faithfulness/` — `CitationFaithfulnessMetric(BaseMetric)` with `measure`/`a_measure`/`is_successful`, plus `schema.py` and `template.py` (the attribution-aware prompt).
- Registered in `deepeval/metrics/__init__.py`.
- Docs page `docs/.../metrics-citation-faithfulness.mdx` + nav entry.
- `tests/test_metrics/test_citation_faithfulness_metric.py` — keyless, uses a fake `DeepEvalBaseLLM`; proves the metric fails the misattribution case and passes the correctly-cited case. `pytest` passing (3/3).

Mirrors the structure of existing metrics (e.g. the BERTScore metric, #261).

## Usage

```python
from deepeval.metrics import CitationFaithfulnessMetric

metric = CitationFaithfulnessMetric(threshold=0.8)
test_case = LLMTestCase(input=..., actual_output='... [1]. ... [2].', retrieval_context=[...])
metric.measure(test_case)
```

`actual_output` is expected to contain `[N]` citation markers indexing into `retrieval_context`.


rsrijith:
Note on the failing `lint` check: it reports 47 files "would be reformatted", but **none of them are from this PR** — they're pre-existing files (e.g. `deepeval/scorer/scorer.py`, `deepeval/cli/inspect.py`, the `inspect/widgets/*` and `dag/serialization/*` modules) that the latest `psf/black@stable` (26.x) reformats. This looks like a repo-wide black version drift that also affects `main`, independent of this change.

The files added here are black-clean under the same version, and the `test` jobs all pass. Happy to either leave the repo-wide reformat to a separate maintenance PR, or include it here if you'd prefer — let me know. (The `Vercel` check is the deploy preview that needs maintainer authorization for fork PRs.)

penguine-ip:
@rsrijith thanks!

Issue #2817, opened 2026-06-29
`ArgumentCorrectnessMetric` always fails to render: 'stringified_tools_called' is undefined


safina57:
**Describe the bug**
`ArgumentCorrectnessMetric` is unusable in 4.0.7: every call raises a template render error before any LLM request is made. The `generate_verdicts` template references `{{ stringified_tools_called }}`, but the metric only passes `tools_called` to `_get_prompt`, and nothing in the codebase builds `stringified_tools_called`. Since templates render with strict-undefined, the render fails for every input. There is also no `evaluation_template` parameter on the metric, so there is no supported user-side workaround.

**To Reproduce**
1. `pip install deepeval==4.0.7`
2. Run the following (fails offline the error is raised at template render, before the API call, so no valid API key is needed):

```python
import asyncio
from deepeval.metrics import ArgumentCorrectnessMetric
from deepeval.test_case import LLMTestCase, ToolCall

metric = ArgumentCorrectnessMetric(model="gpt-4.1")
test_case = LLMTestCase(
  input="When did Trump first raise tariffs?",
  actual_output="Trump first raised tariffs in 2018.",
  tools_called=[ToolCall(name="WebSearch", input_parameters={"q": "Trump tariffs year"})],
)
asyncio.run(metric.a_measure(test_case))
```

3. See error:

```
deepeval.templates.resolver.MetricTemplateInterpolationError:
Missing variable during template render: 'stringified_tools_called' is undefined
```

**Expected behavior**
The metric renders its `generate_verdicts` prompt and returns a score

**Root cause**
- `deepeval/metrics/argument_correctness/argument_correctness.py`: `_a_generate_verdicts` / `_generate_verdicts` call `self._get_prompt("generate_verdicts", input=..., tools_called=tools_called, multimodal=...)`.
- The `generate_verdicts` template (`deepeval/metrics/argument_correctness/templates/generate_verdicts.txt` and the compiled `deepeval/templates/metrics/templates.json`) references `{{ stringified_tools_called }}`.
- `stringified_tools_called` is built nowhere (grep finds it only in the templates); `resolve_template` renders with strict-undefined, so the render raises.

**Suggested fix (either works)**
1. Template fix (minimal): change `generate_verdicts` to use `{{ tools_called }}`: it is already passed, and `ToolCall.__repr__` renders the formatted block the template's own example expects. Update both the `.txt` source and the compiled `templates.json`.
[...]

PR #2820, opened 2026-06-29
fix: align ArgumentCorrectness prompt with tool calls

DivyamTalwar:
## Summary

`ArgumentCorrectnessMetric` evaluates tool call arguments, but its
`generate_verdicts` prompt still instructed the model to return verdicts for
`statements`. This is a separate prompt-wording fix; it does not address the
historical undefined-variable render failure reported in #2817. The PR updates
that instruction to refer to tool calls and adds a regression test for the
compiled template text. It also hardens the template compiler/workflow so this
CI guard does not read symlinked template sources or persist checkout
credentials; these are bundled because the new regression coverage runs through
the same template sync guardrail.

## Changes

- Update the ArgumentCorrectness verdict-count instruction from `statements` to
  tool calls.
- Regenerate the Python and TypeScript metric template bundles with
  `python3 scripts/compile_metric_templates.py`.
- Add a template regression test that builds the template bundle and checks the
  ArgumentCorrectness instruction wording without installing the full project or
  rendering PR-controlled Jinja template text in CI.
- Reject symlinked metric template source files before the compiler reads them
  and symlinked generated output paths before the compiler writes them, with
  regression coverage using isolated temporary repo layouts.
- Add explicit read-only workflow permissions and disable persisted checkout
  credentials in the dedicated metric-template workflow.

## Why this matters

The metric is supposed to judge tool argument correctness per tool call. Keeping
the prompt aligned with that contract avoids confusing judge-model instructions
and makes future template regressions easier to catch without API keys or model
calls.

## Tests

Commands run:

- `python3 -m pytest tests/test_templates/test_metric_templates.py -q` — passed
  (9 tests)
- `python3 scripts/compile_metric_templates.py` — passed and updated both
  committed template bundles
- `python3 -m py_compile scripts/compile_metric_templates.py tests/test_templates/test_metric_templates.py`
  — passed
- `git diff --check` — passed

Not run:

- `python3 -m black --check scripts/compile_metric_templates.py tests/test_templates/test_metric_templates.py`
  — not run successfully because `black` is not installed in this local Python
  environment.

## Risk

Risk level: implementation low; score/baseline comparability moderate

Implementation risk: low. This is a small-surface prompt correction plus
regression coverage. It does not change metric scoring code, public APIs, or
template variables. It does make the template compiler reject symlinked template
sources/outputs; the current repository templates compile successfully after
that guard.

Score and baseline-comparability risk: moderate. The prompt wording can change
`ArgumentCorrectnessMetric` scores for some runs because the judge model now
sees a clearer per-tool-call verdict-count instruction. Teams that gate CI or
benchmark comparisons on this metric may want to compare scores before and after
the change on a representative eval set; pass/fail flips or material score drift
would be a good trigger for a release-note callout or broader validation before
release. Rollback is straightforward: revert the prompt text, compiled bundles,
compiler guard, workflow hardening, and tests.

## Issue

Related to #2817. This does not attempt to resolve the original
undefined-variable render failure described there. It fixes a separate confirmed
wording bug discovered in the same ArgumentCorrectness prompt area.

## Notes for maintainers

This intentionally does not rename the existing `stringified_tools_called`
template variable because the current code and compiled bundle already agree on
that variable. The PR only corrects the model-facing instruction.

CI note after opening:

- `Metric Templates / test` passed on this PR.
- `Core Tests`, `Metrics Tests`, `Confident Tests`, and integration jobs passed.
- `Python Black` is currently red because 47 existing files would be reformatted;
  the CI log reports the two PR-owned Python files as already well formatted.
- `TypeScript Lint` is currently red for existing Prettier drift in 66 TypeScript
  files.
- `TypeScript Tests` is currently red because the workflow runs secret-backed
  tests with empty `OPENAI_API_KEY` / `CONFIDENT_API_KEY`.
- `Vercel` is red due fork deployment authorization.


DivyamTalwar:

- **Issue:** The `ArgumentCorrectnessMetric` prompt/template described verdict counts in terms of `statements`, while the metric is evaluating tool calls. That wording can confuse evaluator output and future template maintenance.
- **PR intent:** Align the prompt with the actual tool-call evaluation flow and prevent template drift.
- **Fix:** Updates the generated metric templates and adds/strengthens template compile/sync regression coverage so Python and TypeScript template artifacts stay consistent.

DivyamTalwar:

**Housekeeping**
- Verified on a clean install (**deepeval 4.1.1, Python 3.12**) by loading the compiled template bundle (`deepeval/templates/metrics/templates.json`) the metric actually renders at runtime.

### ❌ Before — `main`
`ArgumentCorrectnessMetric` scores the arguments of **tool calls**, but the shipped judge prompt tells the model to count **statements**:

```
judge instruction -> Since you are going to generate a verdict for each statement, the number of 'verdicts' SHOULD BE STRICTLY EQUAL to the number of `statements`.
RESULT: MISMATCH — the judge is told to count 'statements' while the metric scores tool calls
```

Before: judge told to count statements

### ✅ After — this PR
The shipped prompt now matches the metric contract:

```
judge instruction -> Since you are going to generate a verdict for each tool call, the number of 'verdicts' SHOULD BE STRICTLY EQUAL to the number of tool calls.
RESULT: ALIGNED — the judge is told to count 'tool calls', matching the metric contract
```

After: judge told to count tool calls

Full evidence
</details>

> Regenerates the Python + TypeScript template bundles and adds regression coverage for the compiled text, plus hardens the template compiler against symlinked source/output paths. Non-goal: this does not rename the already-consistent `stringified_tools_called` variable and does not claim to resolve the separate undefined-variable report in #2817. Red `lint` / `Vercel` checks are the known repo-wide/external items.

PR #2823, opened 2026-06-30
ci: format typescript src and test with prettier

HarperZ9:
### Problem
The **TypeScript Lint** workflow (`.github/workflows/typescript_lint.yml`) runs:

```
npx prettier --check "src/**/*.ts" "test/**/*.ts"
```

On `main` this currently fails with `Code style issues found in 66 files`, which exits non-zero and (via fail-fast) cancels the rest of the TypeScript CI matrix.

### Change
Ran `prettier --write "src/**/*.ts" "test/**/*.ts"` using the pinned `prettier@3.6.2` from `typescript/package.json`. This is **formatting-only** (no behavioral changes) and is exactly what the failing step asks for ("Run Prettier with --write to fix.").

After this change, `npx prettier --check "src/**/*.ts" "test/**/*.ts"` reports `All matched files use Prettier code style!`.

The two pre-existing eslint **warnings** (unused `SLUG_PATTERN` and `confidentApiKey`) are intentionally left out of scope, since they are warnings and do not fail the build.


HarperZ9:
A quick CI note so the red status is easier to interpret:

- The TypeScript lint check passed on this PR, which is the check this formatting-only PR is meant to address.
- The failing `test` job is failing in credential-dependent integration suites, not on a formatting assertion. The job env shows empty credentials:
  - `OPENAI_API_KEY:`
  - `CONFIDENT_API_KEY:`
- The failures then come from those missing/empty keys, e.g. OpenAI returns `401 Incorrect API key provided: ''`, and Confident API setup throws `Please provide a valid Confident AI API Key.`
- Recent `main` runs for `TypeScript Tests` are also red, including the June 30 run on `11375f81`, so this does not appear to be introduced by the Prettier-only changes here.
- `Vercel` is still the fork-deploy authorization check.

I left the branch scoped to formatting because changing test credential behavior would be a separate CI/test policy change.

HarperZ9:
Regenerated against current main. The branch now formats exactly what the Ts Lint workflow checks (npx prettier --check "src/**/*.ts" "test/**/*.ts") and nothing else, which collapses the diff to 6 files. That workflow has failed on every push to main today; with this it passes. The earlier test-job failure on this PR was the fork-secrets gap (empty OPENAI_API_KEY), not a defect in the change.

PR #2828, opened 2026-07-01
optimize tracing hot paths and add OOM protection

theanuragg:
Eliminates four categories of waste in the observe() hot path identified during tracing profiling, and replaces an O(n) active-span scan with an O(1) counter.

## Changes

**Trace lifecycle — O(n) span scan → O(1) counter**  
Add a `traceSpanCounters` map to `TraceManager` that tracks the number of active spans per trace. `addSpan` and `removeSpan` manage the counter, replacing the `Array.from(activeSpans.values()).filter(...)` scan in `Tracer.exit()` with a constant-time counter lookup. This eliminates the dominant bottleneck for deeply nested traces with many concurrent spans.

**Tracer.trace() — inline enter(), remove dead span 1**  
`trace()` previously created two spans per invocation: span 1 was passed to `withTracingContext()` as an ALS context token but never registered in the trace tree, and span 2 was created by the subsequently called `enter()`. Inlining `enter()` directly into `trace()` eliminates the duplicate `createSpanInstance()` call and the redundant object allocation, saving one span object + one `children: []` array per traced call.

**Tracer — remove dead `new Date()` field initializers**  
`startTime` and `endTime` were unconditionally overwritten by `enter()` and `exit()`. Removed the initializer allocations, saving two `Date` objects per observe invocation.

**Tracer.trace() — remove double-await dead code**  
After `this.result = await callback()`, the result is already unwrapped. The subsequent `instanceof Promise ? await ...` branch is never reached and has been removed.

**createSpanInstance() — direct construction per type**  
Replaced the intermediate `spanKwargs` object + spread operator pattern with direct object literals for each span type (AGENT, LLM, RETRIEVER, TOOL, CUSTOM). Avoids the intermediate allocation and property-copy overhead for type-specific spans.

**Fix `no-unsafe-finally` lint error**  
Inverted the error guard in `trace()`'s `finally` block from `if (err) { log; return; }` to `if (err) { log; } else { ... }` to comply with ESLint's `no-unsafe-finally` rule.

## Performance

| Benchmark | Before | After | Δ |
|---|---|---|---|
| observe() wrapper (no-op fn) | 335,538 ops/sec | ~379,000 ops/sec | **+13%** |
| observe() nested 2-deep | 214,018 ops/sec | ~255,000 ops/sec | **+19%** |
| observe LLM + token tracking | 332,130 ops/sec | ~375,000 ops/sec | **+13%** |
| full trace lifecycle (span→cleanup) | 597,849 ops/sec | ~623,000 ops/sec | **+4%** |

## Trade-off

`addSpan + removeSpan` standalone throughput drops ~25% (2.0M → 1.5M) due to three extra Map operations for the span counter. This is dwarfed by the elimination of the O(n) active-span scan it enables, and the counter adds no overhead to the common observe path.

### DIFF
--- a/typescript/src/tracing/tracing.ts
+++ b/typescript/src/tracing/tracing.ts
@@ -345,6 +345,7 @@ export interface TraceManagerConfig {
   mask?: MaskFunction;
   confidentApiKey?: string;
   tracingEnabled?: boolean;
+  maxTraces?: number;
 }
 
 export class TraceManager {
@@ -369,6 +370,7 @@ export class TraceManager {
   private traceCaptureSink?: (trace: Trace) => void;
   private confidentApiKey: string = "";
   private tracingEnabled: boolean = true;
+  private maxTraces: number = 0; // 0 = unlimited
 
   /** Register/clear a sink that receives each completed trace (used by `evalsIterator`). */
   public setTraceCaptureSink(sink?: (trace: Trace) => void): void {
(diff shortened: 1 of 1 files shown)


theanuragg:
Only src/tracing/tracing.ts contains actual code changes. The other modified files in the diff are solely Prettier formatting fixes (pre-existing style issues).
 
Flamegraph:


Screenshot 2026-07-02 at 11 47 57 AM

@jeffreyip @kritinv @trevor-cai @A-Vamshi

penguine-ip:
@theanuragg hey thanks for the PR! unfortunately it is not reviewable - please undo all formatting to only highlight the changed code, thanks.

theanuragg:
> @theanuragg hey thanks for the PR! unfortunately it is not reviewable - please undo all formatting to only highlight the changed code, thanks.

I’ve reverted all unrelated formatting changes Please let me know if it looks better now

PR #2833, opened 2026-07-02
feat(typescript): add assertTest and TestRunManager for unit test workflows

sahniaditya007:
## Summary

This PR implements the first part of the TypeScript unit test workflow requested in #2734 by introducing the core building blocks:

* `assertTest()`
* `TestRunManager`

The implementation is intentionally framework-agnostic and does not depend on Jest, Vitest, or the CLI. These integrations can build on top of these APIs in follow-up PRs.

### New files

**`typescript/src/evaluate/assert-test.ts`**

* Added `assertTest(testCase, metrics, options?)`
* Added `AssertTestError`
* Added `AssertTestOptions`

`assertTest()` executes the supplied metrics using the existing evaluation pipeline, records the result through `testRunManager`, and throws an `AssertTestError` containing the failed metrics whenever one or more assertions fail.

---

**`typescript/src/evaluate/test-run-manager.ts`**

* Added `TestRunManager`
* Added `testRunManager` singleton
* Added `TestRunSummary`

The manager provides an in-memory accumulator for test execution state with support for:

* `record()`
* `getResults()`
* `getSummary()`
* `reset()`

Its API is designed to serve as the foundation for future Jest/Vitest integrations.

---

**`typescript/test/test-core/assert-test.test.ts`**

Added comprehensive unit tests covering:

* passing metrics
* failing metrics
* async metrics
* multiple metrics
* error handling
* invalid inputs
* `ConversationalTestCase`
* `TestRunManager` lifecycle
* singleton behavior
* edge cases

## Modified files

* `src/evaluate/index.ts`

  * Exported:

    * `assertTest`
    * `AssertTestError`
    * `AssertTestOptions`
    * `testRunManager`
    * `TestRunManager`
    * `TestRunSummary`

* `src/index.ts`

  * Re-exported the new public APIs from the package root.

* `typescript/README.md`

  * Added API documentation
  * Added usage examples
  * Documented error behavior

## Design decisions

* Reused the existing evaluation pipeline instead of duplicating metric execution logic.
* Kept the implementation completely framework-agnostic.
* Introduced only additive public APIs with no breaking changes.
* Structured the implementation so future Jest/Vitest reporters and `deepeval test run` can be implemented on top of these primitives.

## Future work

Follow-up PRs can build on this foundation by adding:

* Jest integration
* Vitest integration
* `deepeval test run` CLI support
* Local persistence
* Additional testing workflow features

Closes the first milestone for the TypeScript unit test workflow discussed in #2734.


sahniaditya007:
Anything else is required by me to fully integrate this block of code? @penguine-ip @A-Vamshi or maybe I can work on something else?

typescript/src/confident/evaluate.ts
```diff
@@ -0,0 +1,272 @@
+import { Api, Endpoints, API_BASE_URL, HttpMethods } from "./api";
```
A-Vamshi: I believe this file is now replaced by `src/evaluate/evaluate.ts` with the new evaluate method. Are we importing or using this file anywhere for the `assertTest` or `TestRunManager` logic?

typescript/src/evaluate/assert-test.ts
```diff
@@ -0,0 +1,172 @@
...
+  options: AssertTestOptions = {},
+): Promise<void> {
+  if (!testCase) {
+    throw new Error("assertTest: 'testCase' must not be null or undefined.");
+  }
+  if (!Array.isArray(metrics) || metrics.length === 0) {
+    throw new Error(
+      "assertTest: 'metrics' must be a non-empty array of metrics.",
+    );
+  }
```
A-Vamshi: For this validation, I don't think we're validating if the metrics and test cases are compatible? For example, for ConversationalTestCases we only want conversational metrics to be valid and vice versa, just needs one more validation for this here. Just need to error early when incompatible types are given

typescript/src/metrics/answer-relevancy/answer-relevancy.ts
```diff
@@ -75,9 +75,14 @@ export class AnswerRelevancyMetric extends BaseMetric {
   }
```
A-Vamshi, line 75: For files here and below, I think the changes are just prettier formatting, could you please undo them so the diff is small and easier to review for us?

A-Vamshi (changes requested):
Hey @sahniaditya007, thanks for the PR! I  added some comments and requested for the prettier only files to be reverted back for ease of review, can you please do that? Thank you, overall the base structure looks fine, would be good for a merge after the clean up and addressing the comments, thanks!

typescript/src/confident/evaluate.ts
sahniaditya007: Thanks for pointing this out! I checked the repository and confirmed that src/confident/evaluate.ts isn't used by the assertTest or TestRunManager implementation.

I searched for any remaining imports/references and couldn't find any. The new implementation uses src/evaluate/evaluate.ts exclusively.

Since this file isn't part of the changes introduced by this PR, I've left it untouched for now to keep the scope focused. If you'd prefer, I'm happy to remove the unused file in a follow-up cleanup PR.

typescript/src/evaluate/assert-test.ts
sahniaditya007: I've added an early compatibility check in assertTest() before any metrics are executed. It now validates that:

LLMTestCase only accepts single-turn metrics.
ConversationalTestCase only accepts conversational metrics.

To avoid duplicating logic, I reused the existing metricMatchesCase() helper rather than introducing a separate validation path.

I've also added tests covering both valid and invalid combinations so incompatible metric/test case pairs now fail immediately with a clear error message.

typescript/src/metrics/answer-relevancy/answer-relevancy.ts
sahniaditya007: I've reverted the formatting-only changes so the diff is focused on the functional updates for this feature. The remaining changes are limited to the assertTest/TestRunManager implementation, exports, tests, and the small example update required for the new evaluation flow.

sahniaditya007:
Hey @A-Vamshi,

I've updated this PR to address the review comments and align it with the current upstream structure after the changes introduced in `d924f0f` / #2880.

## Changes made

### 1. Removed unused file
- Deleted `typescript/src/confident/evaluate.ts` since it is no longer referenced and has been superseded by the new evaluation pipeline.

### 2. Reverted formatting-only changes
- Reverted all Prettier/formatting-only edits in the metrics modules to keep the review focused on functional changes.

### 3. Added metric compatibility validation
- `assertTest()` now validates metric/test-case compatibility before execution and throws a clear error when conversational metrics are used with `LLMTestCase` (or vice versa).

### 4. Aligned with the new modular structure
- Updated the implementation to match the current `src/evaluate/assert-test/` layout, including:
  - `assert-test`
  - `collector`
  - `errors`
  - `test-run-utils` (JSONL persistence)
  - `trace-scope`

### 5. Added test coverage
- Added `test-run-store.test.ts` to verify JSONL persistence.
- Added `trace-scope.test.ts` to verify local trace capture behavior.
- Added tests covering metric/test-case compatibility validation.

The implementation has also been cleaned up so the remaining diff focuses only on the functional changes required for this feature.

Could you please take another look when you have a chance? Thanks!

PR #2840, opened 2026-07-03
fix: include metric compatibility metadata in cache reuse

DivyamTalwar:
## Summary

DeepEval's test-run cache previously compared only a fixed set of metric configuration fields before reusing cached metric results. Constructor arguments and free-form metric config fields that can change scoring behavior, such as `PatternMatchMetric.pattern`, `GEval.criteria`, or custom metric settings, were not consistently represented as safe cache-compatibility metadata and could produce stale cached results or stale serialized cache payloads.

This change stores safe constructor-parameter and free-form metric config compatibility metadata in cache metadata and requires them to match before reusing cached metric data. Inputs that cannot be represented safely or passively miss closed instead of being serialized or reused incorrectly.

## Changes

- Add an unreleased changelog note for the user-visible cache invalidation behavior.
- Add constructor-parameter and free-form metric config compatibility metadata to `MetricConfiguration` and cache comparison, with the cache-identity policy isolated in `deepeval/test_run/cache_identity.py`.
- Include built-in/custom metric constructor parameters and free-form config fields in cache compatibility when they can be recovered safely from metric instances.
- Allow native model-backed metrics to reuse cache on subsequent runs once constructor compatibility metadata is written, when safe passive model identity/settings match.
- Miss closed instead of deriving cache metadata for unsafe, sensitive, opaque, cyclic, unrecoverable, or oversized cache identity inputs.
- Add focused regression coverage for cache hit/miss behavior, native and opaque model identity, free-form config redaction, collection/type identity, Pydantic schema identity, DAG structure, secret-bearing path values, custom container fail-closed behavior, and signed/encoded/relative URL handling, current-format save/load, and legacy cache payloads.

## Why this matters

This is a cache-correctness fix, not a new caching feature. It prevents stale metric results from being reused when constructor-level or free-form config scoring inputs change, while preserving reuse on subsequent runs for model-backed metrics when their safe passive native-model identity/settings match.

## Migration / compatibility note

Existing `.deepeval-cache.json` files remain loadable. After this change, cache hits also require constructor and free-form config compatibility metadata, so constructor- or config-sensitive metrics can miss once against pre-upgrade cache entries and regenerate when rerun.

Constructor and free-form config state that cannot be represented safely intentionally does not reuse cache. This includes secret-like values, credential-bearing URL/DSN-like values, opaque objects, callables, variadic or unrecoverable constructor params, non-finite values, unsafe Pydantic/DAG/embedding state, and too-large structures.

## User-visible behavior

- Existing local cache files still load safely.
- Constructor- or config-sensitive metrics can miss once against pre-upgrade cache entries and regenerate when rerun because scoring inputs are now part of cache compatibility.
- Metrics with unsafe or unrecoverable constructor/config state intentionally miss cache on every lookup instead of risking stale reuse.
- Native model-backed metrics keep cache reuse on subsequent runs once constructor compatibility metadata is written and their safe passive model identity/settings match.
- `.deepeval-cache.json` remains local runtime cache state and should not be shared as a portable artifact.

## Security / privacy note

This is conservative best-effort cache redaction, not a secrecy boundary. The cache identity treats obvious secret-like values, credential-bearing inputs, and unsafe runtime state as uncacheable so those cases miss closed instead of storing raw values or deterministic cache metadata.

Heuristics can false-positive on innocuous names and can miss unconventional secret formats. Recoverable non-secret primitive/config values and safe passive native-model identity/settings still participate through deterministic cache metadata, so `.deepeval-cache.json` should remain local runtime cache state rather than a shareable artifact.

## Tests

Commands run:

- `.venv/bin/python -m pytest tests/test_core/test_run/test_cache.py -q` — passed (101 passed)
- `.venv/bin/python -m pytest tests/test_core/test_run/test_cache.py tests/test_core/test_evaluation/test_execute/test_execute_llm_test_case.py -q` — passed (combined 101 passed, 4 skipped because `OPENAI_API_KEY` is not set, 2 warnings)
- `.venv/bin/python -m py_compile deepeval/test_run/cache.py deepeval/test_run/cache_identity.py tests/test_core/test_run/test_cache.py` — passed
- `.venv/bin/black --check deepeval/test_run/cache.py deepeval/test_run/cache_identity.py tests/test_core/test_run/test_cache.py` — passed
- `git diff --check` / `git diff --cached --check` — passed

## Risk

Risk level: medium

This intentionally reduces false cache hits and may cause additional recomputation for metrics whose constructor state was previously ignored or cannot be safely represented. Large or unsupported config/DAG shapes can intentionally miss cache and recompute rather than risk unsafe reuse. Rollback is straightforward by reverting the cache comparison/compatibility-metadata changes and the accompanying test file.

## Issue

No existing issue. Discovered during repository analysis.

## Notes for maintainers

Non-goals:

- This does not try to preserve cache reuse for opaque objects, callables, secrets, free-form schema metadata, unknown DAG nodes, opaque embeddings, or custom metrics whose constructor params cannot be recovered safely.
- This does not add a public metric-owned cache identity hook; that could be a future explicit extension point if maintainers want custom metrics to opt into broader cache reuse safely.

### DIFF
--- a/deepeval/test_run/cache.py
+++ b/deepeval/test_run/cache.py
@@ -10,12 +10,12 @@ from deepeval.utils import make_model_config
 
 from deepeval.test_case import SingleTurnParams, LLMTestCase, ToolCallParams
 from deepeval.test_run.api import MetricData
+from deepeval.test_run.cache_identity import CacheIdentity
 from deepeval.utils import (
     delete_file_if_exists,
     is_read_only_env,
     serialize,
 )
-from deepeval.metrics import BaseMetric
 from deepeval.constants import HIDDEN_DIR
 
 logger = logging.getLogger(__name__)
@@ -39,20 +39,24 @@ class MetricConfiguration(BaseModel):
     model_config = make_model_config(arbitrary_types_allowed=True)
 
     ##### Required fields #####
-    threshold: float
+    threshold: Union[float, str]
     evaluation_model: Optional[str] = None
-    strict_mode: bool = False
+    strict_mode: Union[bool, str] = False
     criteria: Optional[str] = None
-    include_reason: Optional[bool] = None
-    n: Optional[int] = None
+    include_reason: Optional[Union[bool, str]] = None
+    n: Optional[Union[int, str]] = None
 
     ##### Optional fields #####
-    evaluation_steps: Optional[List[str]] = None
-    assessment_questions: Optional[List[str]] = None
+    language: Optional[str] = None
+    evaluation_steps: Optional[Union[List[str], str]] = None
+    assessment_questions: Optional[Union[List[str], str]] = None
     embeddings: Optional[str] = None
     evaluation_params: Optional[
-        Union[List[SingleTurnParams], List[ToolCallParams]]
+        Union[List[SingleTurnParams], List[ToolCallParams], str]
     ] = None
+    # Stores constructor-parameter fingerprints/sentinels for cache compatibility.
+    # This field intentionally never contains raw constructor values.
+    custom_parameters: Optional[Dict[str, str]] = None
 
 
 class CachedMetricData(BaseModel):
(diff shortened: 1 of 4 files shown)


DivyamTalwar:
CI follow-up after the latest push (`83bdaf13fa4d26f6111441bf3032eb2209025fc6`):

- Core Tests now pass after the Pydantic `AliasChoices` compatibility fix.
- Metrics Tests, Confident Tests, and the integration matrix jobs shown in the check rollup are passing.
- `lint (ubuntu-latest)` is still failing because the repo-wide Black check reports 48 unrelated pre-existing files that would be reformatted. The PR-owned Python files are explicitly reported as already formatted in that log:
  - `deepeval/test_run/cache.py`
  - `deepeval/test_run/cache_identity.py`
  - `tests/test_core/test_run/test_cache.py`
- Vercel is failing with `Authorization required to deploy`, which appears to require project-side authorization rather than a code change in this PR.

I did not include unrelated repo-wide formatting changes in this focused cache-correctness PR.

DivyamTalwar:

- **Issue:** Cache reuse did not include enough metric compatibility metadata, so changed constructor/config state could still reuse stale metric results.
- **PR intent:** Make cache hits safer and more conservative when metric behavior-affecting state changes.
- **Fix:** Adds metric cache identity metadata/fail-closed handling and regression coverage for constructor/config changes, including Pydantic alias-helper cases discovered during CI.

DivyamTalwar:

**Housekeeping**
- Verified on a clean install (**deepeval 4.1.1, Python 3.12**), fully keyless: `PatternMatchMetric` (a string-matching metric, no LLM) + the real reuse decision `Cache.get_metric_data(metric, cached_test_case)`.

### ❌ Before — `main`
A cache entry computed for `PatternMatchMetric(pattern="expected")` is **reused** for a *different* metric `PatternMatchMetric(pattern="actual")` — the compatibility check ignores the `pattern` constructor arg:

```
Cache.get_metric_data(new, cached) -> REUSED the cached entry
RESULT: WRONG — a different pattern reused the cached score; the constructor param was ignored
```

Before: wrong cache reuse across a changed constructor param

### ✅ After — this PR
The changed constructor param yields a cache **miss**, so the score is recomputed:

```
Cache.get_metric_data(new, cached) -> None  (cache miss)
RESULT: CORRECT — different pattern -> cache miss -> the score is recomputed
```

After: cache miss on the changed param

Full evidence
</details>

> This is a cache-correctness fix: it prefers recomputation over silently returning a result produced by a materially different metric configuration (secrets redacted; opaque/cyclic/oversized values miss closed). Red `lint` / `Vercel` checks are the known repo-wide/external items; this PR's files pass Black + Ruff.

PR #2846, opened 2026-07-05
fix: run async arena compare when progress is hidden

DivyamTalwar:
## Summary

`compare()` currently skips async arena evaluation entirely when progress output is disabled with `DisplayConfig(show_indicator=False)`. The async path only scheduled `ArenaGEval` tasks inside the Rich progress-bar branch, so quiet CI/notebook/server usage returned `{}` without evaluating any test cases.

This PR moves async arena task scheduling into a shared helper so both visible and quiet progress modes execute the same measurement path. It also adds a deterministic no-network regression test that verifies quiet async compare invokes `ArenaGEval.a_measure`, returns the expected winner counts, populates test-run data, stops scheduling queued work after the first observed failure, and cancels unfinished tasks on failure.

## Changes

- Reuse the same async arena task scheduling path for `show_indicator=True` and `show_indicator=False`.
- Stop scheduling further async arena tasks after the first observed failure, then cancel and drain unfinished in-flight tasks so quiet mode does not leave background work running or launch avoidable follow-up work.
- Preserve existing throttling, concurrency, metric-copy creation, winner aggregation, and test-run map updates.
- Preserve the existing module-level `Contestant` import surface while marking it as an intentional compatibility re-export.
- Add `tests/test_core/test_evaluation/test_arena_compare.py` regression coverage using a dummy LLM and monkeypatched arena measurement, so the test does not require provider API keys.

## Why this matters

Disabling progress indicators should only suppress console UI. It should not change evaluation semantics or silently skip all comparisons. This matters for CI, notebooks, and log-sensitive environments where users commonly disable Rich progress output.

## Tests

Commands run:

- `python -m pytest tests/test_core/test_evaluation/test_arena_compare.py -q` — passed (`3 passed`, `1 warning` for existing event-loop deprecation in `deepeval.utils`).
- `python -m pytest --collect-only tests/test_core/test_evaluation/test_arena_compare.py tests/test_confident/test_compare.py -q` — passed (`5 tests collected`).
- `python -m black --check deepeval/evaluate/compare.py tests/test_core/test_evaluation/test_arena_compare.py` — passed.
- `python -m ruff check deepeval/evaluate/compare.py tests/test_core/test_evaluation/test_arena_compare.py` — passed.
- `python -m py_compile deepeval/evaluate/compare.py tests/test_core/test_evaluation/test_arena_compare.py` — passed.
- `git diff --check` — passed.

Not run:

- Full repository test suite — out of scope for this narrow compare-path fix.
- Live provider-backed arena comparison — regression test uses a deterministic dummy LLM to avoid API-key/network dependence.

## Risk

Risk level: low to medium.

The production change is limited to async arena task scheduling in `deepeval/evaluate/compare.py`. The visible progress path continues to use Rich progress, throttling, and concurrency as before; the quiet path now calls the same scheduling helper without progress objects. The helper now observes failures during throttled scheduling, stops launching queued cases after the first observed failure, cancels and drains unfinished in-flight tasks, and then re-raises the original error. Rollback is simple: revert this PR.

## Issue

No existing issue found. This was discovered during repository analysis.

Related context: open PR #2711 also touches `compare()` for return-details support, but does not address quiet async execution being skipped.

## Notes for maintainers

This PR intentionally does not change result-detail APIs, sync compare behavior, or provider/model logic. It only separates display suppression from async execution.

### DIFF
--- a/deepeval/evaluate/compare.py
+++ b/deepeval/evaluate/compare.py
@@ -13,7 +13,7 @@ import json
 
 from deepeval.errors import MissingTestCaseParamsError
 from deepeval.evaluate.configs import AsyncConfig, DisplayConfig, ErrorConfig
-from deepeval.test_case import ArenaTestCase, Contestant
+from deepeval.test_case import ArenaTestCase, Contestant  # noqa: F401
 from deepeval.test_case.api import create_api_test_case
 from deepeval.metrics import ArenaGEval
 from deepeval.utils import (
@@ -192,7 +192,73 @@ async def a_execute_arena_test_cases(
             run_duration=run_duration,
         )
 
-    # Create tasks for all test cases
+    async def execute_test_cases(
+        progress: Optional[Progress] = None,
+        pbar_id: Optional[int] = None,
+    ):
+        tasks = []
+
+        def raise_first_task_error():
+            for task in tasks:
+                if not task.done():
+                    continue
+                if task.cancelled():
+                    raise asyncio.CancelledError()
+                exception = task.exception()
+                if exception is not None:
+                    raise exception
+
+        async def wait_for_throttle_or_failure():
+            if throttle_value <= 0:
+                await asyncio.sleep(0)
+                raise_first_task_error()
+                return
+
+            raise_first_task_error()
+            active_tasks = [task for task in tasks if not task.done()]
+            if not active_tasks:
+                await asyncio.sleep(throttle_value)
+                raise_first_task_error()
+                return
+
+            start_time = time.perf_counter()
+            await asyncio.wait(
+                active_tasks,
+                timeout=throttle_value,
+                return_when=asyncio.FIRST_EXCEPTION,
+            )
+            raise_first_task_error()
+
+            remaining_throttle = throttle_value - (
+                time.perf_counter() - start_time
+            )
+            if remaining_throttle > 0 and all(
+                task.done() for task in active_tasks
+            ):
+                await asyncio.sleep(remaining_throttle)
+                raise_first_task_error()
+
+        try:
+            for i, test_case in enumerate(test_cases):
+                task = execute_with_semaphore(
+                    func=evaluate_single_test_case,
+                    test_case=test_case,
+                    progress=progress,
+                    pbar_id=pbar_id,
+                    index=i,
+                )
+                tasks.append(asyncio.create_task(task))
+                await wait_for_throttle_or_failure()
+
+            await asyncio.gather(*tasks)
+        except BaseException:
+            for task in tasks:
+                if not task.done():
+                    task.cancel()
+            if tasks:
+                await asyncio.gather(*tasks, return_exceptions=True)
+            raise
+
     if show_indicator:
         progress = Progress(
             TextColumn("{task.description}"),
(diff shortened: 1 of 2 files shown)


DivyamTalwar:
CI follow-up after opening this PR:

- All Python test/integration jobs passed: `test` (Confident Tests), `test` (Metrics Tests), `test` (Core Tests), plus `langchain`, `crewai`, `pydantic-ai`, `llama-index`, `openai-agents`, `openai`, `agentcore`, and `strands`.
- `lint (ubuntu-latest)` failed on existing repo-wide Black drift: the job reports 48 unrelated files would be reformatted. The two files touched by this PR are explicitly reported as already well formatted in that lint log:
  - `deepeval/evaluate/compare.py`
  - `tests/test_core/test_evaluation/test_arena_compare.py`
- `Vercel` failed with `Authorization required to deploy`, which appears to be an external deployment authorization condition.

Local verification for this PR’s touched files remains green:

- `python -m pytest tests/test_core/test_evaluation/test_arena_compare.py -q` — passed (`3 passed`, `1 warning` for an existing `asyncio.get_event_loop()` deprecation warning from `deepeval.utils`).
- `python -m pytest --collect-only tests/test_core/test_evaluation/test_arena_compare.py tests/test_confident/test_compare.py -q` — passed (`5 tests collected`).
- `python -m black --check deepeval/evaluate/compare.py tests/test_core/test_evaluation/test_arena_compare.py` — passed.
- `python -m ruff check deepeval/evaluate/compare.py tests/test_core/test_evaluation/test_arena_compare.py` — passed.
- `python -m py_compile deepeval/evaluate/compare.py tests/test_core/test_evaluation/test_arena_compare.py` — passed.
- `git diff --check` — passed.

DivyamTalwar:

- **Issue:** Async arena comparison could skip execution entirely when progress display was hidden (`show_indicator=False`).
- **PR intent:** Make hiding progress a display-only choice, not a behavior change.
- **Fix:** Ensures quiet async arena comparisons still schedule/run metric measurements and updates targeted tests for the hidden-progress path.

DivyamTalwar:

**Housekeeping**
- Verified on a clean install (**deepeval 4.1.1, Python 3.12**) with a fully keyless repro: the arena judge's `a_measure` is monkeypatched to a deterministic winner (no LLM, no API key), mirroring this PR's own network-free regression test.

### ❌ Before — `main`
`compare(..., async_config=AsyncConfig(run_async=True), display_config=DisplayConfig(show_indicator=False))` returns an **empty result** and the judge is **never called** — hiding the progress indicator silently skips the whole evaluation:

```
compare(show_indicator=False, run_async=True) returned: {}
arena judge a_measure was called: 0 time(s)
```

Before: compare returns empty when indicator hidden

### ✅ After — this PR
The identical call actually runs the comparison and returns real winner counts:

```
compare(show_indicator=False, run_async=True) returned: {'Version 2': 1}
arena judge a_measure was called: 1 time(s)
```

After: compare runs and returns winner counts

Full evidence
</details>

> This is a silent-wrong-result fix: turning off a progress indicator should only suppress console output, never skip the evaluation. Red `lint` / `Vercel` checks are the known repo-wide/external items; this PR's files pass Black + Ruff.

PR #2759, opened 2026-06-15
fix(synthesizer): correct parameter name in validate_chunk_size error message

Goutham-Annem:
## What
Fixed a wrong parameter name in the `validate_chunk_size()` error message suggestion inside `context_generator.py`.

## Why
When chunking fails due to insufficient chunks, the error message suggested adjusting `` `min_contexts_per_document` `` — a parameter that doesn't exist in the public API. The actual parameter is `` `min_contexts_per_source_file` ``. Users following the error message to fix their `ContextGenerator` call would find no such parameter and be stuck.

**Before:**
```
1. Adjust the `min_contexts_per_document` to no more than 3.
```

**After:**
```
1. Adjust the `min_contexts_per_source_file` to no more than 3.
```

## Changes
- `deepeval/synthesizer/chunking/context_generator.py` — line 933: renamed `min_contexts_per_document` → `min_contexts_per_source_file` in the error suggestion string

Closes #2467

PR #2849, opened 2026-07-05
fix(synthesizer): avoid UnboundLocalError when a document yields 0 chunks

hassaanch23:
## Summary

`ContextGenerator.validate_chunk_size` is meant to raise an **informative `ValueError`** when a source document produces fewer chunks than the requested number of contexts. Instead, when a document yields **0 chunks** (e.g. an empty / whitespace-only / unparseable file), it raises a confusing `UnboundLocalError`, hiding the real cause from the user.

## Reproduction

```python
from deepeval.synthesizer.chunking.context_generator import ContextGenerator

class FakeCollection:          # stand-in Chroma collection
    def count(self): return 0  # document produced 0 chunks

class StubEmbedder:
    def embed_text(self, x): return [0.0]
    async def a_embed_text(self, x): return [0.0]
    def embed_texts(self, xs): return [[0.0] for _ in xs]
    async def a_embed_texts(self, xs): return [[0.0] for _ in xs]

gen = ContextGenerator(document_paths=["doc.txt"], embedder=StubEmbedder(), model=<any model>)
gen.validate_chunk_size(min_contexts_per_source_file=1, collection=FakeCollection())
```

On `main` this raises:

```
UnboundLocalError: cannot access local variable 'adjust_chunk_size' where it is not associated with a value
```

instead of the intended `ValueError` describing the 0-chunk problem.

## Root cause

In `validate_chunk_size`, `adjust_chunk_size` and `adjust_overlap` are only assigned inside the `num_chunks > 0` and `min_contexts_per_source_file > 1 and self.chunk_overlap > 0` branches, but the final combined-suggestion check evaluates them unconditionally:

```python
if adjust_chunk_size or adjust_overlap:   # both unbound when num_chunks == 0
```

When a document yields 0 chunks, none of those branches run, so the names are never bound and Python raises `UnboundLocalError`.

## Fix

- Initialize `adjust_chunk_size = False` and `adjust_overlap = False` before the suggestion branches, so the combined check is always safe.
- Only emit the `"You have the following options:"` header when there is at least one concrete suggestion, so the 0-chunk message no longer dangles with an empty options list.

After the fix, a 0-chunk document produces a clean, actionable error:

```
Impossible to generate 1 contexts from a document with 0 chunks.
```

and the existing multi-suggestion messages are unchanged for the normal shortfall cases.

## Tests

Adds four regression tests in `tests/test_core/test_synthesizer/test_context_generator.py` (no API key required — a stub LLM is injected so `ContextGenerator` constructs offline):

- `test_validate_chunk_size_zero_chunks_raises_value_error` — 0 chunks, `min_contexts=1`
- `test_validate_chunk_size_zero_chunks_with_overlap_raises_value_error` — 0 chunks, `min_contexts>1`, non-zero overlap
- `test_validate_chunk_size_insufficient_chunks_still_lists_options` — normal shortfall path keeps the full suggestion list
- `test_validate_chunk_size_passes_when_enough_chunks` — no error when there are enough chunks

The two zero-chunk tests fail on `main` with `UnboundLocalError` and pass with this change.

## Notes

- Scope is intentionally minimal and does not touch the `min_contexts_per_document` wording that #2759 corrects (that is a separate line), so there is no conflict.
- Formatted with `black`.

### DIFF
--- a/deepeval/synthesizer/chunking/context_generator.py
+++ b/deepeval/synthesizer/chunking/context_generator.py
@@ -923,9 +923,15 @@ class ContextGenerator:
             # Build the error message with suggestions.
             error_lines = [
                 f"Impossible to generate {min_contexts_per_source_file} contexts from a document with {num_chunks} chunks.",
-                "You have the following options:",
             ]
             suggestion_num = 1
+            # Initialize both flags so the combined-suggestion check below is
+            # always safe. When a document produces 0 chunks none of the
+            # suggestion branches run, and without these defaults the final
+            # `adjust_chunk_size or adjust_overlap` check raised an
+            # UnboundLocalError instead of the intended ValueError.
+            adjust_chunk_size = False
+            adjust_overlap = False
 
             # 1. Suggest adjusting the number of contexts if applicable.
             if num_chunks > 0:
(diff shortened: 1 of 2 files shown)


penguine-ip:
@hassaanch23 thank you!

Issue #2859, opened 2026-07-07
`TaskCompletionMetric`: fails with `MetricTemplateInterpolationError: Missing variable during template render: 'tools_called_formatted' is undefined` exception


arsenalzp:
**Describe the bug**
`TaskCompletionMetric.measure()` method fail with the following exception:
```
MetricTemplateInterpolationError: Missing variable during template render: 'tools_called_formatted' is undefined
```

**To Reproduce**
```
try:
                task_metric = TaskCompletionMetric(
                    threshold=0.5,
                    model=self._model_name,
                    include_reason=True
                )
                
                current_trace_uuid = str(uuid.uuid4())
                trace_manager.start_new_trace()
                
                try:
                  if hasattr(trace_manager, 'current_trace') and trace_manager.current_trace:
                    current_trace_uuid = trace_manager.current_trace.uuid
                  elif hasattr(trace_manager, 'active_trace') and trace_manager.active_trace:
                    current_trace_uuid = trace_manager.active_trace.uuid
                  else:
                    current_trace_uuid = getattr(trace_manager.get_trace(), 'uuid')
                except Exception: 
                  current_trace_uuid = str(uuid.uuid4())
                  trace_manager.start_new_trace(
                    trace_uuid=current_trace_uuid
                  )
                
                steps = record.trace.get("steps", [])
                for step in steps:
                  node_name = step.get("node", "unknown-node")
                  description = step.get("description", "")
                  step_output = step.get("output", "Step completed")
                  if node_name == "tools" or "toolNames" in step:
                    current_span_type = SpanType.TOOL
                  elif node_name in ["plan", "answer"]:
                    current_span_type = SpanType.LLM
                    step_output = step.get("output")
                  else:
                    current_span_type = SpanType.AGENT
                  
                  span_obj = AgentSpan(
                    uuid=str(uuid.uuid4()),
                    trace_uuid=current_trace_uuid,
                    status=TraceSpanStatus.SUCCESS,
                    start_time=time.time(),
                    name=f"Node:{node_name}",
                    span_type=current_span_type,
                    input=description,
                    output=step_output
                  )
                  trace_manager.add_span_to_trace(span=span_obj)
                
                trace_manager.end_trace(trace_uuid=current_trace_uuid)
                task_metric.measure(test_case)
                score.task_completion = safe_float(task_metric.score)

                if getattr(task_metric, "reason", None):
                    score.notes.append(f"task_completion_reason={task_metric.reason}")

[...]

PR #2808, opened 2026-06-26
fix(task_completion): pass tools_called_formatted to the extract_goal_and_outcome template

ppcvote:
## Summary

`TaskCompletionMetric` raised `MetricTemplateInterpolationError` whenever the non-trace path ran with a real test case (the common case). The `extract_goal_and_outcome` template renders `{{ tools_called_formatted }}`, but both call sites passed `tools_called=test_case.tools_called` — the wrong kwarg, and an un-stringified `List[ToolCall]`. Because `_get_prompt` runs under `jinja2.StrictUndefined`, the template variable was never supplied and the render hard-failed, so the tool list never reached the prompt. This is the error reported in #2807.

## Fix

At both call sites (`_a_extract_task_and_outcome` and `_extract_task_and_outcome`, non-trace branch) pass:

```python
tools_called_formatted=print_tools_called(test_case.tools_called)
```

`print_tools_called` (from `deepeval.metrics.utils`) is the existing helper already used by `tool_correctness` and `goal_accuracy` for exactly this — it JSON-serializes the `ToolCall` list and returns `""` for `None`/empty, so the no-tools case renders cleanly. No template or `templates.json` change is needed (the template already expects `tools_called_formatted`), so the bundle stays in sync. The trace branch is untouched (it uses a different template).

## Tests

`tests/test_metrics/test_task_completion_tools_called.py` (new, API-free — the model call is monkeypatched, so it needs no key):

- `test_sync_extract_includes_tools_called` / `test_async_extract_includes_tools_called` drive the real prompt-construction path and assert the tool names reach the rendered prompt. The tool names (`quantum_router`, `cryo_vault`) are deliberately absent from the template's own example text, so these assertions can only pass if the user's `tools_called` actually reaches the render — they stay non-vacuous even against a future "renders to empty string" regression. Both **fail on the old `tools_called=` kwarg** (with the exact `MetricTemplateInterpolationError`) and pass on the fix.
- `test_buggy_kwarg_raises_under_strict_undefined` documents the root cause: rendering the template without `tools_called_formatted` raises under `StrictUndefined`.
- `test_extract_goal_and_outcome_template_renders_tool_names` is a template-contract check (tests 1 & 2 are the call-site fix proof).

Verified: `pytest tests/test_metrics/test_task_completion_tools_called.py` → 4 passed; reverting just the source fix makes the two call-site tests fail. `black --check` and `ruff check` clean.

Fixes #2807

PR #2862, opened 2026-07-07
fix: #2859

dvorobej:
Fixes `TaskCompletionMetric` crashing in the no-trace fallback path with:

`MetricTemplateInterpolationError: Missing variable during template render: 'tools_called_formatted' is undefined`

The Python implementation was passing raw `tools_called` into the `extract_goal_and_outcome` template, while the template expects `tools_called_formatted`. This PR formats tool calls with `print_tools_called()` before rendering, matching the existing TypeScript implementation.

## Changes

- Import and use `print_tools_called` in `TaskCompletionMetric`.
- Pass `tools_called_formatted` in both sync and async no-trace extraction paths.
- Add regression tests covering sync and async fallback behavior with `ToolCall` objects.
- Ensure the tests do not require an API key or real LLM call.

## Test Plan

```bash
poetry run pytest -vv tests/test_templates/test_metric_templates.py \
  tests/test_metrics/test_task_completion_metric_tools_called_template.py

### DIFF
--- a/deepeval/metrics/task_completion/task_completion.py
+++ b/deepeval/metrics/task_completion/task_completion.py
@@ -7,6 +7,7 @@ from deepeval.metrics.utils import (
     initialize_model,
     a_generate_with_schema_and_extract,
     generate_with_schema_and_extract,
+    print_tools_called,
 )
 from deepeval.test_case import (
     LLMTestCase,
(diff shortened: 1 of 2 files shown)


penguine-ip:
Thanks for the contribution. This PR duplicates the fix that was merged in #2808, so I'm closing it to keep the queue tidy.

PR #2873, opened 2026-07-10
fix(crewai): import integration cleanly when crewai is missing (was NameError; dead guard)

hassaanch23:
## Summary

The crewai integration is written to **degrade gracefully** when `crewai` isn't installed: `handler.py` keeps a `crewai_installed` flag and defines `is_crewai_installed()` (called first in `CrewAIEventsListener.__init__`), mirroring how the llama_index integration fails with a clean message. But that path is broken in several ways, so `import deepeval.integrations.crewai` crashes at import time when crewai is absent — and one of the guards can never fire at all.

## Reproduction (without `crewai` installed)

```python
import deepeval.integrations.crewai
```
```
NameError: name 'BaseEventListener' is not defined
```

## Root causes

1. **handler.py** — `class CrewAIEventsListener(BaseEventListener)` references a base class that is undefined in the `except ImportError` branch → `NameError` at import.
2. **subs.py** — the module-scope factory calls `create_deepeval_class(Crew, ...)`, `create_deepeval_class(Agent, ...)`, `create_deepeval_llm(LLM)` reference `Crew`/`Agent`/`LLM`, which are undefined when the import failed → `NameError` at import.
3. **subs.py** — name collision: the flag `is_crewai_installed = True/False` and the function `def is_crewai_installed():` share a name. The function ends up shadowing the flag, so its check `if not is_crewai_installed:` tests the **function object** (always truthy) and the `ImportError` is **never raised**. The guard is dead.
4. **tool.py** — `from crewai.tools import tool as crewai_tool` is a hard top-level import → `ModuleNotFoundError` at import.

## Fix

- **handler.py**: add `BaseEventListener = object` fallback in the except branch.
- **subs.py**: rename the flag to `crewai_installed` (fixing the shadow), and add `Crew = Agent = object` / `LLM = None` fallbacks so the module-scope factories don't `NameError`.
- **tool.py**: guard the `crewai_tool` import and raise the helpful `ImportError` at the top of `tool()`.

After this, `import deepeval.integrations.crewai` succeeds without crewai, and every public entry point — `Crew()`, `Agent()`, `LLM()`, `tool()`, `instrument_crewai()`, and the now-working `is_crewai_installed()` — raises `ImportError: CrewAI is not installed...`.

## Tests

Adds `tests/test_core/test_crewai_import_guard.py`, which simulates crewai being absent via an import blocker (so it runs regardless of whether crewai is installed) and asserts the module imports and each entry point raises the helpful `ImportError`. Fails on `main` (`NameError`), passes with this change. Formatted with `black`.

## Note

This is the crewai counterpart to the LangChain fix in #2872 (same "graceful degradation is broken at import" theme). The two PRs touch disjoint files and can be reviewed/merged independently.

### DIFF
--- a/deepeval/integrations/crewai/handler.py
+++ b/deepeval/integrations/crewai/handler.py
@@ -51,6 +51,11 @@ except ImportError as e:
             )
 
     crewai_installed = False
+    # Fallback base class so `class CrewAIEventsListener(BaseEventListener)`
+    # below can still be defined when crewai is absent. Without it, importing
+    # this module raises a cryptic NameError before is_crewai_installed() can
+    # surface the helpful error on use.
+    BaseEventListener = object
 
 # GLOBAL STATE to prevent duplicate listeners
 IS_WRAPPED_ALL = False
--- a/deepeval/integrations/crewai/subs.py
+++ b/deepeval/integrations/crewai/subs.py
@@ -6,13 +6,23 @@ from deepeval.metrics.base_metric import BaseMetric
 try:
     from crewai import Crew, Agent, LLM
 
-    is_crewai_installed = True
+    crewai_installed = True
 except ImportError:
-    is_crewai_installed = False
+    crewai_installed = False
+    # Fallbacks so the factory calls at the bottom of this module do not raise a
+    # NameError at import time when crewai is absent. `Crew`/`Agent` are used as
+    # base classes (must be classes); `LLM` is only invoked lazily inside the
+    # factory wrapper, after the is_crewai_installed() guard has run.
+    Crew = Agent = object
+    LLM = None
 
 
 def is_crewai_installed():
-    if not is_crewai_installed:
+    # NOTE: this must check the `crewai_installed` flag, not a name that shadows
+    # this function. Previously the flag was also named `is_crewai_installed`,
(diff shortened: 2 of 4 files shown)


hassaanch23:
Rebased onto current `main` — CI on this was red for two reasons, neither of them this change.

`Py Lint` was failing on `main`'s own Black debt (since fixed there), and `Py Core Tests` was failing on `test_update_current_span.py::TestUpdateCurrentSpan::test_update_tools`, an unrelated pre-existing failure. Both are resolved on current `main`, so the rebase should clear the run.

While I was here I re-verified the guard test actually catches the bug rather than passing for the wrong reason — reverting the three source files to `main`:

```
E   NameError: name 'BaseEventListener' is not defined
handler.py:90: NameError
1 failed
```

and passing with them in place. The blocker targets `crewai`, which is right here: the integration imports `crewai.tools` and `crewai.events`, both under that top-level package.

(Worth noting for contrast — the sibling LangChain PR #2872 had a real defect in exactly this spot. It blocked `langchain`, but that integration imports `langchain_core`, so the dependency stayed importable and the test asserted against the installed path. Fixed there separately.)

`black --check` clean under the pinned 25.12.0.

PR #2875, opened 2026-07-11
fix(dataset): serialize ToolCall type enum as JSON in save_as

otiscuilei:
### Who hits this

Anyone calling `EvaluationDataset.save_as("json" | "jsonl" | "csv", ...)` on a dataset where any `Golden` (or, for multi-turn, any `Turn`) has a non-empty `tools_called` or `expected_tools`. Since tool calls are central to evaluating agent/tool-use, and exporting a dataset to disk is a primary workflow, this makes dataset export a hard crash (no file written) for every tool-use dataset across all three formats.

### Symptom

```
TypeError: Object of type ToolCallType is not JSON serializable
```

### Root cause

In `save_as` (and in `utils.format_turns` for multi-turn), each `ToolCall` is serialized with `t.model_dump(by_alias=True, exclude_none=True)`. That uses Pydantic v2's default `mode="python"`, which leaves the field `type: ToolCallType = ToolCallType.FUNCTION` as a bare `Enum` member. `ToolCallType` is a plain `Enum` (not `str, Enum`), and `type` is never `None`, so `exclude_none` does not drop it. The resulting dict is handed to `json.dump` / `json.dumps`, which cannot serialize a bare Enum.

### Fix

Add `mode="json"` to the `ToolCall.model_dump(...)` at all four serialization sites (single-turn json/jsonl/csv in `deepeval/dataset/dataset.py`; multi-turn `format_turns` in `deepeval/dataset/utils.py`). Pydantic then emits the enum's string value, e.g. `{"name": "x", "type": "FUNCTION"}`. This is the localized, low-risk fix; it has no effect on tool-call-free data.

### Before / after

```python
from deepeval.test_case import ToolCall
import tempfile

d = EvaluationDataset()
d.add_golden(Golden(input="hi", tools_called=[ToolCall(name="search")]))
d.save_as("json", tempfile.mkdtemp(), file_name="x")

# Before: TypeError: Object of type ToolCallType is not JSON serializable
# After:  writes x.json with "tools_called": [{"name": "search", "type": "FUNCTION"}]
```

### Tests

Adds `TestSaveAndLoad::test_save_as_serializes_tool_call_type_as_json_string`, which saves a tool-call golden to json, jsonl and csv, asserts no exception, asserts the on-disk `type` is the string `"FUNCTION"` (no bare-enum repr), and round-trips through each loader. The three single-turn export paths are covered directly; the multi-turn `format_turns` path receives the same one-line fix for parity.

### DIFF
--- a/deepeval/dataset/dataset.py
+++ b/deepeval/dataset/dataset.py
@@ -1269,7 +1269,9 @@ class EvaluationDataset:
                                 if hasattr(t, "model_dump"):
                                     dumped.append(
                                         t.model_dump(
-                                            by_alias=True, exclude_none=True
+                                            mode="json",
+                                            by_alias=True,
+                                            exclude_none=True,
                                         )
                                     )
                                 elif hasattr(t, "dict"):
@@ -1395,7 +1397,9 @@ class EvaluationDataset:
                                 if hasattr(t, "model_dump"):
                                     dumped.append(
                                         t.model_dump(
-                                            by_alias=True, exclude_none=True
+                                            mode="json",
+                                            by_alias=True,
+                                            exclude_none=True,
                                         )
                                     )
                                 elif hasattr(t, "dict"):
@@ -1476,7 +1480,9 @@ class EvaluationDataset:
                                 if hasattr(t, "model_dump"):
                                     dumped.append(
                                         t.model_dump(
-                                            by_alias=True, exclude_none=True
+                                            mode="json",
+                                            by_alias=True,
+                                            exclude_none=True,
                                         )
                                     )
                                 elif hasattr(t, "dict"):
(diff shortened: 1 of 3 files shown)


dchou1618:
Would you be able to link this PR to tracking issue https://github.com/confident-ai/deepeval/issues/2922? This change fixes a concrete bug in dataset export serialization, and tying it to an issue would make the change easier to track and review. Thank you

otiscuilei:
Closing this one: the same fix has since landed on main via #2928 and #2926 (nice work @Solaris-star @nightcityblade), which also resolves #2922 that @dchou1618 linked. The only part of this PR still adding value was the single-turn regression test, so I re-cut just that against current main as #2961.

PR #2879, opened 2026-07-13
feat(benchmarks): add PubMedQA benchmark

hiajoo:
## What

Adds a `PubMedQA` benchmark backed by the 1,000 expert-labeled examples in the PQA-L subset.

The benchmark presents each biomedical question together with its labeled PubMed abstract sections and evaluates the model's `yes`, `no`, or `maybe` answer using exact-match accuracy.

## Changes

- Add `PubMedQA` with configurable `n_problems`, verbose output, and custom confinement instructions.
- Load the `pqa_labeled` configuration from `qiaojin/PubMedQA`.
- Preserve abstract section labels such as `BACKGROUND`, `METHODS`, and `RESULTS` when formatting prompts.
- Add a `PubMedQASchema` that restricts structured responses to `yes`, `no`, or `maybe`.
- Support both direct schema responses and native model `(response, cost)` tuples.
- Fall back to confined plain-text generation when a model does not support schemas.
- Normalize predictions to lowercase and strip surrounding whitespace before exact-match scoring.
- Export `PubMedQA` from `deepeval.benchmarks`.
- Add a benchmark documentation page and navigation entry.

## Usage

```python
from deepeval.benchmarks import PubMedQA

benchmark = PubMedQA(n_problems=10)
benchmark.evaluate(model=my_model)

print(benchmark.overall_score)
```

## Testing

- `pytest -q tests/test_benchmarks/test_pub_med_qa.py` → **5 passed**
- Unit coverage includes:
  - abstract-section prompt formatting
  - missing section labels
  - all three answer labels (`yes`, `no`, and `maybe`)
  - structured and plain-text generation
  - native `(response, cost)` tuples
  - end-to-end evaluation with an in-memory dataset
- Black formatting check passes.
- Loaded and converted all 1,000 PQA-L examples:
  - 552 `yes`
  - 338 `no`
  - 110 `maybe`
- Smoke-tested the live `deepseek-chat` API on one example from each gold-label category (`yes`, `no`, and `maybe`); all three requests completed successfully and produced schema-valid predictions.

The live API smoke test validates provider integration and structured response parsing only; it is not intended to measure model quality.

## References

- Dataset: https://huggingface.co/datasets/qiaojin/PubMedQA
- Paper: https://arxiv.org/abs/1909.06146

### DIFF
--- a/deepeval/benchmarks/__init__.py
+++ b/deepeval/benchmarks/__init__.py
@@ -15,6 +15,7 @@ from .lambada.lambada import LAMBADA
 from .winogrande.winogrande import Winogrande
 from .equity_med_qa.equity_med_qa import EquityMedQA
 from .ifeval.ifeval import IFEval
+from .pub_med_qa.pub_med_qa import PubMedQA
 
 __all__ = [
     "BigBenchHard",
@@ -34,4 +35,5 @@ __all__ = [
     "Winogrande",
     "EquityMedQA",
     "IFEval",
+    "PubMedQA",
 ]
--- a/deepeval/benchmarks/pub_med_qa/__init__.py
+++ b/deepeval/benchmarks/pub_med_qa/__init__.py
@@ -1,0 +1,3 @@
+from .pub_med_qa import PubMedQA
+
+__all__ = ["PubMedQA"]
--- a/deepeval/benchmarks/pub_med_qa/pub_med_qa.py
+++ b/deepeval/benchmarks/pub_med_qa/pub_med_qa.py
@@ -46,9 +46,13 @@ class PubMedQA(DeepEvalBaseBenchmark):
             correct_predictions = 0
             predictions_row = []
             goldens = self.load_benchmark_dataset()[: self.n_problems]
+            if not goldens:
+                raise ValueError(
+                    "PubMedQA dataset must contain at least one problem."
+                )
 
             for idx, golden in enumerate(
-                tqdm(goldens, desc=f"Processing {self.n_problems} problems")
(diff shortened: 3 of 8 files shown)


hiajoo:
Hi maintainers — I investigated the failing checks and confirmed that they are reproducible on the current `main` branch without this PR's changes:

  - **Core Tests:** `test_update_tools` fails on `main` because the
    existing tracing snapshot does not include the `type` field:
    `Extra keys at '...expectedTools[0]': {'type'}`.
  - **Lint:** running the CI version of Black (`26.5.1`) against `main`
    reports 49 pre-existing files that would be reformatted. None of the
    Python files added or modified by this PR are among them.
  - **Vercel:** deployment requires Confident AI team authorization for
    external fork PRs.

  The PubMedQA-specific tests pass locally (`5 passed`), and the integration, metrics, confident, and PR-title checks are green.

penguine-ip:
Thanks @hiajoo !

One scoring edge case: `overall_accuracy` divides by `self.n_problems`, but `goldens` can contain fewer items when a custom dataset is supplied. That silently underreports accuracy—for example, 3/3 correct with `n_problems=10` becomes 0.3. Could we divide by `len(goldens)` instead (and handle an empty dataset explicitly) and add a short regression test?

hiajoo:
Thanks for catching this! 

I’ve updated overall_accuracy to divide by the number of evaluated goldens, added an explicit error for an empty dataset, and added regression coverage for both cases. I also updated the progress display to reflect the actual number of evaluated problems. The PubMedQA tests now pass locally (6 passed).

PR #2882, opened 2026-07-13
fix(synthesizer): don't drop context-generation cost when synthesis_cost is 0

### LINKED ISSUE #2881: Synthesizer.synthesis_cost silently drops context-generation cost for native models
### Bug Description

`Synthesizer.synthesis_cost` silently drops the first cost accrual from context generation when using a native model, because the truthiness check treats the initial `0` value as "no cost yet."

### Where

`deepeval/synthesizer/synthesizer.py`, in `generate_goldens_from_docs`, `a_generate_goldens_from_docs`, `generate_conversational_goldens_from_docs`, and `a_generate_conversational_goldens_from_docs` (4 call sites — lines 489, 608, 2122, 2239 on current `main`):

```python
self.synthesis_cost = 0 if self.using_native_model else None
...
if self.synthesis_cost:
    self.synthesis_cost += context_generator.total_cost
```

`self.synthesis_cost` is reset to `0` (not `None`) whenever a native model is used. `0` is falsy in Python, so `if self.synthesis_cost:` evaluates to `False` on this very first accrual, and `context_generator.total_cost` — the real per-chunk LLM cost from context quality scoring — is never added.

The same file already has the correct pattern at 4 other cost-accrual sites (lines 1689, 1715, 1736, 1750): `if self.synthesis_cost is not None:`. This is a clear copy/refactor slip in the 4 sites above, not an intentional design choice — verified by grepping every `if self.(evaluation_cost|synthesis_cost|total_cost|cost):`-style check in the codebase; only these 4 lines use the buggy truthy form.

### Steps to Reproduce

Any call to `generate_goldens_from_docs`/`a_generate_goldens_from_docs` (or the conversational variants) with a native model (the default) under-reports `synthesis_cost` by exactly the context-generation cost every single time — the cost silently vanishes rather than being tracked.

### Expected Behavior

`synthesis_cost` should include the context-generation cost, matching the correct `is not None` pattern already used elsewhere in the same file.

### Actual Behavior

The cost is silently dropped every time, so any cost-tracking/budgeting built on `synthesizer.synthesis_cost` under-reports actual spend.

### Fix

Change the 4 truthy checks to `if self.synthesis_cost is not None:`. I have a fix + regression tests ready (both sync and async paths) and will open a PR referencing this issue.

okaditya84:
Closes #2881

## What

`Synthesizer.synthesis_cost` resets to `0` (not `None`) whenever a native model is used, but 4 cost-accrual sites checked `if self.synthesis_cost:` before adding the context generator's cost. `0` is falsy in Python, so the very first accrual was silently dropped every time `generate_goldens_from_docs`/`a_generate_goldens_from_docs`/`generate_conversational_goldens_from_docs`/`a_generate_conversational_goldens_from_docs` ran with a native model — real context-generation cost (LLM calls used to score candidate contexts) vanished from `synthesis_cost` unreported.

The same file already has the correct pattern at 4 other cost-accrual sites (`if self.synthesis_cost is not None:`), so this was a clear copy/refactor slip rather than an intentional check, isolated to these 4 sites.

## How

Change all 4 sites to `if self.synthesis_cost is not None:`, matching the existing correct pattern used elsewhere in the same file. No other behavior changes.

## Tests

Added two regression tests in `tests/test_core/test_synthesizer/test_synthesizer.py`:
- `test_generate_goldens_from_docs_accrues_context_generation_cost` (sync)
- `test_a_generate_goldens_from_docs_accrues_context_generation_cost` (async)

Both build a `Synthesizer` via `Synthesizer.__new__` (no API key needed, following the existing pattern used by the nearby deadlock-regression tests), mock out `ContextGenerator` and the golden-generation step (the only two things that would make real network calls), and assert the mocked context-generation cost survives into `synthesis_cost`.

I verified both tests **fail on `main`** with `assert 0 == 2.5` / `assert 0 == 3.5` (i.e. they reproduce the exact silent-drop) and **pass** with this fix.

Full local verification: `pytest tests/test_core/test_synthesizer/` — 21 passed, 21 skipped (pre-existing `@requires_openai_key`-gated integration tests, unaffected). `black --check` clean on the changed source file; the one docstring-wrapping tweak `black` requested in my new test code is included.

## Type of change

- [x] Bug fix

### DIFF
--- a/deepeval/synthesizer/synthesizer.py
+++ b/deepeval/synthesizer/synthesizer.py
@@ -486,7 +486,7 @@ class Synthesizer:
                         pbar_id=pbar_id,
                     )
                 )
-                if self.synthesis_cost:
+                if self.synthesis_cost is not None:
                     self.synthesis_cost += context_generator.total_cost
                 print_synthesizer_status(
                     SynthesizerStatus.SUCCESS,
@@ -605,7 +605,7 @@ class Synthesizer:
                     pbar_id=pbar_id,
                 )
             )
-            if self.synthesis_cost:
+            if self.synthesis_cost is not None:
                 self.synthesis_cost += context_generator.total_cost
             print_synthesizer_status(
                 SynthesizerStatus.SUCCESS,
@@ -2119,7 +2119,7 @@ class Synthesizer:
                         pbar_id=pbar_id,
                     )
                 )
-                if self.synthesis_cost:
+                if self.synthesis_cost is not None:
                     self.synthesis_cost += context_generator.total_cost
                 print_synthesizer_status(
                     SynthesizerStatus.SUCCESS,
@@ -2236,7 +2236,7 @@ class Synthesizer:
                     pbar_id=pbar_id,
                 )
             )
-            if self.synthesis_cost:
+            if self.synthesis_cost is not None:
                 self.synthesis_cost += context_generator.total_cost
             print_synthesizer_status(
                 SynthesizerStatus.SUCCESS,
--- a/tests/test_core/test_synthesizer/test_synthesizer.py
+++ b/tests/test_core/test_synthesizer/test_synthesizer.py
@@ -2,6 +2,7 @@ import asyncio
 import pytest
 import os
 
+from contextlib import contextmanager
 from typing import List
 
 from deepeval.synthesizer.synthesizer import Synthesizer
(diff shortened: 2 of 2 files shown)


Comui520:
Nice catch — `0` being falsy is an easy thing to miss. 

This is complementary to PR #2736, which guards against `None` cost from `calculate_cost()` (when a model's pricing isn't in the registry, e.g. `deepseek-v4-flash`). Both bugs hit the same cost-accrual sites in `synthesizer.py` — merging either one will likely cause a minor conflict for the other, but they coexist cleanly. Happy to rebase whichever lands second.

okaditya84:
Good catch, thanks for flagging! Confirmed #2736 touches the same `if self.synthesis_cost:`/`if self.synthesis_cost is not None:` sites but for a different root cause (`None` from an unregistered model's `calculate_cost()` vs. this PR's falsy-zero check) — no logical conflict, just adjacent lines. Happy to rebase on top of whichever merges first.

PR #2889, opened 2026-07-14
feat(evaluate): recalculate pass rates against new thresholds without re-running

### LINKED ISSUE #1996: Need to Recalculate Pass Rates with New Thresholds on Existing Evaluation Data Without Re-running LLM Evaluations
Description:

I have completed a full evaluation run of my LLM application using DeepEval, and the results are saved with a specific Test Run ID in Confident AI. The evaluation used GEval metrics with initially set thresholds.

After reviewing the results, I found that some thresholds (e.g., for "Faithfulness" metric) were set too high, causing unnecessarily low pass rates. I want to explore how different threshold values would affect the overall pass rates and metrics.

I need to efficiently recalculate all derived statistics (pass/fail status, overall pass rates, aggregate scores) based on existing evaluation data using new threshold values, without:
  1. Re-running any LLM evaluations (to avoid additional API costs and time)
  2. Creating duplicate test cases or test runs
  3. Losing the original evaluation scores and reasoning

sanidhyasin:
Closes #1996.

### Problem

After an `evaluate()` run, the only way to see how different thresholds would affect pass/fail is to re-run the whole evaluation — paying the LLM cost again — even though the per-metric scores are already computed and stored. There was no way to replay those scores against new thresholds.

### Change

Add `EvaluationResult.recalculate_results(thresholds, print_results=True)`:

- Recomputes each metric's pass/fail and each test case's overall pass/fail against new per-metric thresholds, reusing the scores already produced by `evaluate()`. **No LLM evaluations are re-run.**
- Non-destructive: returns a **new** `EvaluationResult`; the original result (including its scores and reasons) is left untouched, so nothing is duplicated or lost.
- Metrics whose name is not in `thresholds` keep their original threshold and status.
- Pass/fail semantics match evaluation exactly (errored or score-less metrics fail; otherwise pass when `score >= threshold`).
- When `print_results=True`, prints the recalculated overall pass rates via the existing `aggregate_metric_pass_rates`.

```python
results = evaluate(test_cases, metrics=[faithfulness, relevancy])

# Explore a lower Faithfulness bar without spending anything on the LLM again
what_if = results.recalculate_results({"Faithfulness": 0.5})
```

### Tests

`tests/test_core/test_evaluation/test_recalculate_results.py` covers lowering/raising thresholds (pass↔fail flips at both metric and test-case level), untouched originals, metrics absent from the map, errored metrics, and cases without metric data. All run fully offline.

### DIFF
--- a/deepeval/evaluate/types.py
+++ b/deepeval/evaluate/types.py
@@ -1,5 +1,5 @@
 from typing import Optional, List, Union, Dict
-from dataclasses import dataclass
+from dataclasses import dataclass, replace
 from pydantic import BaseModel
 
 from deepeval.test_run.api import MetricData, TurnApi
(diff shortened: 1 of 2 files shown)


sanidhyasin:
@penguine-ip this one has been open a couple of weeks, could you take a look when you get a chance?

PR #2898, opened 2026-07-16
fix: prevent trim_and_load_json from corrupting valid JSON strings

### LINKED ISSUE #2770: trim_and_load_json in models/llms/utils.py corrupts valid JSON string values (incomplete fix from #2701)
### Summary

`deepeval.models.llms.utils.trim_and_load_json` corrupts valid JSON whose **string values** contain `", ]"` or `", }"`. It unconditionally runs `re.sub(r",\s*([\]}])", r"\1", jsonStr)` on the whole string *before* parsing, so it cannot tell a structural trailing comma from a comma that legitimately appears inside a string value.

This is the **same bug** that PR #2701 (commit `2340aab`) fixed — but that fix only patched `deepeval/metrics/utils.py` and `deepeval/dataset/utils.py`. The third copy, `deepeval/models/llms/utils.py::trim_and_load_json`, was missed. This copy is imported by **every** concrete model backend (`openai_model`, `anthropic_model`, `azure_model`, `amazon_bedrock_model`, `gemini`/`local_model`, `deepseek_model`, `litellm_model`, `grok_model`, `kimi_model`, `gateway_model`) to parse generated output, so any metric whose JSON `reason`/`verdict` free-text contains `, }` or `, ]` is silently corrupted.

### Reproduction (current `main`)

```python
from deepeval.models.llms.utils import trim_and_load_json

assert trim_and_load_json('{"reason": "score is 1,} good"}') == {"reason": "score is 1,} good"}
# AssertionError: returns {'reason': 'score is 1} good'} — the comma is silently dropped
```

### Expected vs. actual

- **Expected:** valid JSON parses unchanged.
- **Actual:** the comma inside the string value is stripped, mangling the data.

### Environment

- deepeval `4.0.6` (current `main`)
- Python 3.x, macOS

### Suggested fix

Apply the same approach used by #2701 in the other two copies: attempt a direct `json.loads` first, and only fall back to the comma-stripping regex when the direct parse raises `JSONDecodeError`. I'll open a PR.

Mohak-Agrawal:
## What was the bug

`trim_and_load_json` in `deepeval/models/llms/utils.py` corrupts valid JSON whose **string values** contain `, ]` or `, }`. It unconditionally ran `re.sub(r",\s*([\]}])", r"\1", jsonStr)` on the whole string *before* parsing, so it can't tell a structural trailing comma from a comma that legitimately appears inside a string value.

This copy of the helper is imported by every concrete model backend (`openai_model`, `anthropic_model`, `azure_model`, `amazon_bedrock_model`, `local_model`, `deepseek_model`, `litellm_model`, `grok_model`, `kimi_model`, `gateway_model`), so any metric whose JSON `reason`/`verdict` free-text contains `, }` or `, ]` gets silently corrupted.

## Root cause

This is the same bug PR #2701 (commit `2340aab`) fixed — but that fix only patched the other two copies of this helper (`deepeval/metrics/utils.py` and `deepeval/dataset/utils.py`). This third copy, `deepeval/models/llms/utils.py::trim_and_load_json`, was missed.

## What the fix does

Attempts `json.loads(jsonStr)` directly first. Only on `JSONDecodeError` does it fall back to the trailing-comma-stripping regex and retry — matching the pattern already used in the two sibling copies. Valid JSON with commas inside string values is never touched; malformed JSON with a genuine trailing comma is still recovered.

## How to reproduce / verify

```python
from deepeval.models.llms.utils import trim_and_load_json

# Before this fix: silently corrupted to {'reason': 'score is 1} good'}
# After this fix: passes unchanged
assert trim_and_load_json('{"reason": "score is 1,} good"}') == {"reason": "score is 1,} good"}

# Trailing-comma recovery still works as a fallback
assert trim_and_load_json('{"a": 1, "b": 2,}') == {"a": 1, "b": 2}
```

Both assertions pass against the patched function.

Closes #2770

### DIFF
--- a/deepeval/models/llms/utils.py
+++ b/deepeval/models/llms/utils.py
@@ -17,12 +17,18 @@ def trim_and_load_json(
         input_string = input_string + "}"
         end = len(input_string)
     jsonStr = input_string[start:end] if start != -1 and end != 0 else ""
-    jsonStr = re.sub(r",\s*([\]}])", r"\1", jsonStr)
+
     try:
         return json.loads(jsonStr)
     except json.JSONDecodeError:
-        error_str = "Evaluation LLM outputted an invalid JSON. Please use a better evaluation model."
-        raise DeepEvalError(error_str)
+        # Some models emit a trailing comma before a closing ] or }. Strip it
+        # and retry, but only after a direct parse fails, so valid JSON string
+        # values containing ", ]" or ", }" are never corrupted.
+        try:
+            return json.loads(re.sub(r",\s*([\]}])", r"\1", jsonStr))
+        except json.JSONDecodeError:
+            error_str = "Evaluation LLM outputted an invalid JSON. Please use a better evaluation model."
+            raise DeepEvalError(error_str)
     except Exception as e:
         raise Exception(f"An unexpected error occurred: {str(e)}")
 


Mohak-Agrawal:

Issue #1989, opened 2025-08-29
Add `evaluation_template ` to GEval metric


MagicCube:
**Description**
DeepEval is a widely appreciated LLM evaluation framework, loved by many for its versatility and ease of use. Among its built-in metrics, GEval is one of the most commonly utilized due to its effectiveness in evaluating model performance. However, GEval currently lacks support for an `evaluation_template` parameter, which limits its flexibility in certain use cases.

**Proposed Feature**
Introduce an `evaluation_template` parameter to the GEval metric. This parameter would allow users to customize the evaluation process by specifying a template for the evaluation prompts or responses. This addition would enhance the adaptability of GEval, making it more suitable for diverse evaluation scenarios.

**Benefits**
1. **Improved Customization**: Users can tailor the evaluation process to better fit their specific requirements.
2. **Enhanced Flexibility**: The `evaluation_template` parameter would make GEval more versatile for different tasks and datasets.
3. **Broader Applicability**: By supporting templates, GEval could be used in more complex evaluation setups, such as multi-turn conversations or domain-specific tasks.

penguine-ip:
hy @MagicCube releasing a new version right now to support this :)

kumar-hardik:
@penguine-ip can we add this in ConversationalGEval as well?

A-Vamshi:
Hey @kumar-hardik, we added this to `ConversationalGEval` as well :)

PR #2908, opened 2026-07-16
feat(metrics): restore evaluation_template for custom prompt templates (#1757)

### LINKED ISSUE #1757: Option to overwrite GEval default templete.
Metrics like [FaithfulnessMetric](https://github.com/confident-ai/deepeval/blob/main/deepeval/metrics/faithfulness/faithfulness.py) or [AnswerRelevancyMetric](https://github.com/confident-ai/deepeval/blob/main/deepeval/metrics/answer_relevancy/answer_relevancy.py) have a field `evaluation_template`.  On the other hand, metrics like [GEval](https://github.com/confident-ai/deepeval/blob/main/deepeval/metrics/g_eval/g_eval.py) or [ConversationRelevancyMetric](https://github.com/confident-ai/deepeval/blob/main/deepeval/metrics/conversation_relevancy/conversation_relevancy.py) lacks such field.
Is there an option to add such field (with default value set on `GEvalTemplate` for GEval and so on) to those metrics which are missing it? It would be really helpful for evaluating other languages than english.

hassaanch23:
## Summary

deepeval's docs promise that **every metric supports an `evaluation_template`** for customizing its LLM-as-a-judge prompts:

> *"Simply provide a custom template class as the `evaluation_template` parameter to your metric of choice."* — `metrics-introduction.mdx`
> *"Yes. Every metric in `deepeval` supports an `evaluation_template` parameter."* — `faq.mdx`

The maintainers built and shipped this (see #1989). But the template refactor that moved prompts into `templates.json` (`ee5695319` + the resolver) **silently removed the `evaluation_template` parameter and the per-metric template classes from every metric**. So the documented API is now broken everywhere:

```python
from deepeval.metrics.answer_relevancy import AnswerRelevancyTemplate
# ImportError: cannot import name 'AnswerRelevancyTemplate'
```
```python
AnswerRelevancyMetric(evaluation_template=CustomTemplate)
# TypeError: __init__() got an unexpected keyword argument 'evaluation_template'
```

This is tracked by the open issue **#1757**. This PR restores the documented API on top of the current resolver architecture, with **zero behavior change** for existing users.

## How it works

The new architecture renders every prompt through a single choke point — `PromptMixin._get_prompt(method, **kwargs)` → `resolve_template(...)`. The restoration hooks in there:

```python
def _get_prompt(self, method, *, template_class=None, multimodal=False, strict=True, **kwargs):
    evaluation_template = getattr(self, "evaluation_template", None)
    if evaluation_template is not None:
        override = getattr(evaluation_template, method, None)
        if callable(override):
            return override(**kwargs)          # user's custom prompt string
    return resolve_template("metrics", template_class or self.__class__.__name__, method,
                            multimodal=multimodal, strict=strict, **kwargs)  # default
```

Design choices that keep this safe and faithful to the documented API:

- **Marker template classes.** `AnswerRelevancyTemplate` / `FaithfulnessTemplate` / `GEvalTemplate` are empty, subclassable `BasePromptTemplate`s. A user overrides only the methods they care about; **methods they don't override aren't on the class, so they fall straight through to the resolver** — which means `multimodal`/`strict`/`_fragments` rendering is fully preserved for defaults.
- **Zero behavior change.** Metrics that don't set `evaluation_template` (all but the three here) hit `getattr(..., None)` and take the exact same resolver path as before. A test asserts the default output is **byte-identical** to `resolve_template(...)`.
- **Matches the published contract exactly** — the same `class CustomTemplate(AnswerRelevancyTemplate)` + `evaluation_template=` usage the docs already show.

## What's included

- Core mechanism in `base_metric.py` + new `BasePromptTemplate`.
- `evaluation_template` parameter + exported template class for **`GEval`** (the metric maintainers explicitly added this to before — #1989/#1757) and the **full RAG metric suite**: `AnswerRelevancyMetric`, `FaithfulnessMetric`, `ContextualPrecisionMetric`, `ContextualRecallMetric`, and `ContextualRelevancyMetric`.
- Fixes an `evaluation_params` → `evaluation_template` typo in `metrics-introduction.mdx`.

## Example (now works again)

```python
from deepeval.metrics.answer_relevancy import AnswerRelevancyTemplate

class CustomTemplate(AnswerRelevancyTemplate):
    @staticmethod
    def generate_statements(actual_output):
        return f"""Break down the text into standalone statements.\n\nText:\n{actual_output}\n\nJSON:"""

metric = AnswerRelevancyMetric(evaluation_template=CustomTemplate)
```

## Tests

`tests/test_metrics/test_evaluation_template.py` — 10 offline tests (stub LLM, no network/API key). They assert, for all three metrics: the override is used, non-overridden methods fall back to the default, the default path is **byte-identical** to the resolver, and override kwargs are forwarded. Since every prompt a metric builds flows through `_get_prompt`, these cover the `measure()` path too. `black` clean; existing template-integrity tests still pass.

## Scope / follow-up

The mechanism is metric-agnostic and now covers GEval + the full RAG suite. Extending it to the remaining metrics whose docs advertise `evaluation_template` (bias, toxicity, PII leakage, etc.) is a one-line parameter + a marker template class each — happy to add the rest in this PR or a follow-up, whichever you prefer.

Closes #1757.

### DIFF
--- a/deepeval/metrics/answer_relevancy/__init__.py
+++ b/deepeval/metrics/answer_relevancy/__init__.py
@@ -1,3 +1,4 @@
 from .answer_relevancy import AnswerRelevancyMetric
+from .template import AnswerRelevancyTemplate
 
-__all__ = ["AnswerRelevancyMetric"]
+__all__ = ["AnswerRelevancyMetric", "AnswerRelevancyTemplate"]
--- a/deepeval/metrics/answer_relevancy/answer_relevancy.py
+++ b/deepeval/metrics/answer_relevancy/answer_relevancy.py
@@ -1,4 +1,4 @@
-from typing import Optional, List, Union
+from typing import Optional, List, Type, Union
 
 from deepeval.utils import (
     get_or_create_event_loop,
@@ -21,6 +21,8 @@ from deepeval.metrics.answer_relevancy.schema import (
     Verdicts,
     AnswerRelevancyScoreReason,
 )
+from deepeval.metrics.answer_relevancy.template import AnswerRelevancyTemplate
+from deepeval.metrics.prompt_template import BasePromptTemplate
 
 
 class AnswerRelevancyMetric(BaseMetric):
@@ -37,6 +39,7 @@ class AnswerRelevancyMetric(BaseMetric):
         async_mode: bool = True,
         strict_mode: bool = False,
         verbose_mode: bool = False,
+        evaluation_template: Type[BasePromptTemplate] = AnswerRelevancyTemplate,
     ):
         self.threshold = 1 if strict_mode else threshold
         self.model, self.using_native_model = initialize_model(model)
@@ -45,6 +48,7 @@ class AnswerRelevancyMetric(BaseMetric):
         self.async_mode = async_mode
         self.strict_mode = strict_mode
         self.verbose_mode = verbose_mode
+        self.evaluation_template = evaluation_template
 
     def measure(
         self,
--- a/deepeval/metrics/answer_relevancy/template.py
+++ b/deepeval/metrics/answer_relevancy/template.py
@@ -1,0 +1,20 @@
+from deepeval.metrics.prompt_template import BasePromptTemplate
+
+
+class AnswerRelevancyTemplate(BasePromptTemplate):
+    """Customizable prompt template for :class:`AnswerRelevancyMetric`.
+
+    Subclass this and override any of the static methods below to customize the
+    metric's prompts, then pass the subclass via
+    ``AnswerRelevancyMetric(evaluation_template=YourTemplate)``. Any method you do
+    not override falls back to deepeval's default template.
+
+    Overridable methods (each must return the final prompt string):
+
+    - ``generate_statements(actual_output)``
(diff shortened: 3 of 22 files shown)


hassaanch23:
Hi @penguine-ip — thanks for merging #2849 and #2916! Whenever you get a chance, #2908 restores the documented GEval + RAG evaluation_template with the full metric suite. #2872 and #2873 are small import-crash fixes. Happy to adjust anything if needed.

hassaanch23:
Closing this — `main` has since implemented the feature, and more completely than this PR did.

I went to rebase off the conflicts and found `evaluation_template` already on `main`. All six metrics this PR touched now accept it as a constructor parameter:

```
answer_relevancy       contextual_relevancy
contextual_precision   faithfulness
contextual_recall      g_eval
```

and so do 36 others — 42 metric classes in total, against the 6 here.

`BaseMetric._get_prompt` also handles a case this PR did not: an explicit `template_class` borrows another class's templates, so a metric's own `evaluation_template` is deliberately not allowed to hijack it. That's a real edge I'd missed.

Nothing left here worth your review time. Thanks for picking up #1757.

PR #2909, opened 2026-07-17
fix: RoleViolationMetric strict_mode always succeeded (threshold=0)

### LINKED ISSUE #2901: RoleViolationMetric strict_mode always succeeds due to threshold initialization
### Summary

`RoleViolationMetric` does not enforce strict mode correctly.

### Current Behavior

When `strict_mode=True`, the threshold is initialized as:

```python
self.threshold = 0 if strict_mode else threshold
```

The success check later evaluates:

```python
```

Since the minimum score is `0.0`, a failing role violation still satisfies:

```python
0.0 >= 0
```

making strict mode always succeed.

### Expected Behavior

Strict mode should fail whenever any role violation is detected.

### Reproduction

1. Create a `RoleViolationMetric` with `strict_mode=True`.
2. Evaluate a response containing a clear role violation.
3. Observe that `success=True`.

### Proposed Fix

```python
self.threshold = 1 if strict_mode else threshold
```

### Files

```
deepeval/metrics/role_violation/role_violation.py
```

RahilOp:
## Summary

`RoleViolationMetric.__init__` initializes the threshold as:

```python
self.threshold = 0 if strict_mode else threshold
```

`RoleViolationMetric`'s score is binary — `_calculate_score()` returns `0.0` when a role violation is detected, `1.0` otherwise. With `strict_mode=True`, the threshold becomes `0`, so a detected violation still satisfies `self.success = self.score >= self.threshold` (`0.0 >= 0` is `True`). Strict mode therefore never fails, even when a clear role violation is found.

Every other binary metric in the codebase (e.g. `g_eval`, `answer_relevancy`, `contextual_recall`, `summarization`) uses `1` as the strict-mode threshold; `role_violation` was the sole outlier.

## Fix

```python
self.threshold = 1 if strict_mode else threshold
```

## Testing

Added `tests/test_metrics/test_role_violation_metric_strict_mode.py` — uses `DummyModel` (no `OPENAI_API_KEY` needed, following the pattern in `test_answer_relevancy_metric_empty_output.py`) to test the threshold/success logic directly:
- `strict_mode=True` → `threshold == 1`
- `strict_mode=False` → threshold stays as configured
- a detected violation (`score=0.0`) under `strict_mode=True` → `is_successful() is False`
- no violation (`score=1.0`) under `strict_mode=True` → `is_successful() is True`

Verified these tests actually catch the bug: reverted the fix locally, confirmed 2 of the 4 new tests fail as expected, then restored the fix — all 4 pass. Also ran the full non-API-key-dependent metrics suite (103 tests) to check for regressions — all pass. `black`/`ruff` clean.

Fixes #2901

### DIFF
--- a/.scripts/release.py
+++ b/.scripts/release.py
@@ -77,7 +77,9 @@ class Target:
     json_key: str  # key in sdk-versions.json
     publish_commands: List[str]  # printed at the end, never run
     paths: List[str]  # what shipping this SDK covers, for "did it change?"
-    single_digit: bool  # cap x.y.z at 9 and carry left, rather than plain semver
+    single_digit: (
+        bool  # cap x.y.z at 9 and carry left, rather than plain semver
+    )
 
 
 TARGETS: Dict[str, Target] = {
--- a/deepeval/dataset/golden.py
+++ b/deepeval/dataset/golden.py
@@ -202,9 +202,7 @@ class Golden(BaseModel):
     output_token_count: Optional[int] = Field(
         default=None,
         serialization_alias="outputTokenCount",
-        validation_alias=AliasChoices(
-            "outputTokenCount", "output_token_count"
-        ),
+        validation_alias=AliasChoices("outputTokenCount", "output_token_count"),
     )
     source_file: Optional[str] = Field(
         default=None, serialization_alias="sourceFile"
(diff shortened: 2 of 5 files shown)


RahilOp:
Cleaned up this PR:

- Rebased onto latest `upstream/main` and dropped the merge commit, so the diff is now limited to the `RoleViolationMetric` fix and its test.
- Re-signed the commit with the SSH signing key attached to this account, which should resolve the unsigned-commit warning.

The unrelated CI failures on the previous run were from upstream `main`, not from this change.

RahilOp:
Hi maintainers, the **Vercel** check on this PR is failing because it's opened from a fork and requires a Confident AI team member to authorize the deploy preview. This change is limited to the  fix and its test, plus a few small formatting cleanups — no docs or deployment assets are affected. Could someone please approve the Vercel deploy when you have a moment? Happy to rebase if that helps re-trigger the check. Thanks!

PR #2910, opened 2026-07-17
fix(evaluate): don't suppress exceptions via return in finally block

### LINKED ISSUE #2903: Return inside finally block suppresses exceptions in async evaluation loop
### Summary

The async evaluation loop contains a `return` statement inside a `finally` block.

### Current Behavior

Python suppresses active exceptions whenever execution returns from `finally`.

### Evidence

Python emits

```
SyntaxWarning:
'return' in a finally block
```

### Expected Behavior

Exceptions raised during evaluation should propagate normally.

### Suggested Fix

Remove the `return` and only skip cleanup when the loop has already been closed.

### Files

```
deepeval/evaluate/execute/loop.py
```

sanidhyasin:
### Summary

Fixes #2903.

The async agentic evaluation loop in `deepeval/evaluate/execute/loop.py` has a `return` statement inside a `finally` block. A `return` in `finally` causes Python to discard any exception that is propagating out of the `try`, so real evaluation errors are silently swallowed instead of surfacing to the caller. Python 3.14 flags this at compile time:

```
SyntaxWarning: 'return' in a 'finally' block
```

### Root cause

Inside the task-cleanup `finally` of `evaluate_test_cases`:

```python
finally:
    # if it is already closed, we are done
    if loop.is_closed():
        return          # <- swallows any in-flight exception
    ...stray-task cleanup...
```

When the loop is already closed there is nothing to clean up, but returning from `finally` also suppresses whatever exception was being raised.

### Fix

Skip the cleanup with a guard instead of returning, so control falls through the `finally` normally and any active exception continues to propagate:

```python
finally:
    if not loop.is_closed():
        ...stray-task cleanup...
```

The cleanup logic is unchanged; it is only guarded rather than reached via an early `return`. When the loop is closed the behaviour is the same as before (nothing is cleaned up), minus the exception suppression.

### Tests

Adds `tests/test_core/test_evaluation/test_execute/test_finally_no_return.py`:

- `test_loop_has_no_return_in_finally` — walks the module AST and asserts no `return` lives inside any `finally` block.
- `test_loop_compiles_without_syntax_warning` — compiles the module with `SyntaxWarning` promoted to an error.

Both fail on the current code (at line 809) and pass with the fix. The existing `test_execute` suite continues to pass (`24 passed, 12 skipped` — skips require `OPENAI_API_KEY`).


penguine-ip:
Thanks for the PR! Could you replace the broad AST-based test with a behavioral regression test that verifies the original exception propagates when the event loop is closed?

sanidhyasin:
Thanks for the review! Done — I've replaced the AST-based check with a behavioral regression test (`test_closed_loop_error_propagation.py`).

It drives `a_execute_agentic_test_cases_from_loop` into its task-cleanup `finally` block with a scheduled task and a **closed** event loop, so the gather step raises `RuntimeError("Event loop is closed")`. The test asserts that exception propagates out of the generator. With the old `return`-in-`finally`, the exception was swallowed and the generator completed silently, so the test fails on the pre-fix code and passes with the fix.

sanidhyasin:
@penguine-ip the behavioral regression test you asked for is pushed, ready for another look whenever you have a moment.

PR #2916, opened 2026-07-18
fix(benchmarks): correct silent answer mis-scoring in MathQA, DROP, and BigBenchHard

### LINKED ISSUE #2917: Silent answer mis-scoring in MathQA, DROP, and BigBenchHard benchmarks (correct answers scored 0)
## Summary

Three benchmark answer-scoring paths silently score a **correct** prediction as `0` with no error raised, systematically deflating reported accuracy. Each is reproduced below with expected-vs-actual output.

(For rigor: I also checked `deepeval.utils.normalize_text` against the official SQuAD `normalize_answer` — it matches exactly, so that path is fine.)

Environment: current `main`.

---

### 1. MathQA — the answer schema cannot represent option `(e)`

MathQA (AQuA-RAT) questions **always have five options `a`–`e`** (option `e` is frequently *"none of these"*), but the schema used to constrain/extract the answer only allows four:

```python
# deepeval/benchmarks/schema.py
class MultipleChoiceSchemaLower(BaseModel):
    answer: Literal["a", "b", "c", "d"]   # missing "e"
```

```python
>>> from deepeval.benchmarks.schema import MultipleChoiceSchemaLower
>>> MultipleChoiceSchemaLower(answer="e")
ValidationError: 1 validation error for MultipleChoiceSchemaLower
answer: Input should be 'a', 'b', 'c' or 'd'
```

A schema-constrained (native) model can therefore **never emit `e`**, and the confinement instruction for non-native models likewise says *"Output 'a', 'b', 'c', or 'd'."* Result: **every MathQA item whose gold answer is `e` (~1/5 of the dataset) is scored 0**, even for a perfect model.

### 2. DROP — comma delimiter shatters answers that contain commas

Each golden's answer spans are packed into `expected_output` with a delimiter and split back at scoring time; the delimiter is `","`:

```python
# deepeval/benchmarks/drop/drop.py
DELIMITER = ","
```

```python
>>> from deepeval.benchmarks.drop.template import DROPTemplate
>>> from deepeval.scorer.scorer import Scorer
>>> packed = DROPTemplate.parse_list_to_str(["1,000"], ",")   # dataset build
>>> DROPTemplate.parse_str_to_list(packed, ",")                # scoring time
['1', '000']
>>> Scorer.quasi_contains_score(['1', '000'], "1,000")         # perfect prediction
0     # expected 1
```

Any DROP answer containing a comma — every number ≥ 1,000 written with a thousands separator, spans such as `"New York, New York"` — can **never be matched**.

### 3. BigBenchHard — the batch path corrupts schema answers (diverges from the non-batch path)

`batch_predict` applies free-text extraction to **every** prediction, including schema-constrained outputs:

```python
# deepeval/benchmarks/big_bench_hard/big_bench_hard.py (batch_predict)
prediction = prediction.split()[-1]
prediction = prediction[:-1] if self.enable_cot else prediction
```

With `enable_cot=True` (the default), a schema answer `"(A)"` becomes `"(A"` → scored 0. The non-batch `predict` uses `str(res.answer)` directly, so the two paths disagree on the same input:

```
schema (native) model, enable_cot=True, gold "(A)", model returns "(A)":
  predict(...)        -> prediction "(A)",  score 1
  batch_predict(...)  -> prediction "(A",   score 0   # silently wrong
```

(It also crashes on numeric schema answers, an `int`, at `.split()`.)

---

## Proposed fixes

1. Add `"e"` to `MultipleChoiceSchemaLower` and both MathQA confinement strings.
2. Use a control-char delimiter (`\x1f`) for DROP's internal span packing — it cannot occur inside a DROP answer; multi-span behavior unchanged.
3. Apply BBH's last-token / trailing-punctuation trimming only to free-text generations, matching the non-batch path.

All three are covered by offline regression tests (no model/network/dataset). I've opened a PR with the fixes and tests.

hassaanch23:
Closes #2917.

## Summary

While auditing the benchmark answer-scoring paths, I found **three silent bugs** — cases where a **correct** prediction is scored `0` with no error raised, systematically deflating reported benchmark accuracy. Each is reproduced below and covered by an offline regression test that fails on the current code.

`normalize_text` (the SQuAD-style normalizer behind exact-match) was checked against the official SQuAD `normalize_answer` and matches it exactly, so it is *not* changed here.

---

### 1. MathQA — the answer schema can't represent option `(e)`  *(severe)*

MathQA (AQuA-RAT) questions **always have five options `a`–`e`** (option `e` is frequently *"none of these"*), but the schema used to constrain/extract the answer only allowed four:

```python
# deepeval/benchmarks/schema.py
class MultipleChoiceSchemaLower(BaseModel):
    answer: Literal["a", "b", "c", "d"]   # <- no "e"
```

```python
>>> from deepeval.benchmarks.schema import MultipleChoiceSchemaLower
>>> MultipleChoiceSchemaLower(answer="e")
pydantic_core._pydantic_core.ValidationError: 1 validation error for MultipleChoiceSchemaLower
answer: Input should be 'a', 'b', 'c' or 'd'
```

A schema-constrained (native) model can therefore **never emit `e`**, and the confinement instruction for non-native models also said *"Output 'a', 'b', 'c', or 'd'."* So **every MathQA item whose gold answer is `e` (~1/5 of the dataset) is scored 0**, even for a perfect model.

**Fix:** add `"e"` to `MultipleChoiceSchemaLower` and to both confinement strings. (`MultipleChoiceSchemaLower` is used only by MathQA.)

### 2. DROP — comma-delimited span packing corrupts answers containing commas  *(severe)*

Each golden's answer spans are packed into `expected_output` with a delimiter and split back apart at scoring time. The delimiter was `","`:

```python
# deepeval/benchmarks/drop/drop.py
DELIMITER = ","
```

So a single-span answer that itself contains a comma is shattered:

```python
>>> from deepeval.benchmarks.drop.template import DROPTemplate
>>> from deepeval.scorer.scorer import Scorer
>>> packed = DROPTemplate.parse_list_to_str(["1,000"], ",")     # dataset build
>>> DROPTemplate.parse_str_to_list(packed, ",")                  # scoring time
['1', '000']
>>> Scorer.quasi_contains_score(['1', '000'], "1,000")          # perfect prediction
0     # expected 1
```

Any DROP answer with a comma — every number ≥ 1,000 written with a thousands separator, spans like `"New York, New York"` — can **never be matched**.

**Fix:** use the unit-separator control char (`\x1f`) as the internal delimiter; it cannot occur inside a DROP answer. Multi-span behavior is unchanged.

### 3. BigBenchHard — the batch path corrupts schema-constrained answers (inconsistent with the non-batch path)  *(moderate)*

`batch_predict` applies free-text extraction to **every** prediction, including clean schema outputs:

```python
# deepeval/benchmarks/big_bench_hard/big_bench_hard.py (batch_predict)
prediction = prediction.split()[-1]
prediction = prediction[:-1] if self.enable_cot else prediction
```

With `enable_cot=True` (the default), a schema answer `"(A)"` becomes `"(A"` → scored 0. The non-batch `predict` doesn't do this — it uses `str(res.answer)` directly — so the two paths **disagree** on the same input:

```python
# schema (native) model, enable_cot=True, gold "(A)", model correctly returns "(A)"
predict(...)        -> prediction "(A)",  score 1
batch_predict(...)  -> prediction "(A",   score 0     # silently wrong
```

**Fix:** apply the last-token / trailing-punctuation trimming only to free-text generations, matching the non-batch path (this also fixes a latent crash where numeric schema answers, an `int`, hit `.split()`).

---

## Tests

New offline regression tests under `tests/test_benchmarks/` (no model, network, dataset download, or API key). They fail on the current code and pass with this change:
- MathQA schema accepts all of `a`–`e`.
- DROP single-span answers with commas survive the round-trip and score 1; multi-span still splits.
- BBH `batch_predict` scores a schema answer correctly with and without CoT.

`black` clean. Happy to split into per-benchmark PRs if you'd prefer.

### DIFF
--- a/deepeval/benchmarks/big_bench_hard/big_bench_hard.py
+++ b/deepeval/benchmarks/big_bench_hard/big_bench_hard.py
@@ -246,7 +246,8 @@ class BigBenchHard(DeepEvalBaseBenchmark):
             responses: List = model.batch_generate(
                 prompts=prompts, schemas=[pydantic_model for i in prompts]
             )
-            predictions = [res.answer for res in responses]
+            predictions = [str(res.answer) for res in responses]
+            used_schema = True
         except TypeError:
             prompts = [
                 prompt + "Make sure to output only the numerical answer."
@@ -254,6 +255,7 @@ class BigBenchHard(DeepEvalBaseBenchmark):
             ]
             predictions = model.batch_generate(prompts)
             predictions = [str(pred) for pred in predictions]
+            used_schema = False
 
         if len(predictions) is not len(goldens):
             raise ValueError(
@@ -263,8 +265,14 @@ class BigBenchHard(DeepEvalBaseBenchmark):
         res = []
         for i in range(len(predictions)):
             prediction = predictions[i]
-            prediction = prediction.split()[-1]
-            prediction = prediction[:-1] if self.enable_cot else prediction
+            # Schema-constrained answers are already the final answer (e.g.
+            # "(A)"); only free-text generations need last-token extraction and
+            # trailing-punctuation trimming. Applying these to a schema answer
+            # would turn "(A)" into "(A" and silently score it 0 — and diverged
+            # from the non-batch `predict`, which never post-processes.
+            if not used_schema:
+                prediction = prediction.split()[-1]
+                prediction = prediction[:-1] if self.enable_cot else prediction
             golden = goldens[i]
 
             # Define Metric
--- a/deepeval/benchmarks/drop/drop.py
+++ b/deepeval/benchmarks/drop/drop.py
@@ -22,7 +22,12 @@ from deepeval.benchmarks.schema import (
 from deepeval.telemetry import capture_benchmark_run
 
 logger = logging.getLogger(__name__)
-DELIMITER = ","
+# Internal separator used only to pack/unpack a golden's list of answer spans
+# into its `expected_output` string. It must NOT be a character that can occur
+# inside a DROP answer span — a "," (the previous value) shattered legitimate
+# answers such as "1,000" into ["1", "000"], so the correct prediction could
+# never match. The unit-separator control char cannot appear in DROP answers.
+DELIMITER = "\x1f"
 
 
 class DROP(DeepEvalBaseBenchmark):
--- a/deepeval/benchmarks/math_qa/math_qa.py
+++ b/deepeval/benchmarks/math_qa/math_qa.py
@@ -41,7 +41,7 @@ class MathQA(DeepEvalBaseBenchmark):
         self.verbose_mode = verbose_mode
         if not confinement_instructions:
             self.confinement_instructions = (
-                "Output 'a', 'b', 'c', or 'd'. Full answer not needed."
+                "Output 'a', 'b', 'c', 'd', or 'e'. Full answer not needed."
             )
         else:
             self.confinement_instructions = confinement_instructions
@@ -215,7 +215,7 @@ class MathQA(DeepEvalBaseBenchmark):
         except TypeError:
             prompts = [
                 prompt
-                + "\n\nOutput 'a', 'b', 'c', or 'd'. Full answer not needed."
+                + "\n\nOutput 'a', 'b', 'c', 'd', or 'e'. Full answer not needed."
                 for prompt in prompts
             ]
             predictions = model.batch_generate(prompts)
--- a/deepeval/benchmarks/schema.py
+++ b/deepeval/benchmarks/schema.py
@@ -40,7 +40,11 @@ class TrinaryChoiceSchema(BaseModel):
 
 
 class MultipleChoiceSchemaLower(BaseModel):
-    answer: Literal["a", "b", "c", "d"]
+    # MathQA (AQuA-RAT) questions always have five options, a-e (option "e" is
+    # frequently "none of these"). Omitting "e" made every question whose answer
+    # is "e" unrepresentable, so schema-constrained models could never emit it
+    # and those items were always scored 0.
+    answer: Literal["a", "b", "c", "d", "e"]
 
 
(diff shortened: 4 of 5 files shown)


ebarkhordar (commented):
Nice catch on the MathQA option-(e) case. The same failure mode (a correct label the answer schema cannot represent, so the item scores 0) is present in a benchmark this PR does not touch: ARC.

`benchmarks/arc/template.py` `format_answer` returns the raw `answerKey`, and `format_question` renders each choice under its original label. But `MultipleChoiceSchema` in `benchmarks/schema.py` is `Literal["A", "B", "C", "D"]`, and `arc.py`'s confinement instruction is `Output 'A', 'B', 'C', or 'D'`. So any ARC item whose `answerKey` is not one of A-D can never be matched, no matter how the model reasons.

That set is non-empty in ARC itself. The repo's own few-shot example `NYSEDREGENTS_2006_8_10` (in `template.py`) has `label: ["1", "2", "3", "4"]` and `answerKey: "3"`; any 5-option item (`answerKey` "E") lands the same way. For those, `expected_output` is `"3"` or `"E"` while a schema-constrained model can only emit A-D, so `exact_match_score` returns 0.

Checked against this PR's head with the shipped numeric-label item: `format_answer` yields `"3"`, `MultipleChoiceSchema(answer="3")` raises a ValidationError, and each of A-D exact-matches `"3"` as 0.

Not blocking this PR, which stands on its own. Flagging in case ARC is worth folding into the same sweep or a follow-up. The pattern here is the same one you applied to MathQA (align the schema and confinement with the real label set), with the extra wrinkle that ARC labels are not always letters.

penguine-ip:
@hassaanch23 thanks for the above. its merged and will be in the next release (~1 day). Thanks!

PR #2919, opened 2026-07-18
fix: group overlapping chunks in ContextualRecallMetric (#2788)

### LINKED ISSUE #2788: [Bug] ContextualRecallMetric over-penalises overlapping chunks — parallel issue to #2594 (ContextualPrecision)

Sksalmanraza:
### What

Ports `_group_retrieval_contexts()` from `ContextualPrecisionMetric` (merged in PR #2743) to `ContextualRecallMetric`, so that `RetrievedContextData` chunks sharing the same `source` are merged before LLM evaluation.

### Why

Sliding-window chunking produces overlapping chunks from the same document. Without grouping, the LLM judge sees duplicate information across chunks and may mark the overlap as "not attributable," which artificially lowers the recall score. This is the same bug that was fixed for ContextualPrecision in PR #2743.

Fixes #2788

### Changes

**`contextual_recall.py`**: Added `_group_retrieval_contexts()` method (identical to the one in ContextualPrecision). Wired it into both `measure()` (sync) and `a_measure()` (async) before passing contexts to `_generate_verdicts()`.

**`test_contextual_recall_overlapping_chunks.py`**: Regression tests covering grouping logic **and** full `measure()` path:

- **Unit tests** (`test_group_*`): exercise `_group_retrieval_contexts()` in isolation - merging, different sources, plain strings, mixed types, partial overlap.
- **`test_measure_grouped_overlapping_chunks_full_recall`**: calls `measure()` end-to-end on overlapping `RetrievedContextData` chunks with a mocked LLM. Asserts that (a) grouping merges 2 chunks into 1, (b) the merged context is what reaches `_generate_verdicts` (verified via `call_args`), and (c) recall = 1.0. Pins the fix to the reported symptom.
- **`test_measure_ungrouped_overlapping_chunks_penalised`**: simulates the pre-fix verdict distribution, asserts recall = 0.5, demonstrating the bug scenario for contrast.

### How to Test
pytest test/text_metrics/test_contextual_recall_overlapping_chunks.py -v

### Notes

- Grouping only applies to `RetrievedContextData` objects with a `source` field, matching the existing ContextualPrecision behavior. Plain strings pass through unchanged.
- The `measure()`-level tests mock the LLM judge (no API key required) to make the tests deterministic and CI-friendly.
- The mock's `call_args` are inspected to verify that `_generate_verdicts` actually receives the merged context, not the original separate chunks - closing the loop between grouping and scoring.
- CI status: The 3 failing checks (Lint, Metrics Tests, Vercel) are pre-existing repo-wide CI issues, not introduced by this PR.

### DIFF
--- a/deepeval/metrics/contextual_recall/contextual_recall.py
+++ b/deepeval/metrics/contextual_recall/contextual_recall.py
@@ -117,7 +117,7 @@ class ContextualRecallMetric(BaseMetric):
                 )
             else:
                 expected_output = test_case.expected_output
-                retrieval_context = self._group_retrival_contexts(test_case.retrieval_context)
+                retrieval_context = self._group_retrieval_contexts(test_case.retrieval_context)
 
                 self.verdicts: List[VerdictWithExpectedOutput] = (
                     self._generate_verdicts(
@@ -166,7 +166,7 @@ class ContextualRecallMetric(BaseMetric):
             _in_component=_in_component,
         ):
             expected_output = test_case.expected_output
-            retrieval_context = self._group_retrival_contexts(test_case.retrieval_context)
+            retrieval_context = self._group_retrieval_contexts(test_case.retrieval_context)
 
             self.verdicts: List[VerdictWithExpectedOutput] = (
                 await self._a_generate_verdicts(
--- a/deepeval/metrics/contextual_recall/contextual_recall.py
+++ b/deepeval/metrics/contextual_recall/contextual_recall.py
@@ -14,6 +14,7 @@ from deepeval.metrics.utils import (
 from deepeval.test_case import (
     LLMTestCase,
     SingleTurnParams,
+    RetrievedContextData,
 )
 from deepeval.metrics import BaseMetric
 from deepeval.models import DeepEvalBaseLLM
@@ -116,7 +117,7 @@ class ContextualRecallMetric(BaseMetric):
                 )
             else:
                 expected_output = test_case.expected_output
-                retrieval_context = test_case.retrieval_context
+                retrieval_context = self._group_retrival_contexts(test_case.retrieval_context)
 
                 self.verdicts: List[VerdictWithExpectedOutput] = (
                     self._generate_verdicts(
@@ -165,7 +166,7 @@ class ContextualRecallMetric(BaseMetric):
             _in_component=_in_component,
         ):
             expected_output = test_case.expected_output
-            retrieval_context = test_case.retrieval_context
+            retrieval_context = self._group_retrival_contexts(test_case.retrieval_context)
 
             self.verdicts: List[VerdictWithExpectedOutput] = (
                 await self._a_generate_verdicts(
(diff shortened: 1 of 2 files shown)


ebarkhordar (commented):
Thanks for porting this over, the source-based grouping reads cleanly and the pass-through for plain strings and mixed lists looks right.

One thing I wanted to check on before it lands: the mechanism that made grouping work for ContextualPrecision doesn't map one-to-one onto ContextualRecall, and I don't think the current tests show that it does.

In ContextualPrecision, `_calculate_score` emits one verdict per retrieval node and computes a MAP@k over them, so merging overlapping chunks changes the node set and directly moves the score. That's why #2743 helped there.

ContextualRecall scores differently. `_generate_verdicts` (contextual_recall.py:327) produces one verdict per sentence of `expected_output`, and `_calculate_score` (:293) returns `justified_sentences / number_of_verdicts` where the denominator is the sentence count, not the chunk count. So grouping can't change the denominator here. It can only change the text the judge attributes each sentence against, which may or may not flip a per-sentence verdict.

That might still be enough to fix the under-scoring in #2788, but it's an LLM-behavior claim rather than an arithmetic one, and the added tests only exercise `_group_retrieval_contexts` in isolation, none drive `measure()` on overlapping `RetrievedContextData` and assert the recall score actually moves. Would a full-`measure()` regression test on the overlapping-chunk case, asserting the score changes, be worth adding so the fix is pinned to the reported symptom?

Sksalmanraza:
Thanks for the thorough review @ebarkhordar  the distinction between the arithmetic fix in Precision vs the LLM-behavior claim in Recall was a really good catch.

I've pushed a follow-up commit addressing this directly with two new `measure()`-level tests:

**`test_measure_grouped_overlapping_chunks_full_recall`** :  drives `measure()` end-to-end on overlapping `RetrievedContextData` chunks. The LLM is mocked so the test is deterministic and CI-friendly. It asserts three things:
1. `_group_retrieval_contexts()` merges the 2 overlapping chunks into 1
2. `_generate_verdicts` actually receives the 1 merged context, not the original 2 separate chunks (verified via `call_args`)
3. Final recall score = 1.0

**`test_measure_ungrouped_overlapping_chunks_penalised`** :  simulates the pre-fix verdict distribution (judge says "yes" for one sentence, "no" for the redundant overlap) and asserts recall = 0.5, showing the penalty from the bug in #2788.

Together the two tests show: without grouping -> 0.5, with grouping -> 1.0, and the `call_args` inspection closes the loop by proving the merged context is what reaches the judge.

Happy to adjust if you'd like a different mocking strategy or additional edge cases.

Sksalmanraza:
Hi @ebarkhordar - just circling back on this. The follow-up commit (21b6f5c) adds the two measure()-level regression tests you suggested, covering both the grouped (1.0) and ungrouped (0.5) scenarios with call_args verification.

Let me know if there's anything else you'd like adjusted.

PR #2930, opened 2026-07-20
fix(models): set additionalProperties recursively for gateway schemas

### LINKED ISSUE #2929: Gateway models send invalid strict JSON schemas for nested pydantic models
# Gateway models send invalid strict JSON schemas for nested pydantic models

**Repo:** `confident-ai/deepeval`
**Affects:** `deepeval` 4.1.0 and current `main`
**File:** `deepeval/models/llms/gateway_model.py`, `DeepEvalOpenAICompatibleModel._schema_response_format`

## Summary

Gateway-based models (`OpenRouterModel`, and any other
`DeepEvalOpenAICompatibleModel` subclass) build `response_format` manually and set
`additionalProperties: false` only on the **root** schema object. OpenAI's strict
structured-output mode requires it on **every** object, including those under
`$defs`.

Any pydantic model containing a nested model therefore gets rejected with HTTP 400,
and deepeval silently degrades to text-generation + JSON parsing.

## Reproduction (no API key, no network)

```python
from typing import List
from deepeval.models.llms.gateway_model import DeepEvalOpenAICompatibleModel as M

class Item(BaseModel):
    name: str

class Nested(BaseModel):
    items: List[Item]

schema = M._schema_response_format(Nested)["json_schema"]["schema"]
print(schema.get("additionalProperties"))              # False
print(schema["$defs"]["Item"].get("additionalProperties"))  # None  <-- bug
```

Output:

```
False
None
```

Since the request is sent with `"strict": true`, the provider rejects it.

## Observed error

```
UserWarning: Structured outputs not supported for model 'openai/gpt-4o-mini'.
Falling back to regular generation with JSON parsing.
Error: Error code: 400 - Invalid schema for response_format 'SyntheticDataList':
In context=(), 'additionalProperties' is required to be supplied and to be false.
```

Reproduced with both OpenAI and Azure as the upstream provider behind OpenRouter,
so this is not provider-specific.

## Why this matters

The fallback is not reliable. Running DeepTeam (which uses the nested
`SyntheticDataList` schema for attack simulation) the fallback failed outright:

```
DeepEvalError: Evaluation LLM outputted an invalid JSON.
Please use a better evaluation model.
```

The warning message is also misleading — it says the model does not support
structured outputs, when in fact `openai/gpt-4o-mini` does; the schema deepeval
sent was invalid.

Note `GPTModel` is unaffected: it passes the pydantic class to
`client.beta.chat.completions.parse()`, and the OpenAI SDK transforms the schema
recursively. Only the hand-rolled gateway path has this gap.

## Suggested fix

Apply the flag recursively instead of only at the root:

```python
def _harden(node):
    if isinstance(node, dict):
        if node.get("type") == "object":
            node.setdefault("additionalProperties", False)
        for value in node.values():
            _harden(value)
    elif isinstance(node, list):
        for value in node:
            _harden(value)
    return node


@staticmethod
def _schema_response_format(schema):
    json_schema = _harden(schema.model_json_schema())
    name = schema.__name__ if inspect.isclass(schema) else type(schema).__name__
    return {
        "type": "json_schema",
        "json_schema": {"name": name, "strict": True, "schema": json_schema},
    }
```

Verified locally: with this change the same call returns a validated
`SyntheticDataList` and no fallback warning is emitted.

## Environment

- deepeval 4.1.0 (bug also present on `main`)
- Python 3.12, macOS
- Model: `openai/gpt-4o-mini` via OpenRouter

eno42:
Fixes #2929.

## Problem

`DeepEvalOpenAICompatibleModel._schema_response_format` builds `response_format` by hand and sets `additionalProperties: false` only on the **root** schema object:

```python
# `strict: true` requires additionalProperties to be set at the root.
json_schema.setdefault("additionalProperties", False)
```

OpenAI's strict structured-output mode requires the flag on **every** object, including nested models emitted under `$defs`. Any schema with a nested pydantic model is rejected:

```
400 - Invalid schema for response_format 'SyntheticDataList':
In context=(), 'additionalProperties' is required to be supplied and to be false.
```

deepeval catches this and falls back to text generation + JSON parsing, emitting a misleading warning that says the model doesn't support structured outputs — `openai/gpt-4o-mini` does; the schema we sent was invalid.

The fallback is not always able to recover. Running DeepTeam (whose `SyntheticDataList` schema is nested) it failed outright with `DeepEvalError: Evaluation LLM outputted an invalid JSON`.

`GPTModel` is unaffected because it hands the pydantic class to `client.beta.chat.completions.parse()` and the OpenAI SDK transforms the schema recursively. Only the hand-rolled gateway path has this gap.

## Fix

Walk the schema and set the flag on every object node. `setdefault` is used so schemas that deliberately opt into permissiveness (`model_config = {"extra": "allow"}`) keep `additionalProperties: true`.

## Tests

Three cases added to `tests/test_core/test_models/test_openrouter_model.py`:

- flat schema still gets the flag at the root
- nested schema gets it inside `$defs` — **fails on `main`** with `KeyError: 'additionalProperties'`, passes with this change
- an explicitly permissive schema is not overwritten

```
tests/test_core/test_models/test_openrouter_model.py  21 passed
tests/test_core/test_models/                         207 passed
```

(The one unrelated failure in the wider run is `test_ollama_embedding_model`, which needs the optional `ollama` package.)

Verified end-to-end against OpenRouter with the real nested schema: before the change the call falls back and then fails to parse; after it returns a validated `SyntheticDataList` with no warning.

### DIFF
--- a/deepeval/models/llms/gateway_model.py
+++ b/deepeval/models/llms/gateway_model.py
@@ -27,7 +27,7 @@ Two layers live here:
 
 import inspect
 import warnings
-from typing import Dict, List, Optional, Tuple, Type, Union
+from typing import Any, Dict, List, Optional, Tuple, Type, Union
 
 from openai import OpenAI, AsyncOpenAI
 from pydantic import BaseModel, SecretStr
@@ -359,6 +359,29 @@ class DeepEvalOpenAICompatibleModel(DeepEvalBaseGatewayModel):
         return content
 
     @staticmethod
+    def _set_additional_properties_false(node: Any) -> Any:
+        """Recursively set `additionalProperties: false` on every object node.
+
+        `strict: true` requires the flag on *every* object in the schema, not
+        just the root. Nested pydantic models are emitted under `$defs`, so
+        setting it only at the root makes the provider reject the request with
+        a 400.
+        """
+        if isinstance(node, dict):
+            if node.get("type") == "object":
+                node.setdefault("additionalProperties", False)
+            for value in node.values():
+                DeepEvalOpenAICompatibleModel._set_additional_properties_false(
+                    value
+                )
+        elif isinstance(node, list):
+            for value in node:
+                DeepEvalOpenAICompatibleModel._set_additional_properties_false(
+                    value
+                )
+        return node
+
+    @staticmethod
     def _schema_response_format(
         schema: Union[Type[BaseModel], BaseModel],
     ) -> Dict:
@@ -368,8 +391,11 @@ class DeepEvalOpenAICompatibleModel(DeepEvalBaseGatewayModel):
             if inspect.isclass(schema)
             else schema.__class__.__name__
         )
-        # `strict: true` requires additionalProperties to be set at the root.
-        json_schema.setdefault("additionalProperties", False)
+        json_schema = (
+            DeepEvalOpenAICompatibleModel._set_additional_properties_false(
+                json_schema
+            )
+        )
         return {
             "type": "json_schema",
             "json_schema": {
--- a/deepeval/models/llms/gateway_model.py
+++ b/deepeval/models/llms/gateway_model.py
@@ -360,16 +360,22 @@ class DeepEvalOpenAICompatibleModel(DeepEvalBaseGatewayModel):
 
     @staticmethod
     def _set_additional_properties_false(node: Any) -> Any:
-        """Recursively set `additionalProperties: false` on every object node.
+        """Force `additionalProperties: false` on every object node.
 
-        `strict: true` requires the flag on *every* object in the schema, not
-        just the root. Nested pydantic models are emitted under `$defs`, so
-        setting it only at the root makes the provider reject the request with
-        a 400.
+        The response_format is always sent with `strict: true`, and strict mode
+        requires `additionalProperties: false` on *every* object in the schema.
+        Nested pydantic models are emitted under `$defs`, so this must recurse.
+
+        The flag is OVERWRITTEN, not defaulted: a schema that explicitly sets
+        `additionalProperties: true` (e.g. a pydantic model with
+        `extra="allow"`) is rejected under strict mode — the provider requires
+        the value to be false, not merely present. `extra="allow"` cannot be
+        honored in strict structured output, so we force the schema closed to
+        keep it valid rather than emit a request the provider will 400.
         """
         if isinstance(node, dict):
             if node.get("type") == "object":
-                node.setdefault("additionalProperties", False)
+                node["additionalProperties"] = False
             for value in node.values():
                 DeepEvalOpenAICompatibleModel._set_additional_properties_false(
                     value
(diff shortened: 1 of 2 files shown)


ebarkhordar (commented):
Nice fix. I reproduced #2929 in a clean container against both shas to sanity-check the recursion:

- base `6bb1881`: `_schema_response_format` on a schema with a nested model yields the `$defs` entry with no `additionalProperties`, which is the 400 in the issue.
- this PR `712cf6e`: every `$defs` object gets `additionalProperties: false`, including a model nested two levels deep (a `$defs` model that itself holds another model). Your three new tests pass.

One small suggestion: `test_nested_schema_sets_additional_properties_in_defs` pins a single level of `$defs` nesting. A two-level case (a nested model that itself holds another model) would guard the recursive descent, so a later change that only walked the top-level `$defs` entries could not regress it quietly. I confirmed the current code already handles that case.

eno42:
Thanks for the careful review — reproducing it against both shas is appreciated.

I went to add the two-level `$defs` case and found it doesn't actually guard the recursion, which is worth flagging: Pydantic emits **every** model as a flat `$defs` entry, so a `$defs` model never nests inside another `$defs` model. `SchemaWithNestedModel` and a would-be `DeeplyNestedItem` both land as sibling top-level `$defs` keys. I confirmed a deliberately shallow implementation (root + `node["$defs"].values()` only, no recursion) still passes a two-level `$defs` assertion — so that test wouldn't catch the regression you're guarding against.

The property that *does* need the recursion is an object nested inside another object's `properties` (hand-built schemas, or if Pydantic's layout ever changes). I added `test_recurses_into_inline_nested_objects` that calls `_set_additional_properties_false` on such a schema directly. Verified it fails against the shallow implementation (`KeyError` on the inner object's `additionalProperties`) and passes with the recursive one.

Pushed as `f427cce`. Happy to also keep a two-level `$defs` case as documentation if you'd prefer both, just let me know.

penguine-ip:
Thanks for the thoughtful fix and the clear explanation around testing the recursive descent! One question: since this response format always sets `strict: true`, and strict mode requires `additionalProperties: false` on every object, should the helper overwrite an explicit `true` instead of using `setdefault`? As written, the permissive-schema case appears to preserve a schema the provider will reject—was the intent instead to disable strict mode or fall back for `extra="allow"` models?

eno42:
You're right — with `strict: true` hardcoded here, an explicit `additionalProperties: true` is invalid: the provider requires the value to be **false**, not merely present (its own error is "required to be supplied and to be false"). `setdefault` preserved that `true` for `extra="allow"` models and would 400. Since `extra="allow"` cannot be honored under strict structured output anyway, the intent is to force the schema closed rather than disable strict.

Changed the helper to overwrite instead of default, and flipped the test accordingly:

- `_set_additional_properties_false` now assigns `node["additionalProperties"] = False` on every object.
- `test_permissive_schema_is_forced_closed` (was `..._is_not_overwritten`) now asserts an `extra="allow"` schema comes out with `additionalProperties: false`, after confirming pydantic emits `true` for it.

Verified: permissive → `false`, nested `$defs` still → `false`, 22 tests pass. Pushed as `54014cf`.

I kept strict mode on rather than falling back to non-strict for permissive schemas — deepteam's own schemas are all closed, and forcing the flag closed turns a guaranteed 400 into a valid (stricter) request. Happy to switch to a strict-disable fallback instead if you'd prefer to preserve `extra="allow"` semantics for gateway users.

penguine-ip:
@eno42 it looks great, thanks!

PR #2939, opened 2026-07-21
fix: serialize tool call enums in datasets

yaodong-shen:
## Summary

- serialize nested Pydantic models in JSON mode when formatting conversational turns
- convert `ToolCallType` enum values to JSON-safe strings before the final `json.dumps`
- strengthen both JSON and JSONL export coverage to assert the persisted enum value

## Problem

`EvaluationDataset.save_as(...)` raises `ValueError: Object of type ToolCallType is not JSON serializable` for conversational goldens containing tool calls. The nested model was dumped in Python mode, which preserves enum objects that the standard JSON encoder cannot serialize.

## Validation

- `uvx poetry run pytest tests/test_core -q --maxfail=1` — 1111 passed, 56 skipped
- `uvx ruff check deepeval/dataset/utils.py tests/test_core/test_datasets/test_dataset.py`
- `uvx ruff format --check deepeval/dataset/utils.py tests/test_core/test_datasets/test_dataset.py`
- `uvx poetry run black --check deepeval/dataset/utils.py tests/test_core/test_datasets/test_dataset.py`

### DIFF
--- a/deepeval/dataset/utils.py
+++ b/deepeval/dataset/utils.py
@@ -187,7 +187,9 @@ def format_turns(turns: List[Turn]) -> str:
             for m in models:
                 if hasattr(m, "model_dump"):
                     dumped.append(
-                        m.model_dump(by_alias=True, exclude_none=True)
+                        m.model_dump(
+                            mode="json", by_alias=True, exclude_none=True
+                        )
                     )
                 elif hasattr(m, "dict"):
                     dumped.append(m.dict(exclude_none=True))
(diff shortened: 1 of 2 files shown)


yaodong-shen:
CI note: I rebased this PR onto current `main` (`c6293c12`) and independently ran the CI version, `black==26.5.1`, in a clean detached worktree of that exact upstream commit. It reports the same baseline result: 50 files would be reformatted and 859 left unchanged. Neither file changed by this PR is among those 50; both pass Black 26.5.1 directly. The refreshed full core suite also passes (1113 passed, 56 skipped).

PR #2947, opened 2026-07-23
fix(models): extract Anthropic text block by type, not content[0]

### LINKED ISSUE #2946: AnthropicModel crashes with AttributeError when extended thinking is enabled (content[0] is a ThinkingBlock)
## Describe the bug

`AnthropicModel` reads `message.content[0].text` in both `generate()` and `a_generate()`. That indexing assumes the first content block is always a `TextBlock`. It isn't once extended thinking is on: the API returns `content=[ThinkingBlock, TextBlock, ...]`, and `ThinkingBlock` has no `.text`. Every metric evaluation using that judge then dies with an `AttributeError`.

This is the same `content[0]` assumption reported in #2913 for the Anthropic tracing integration, but in a different file: `deepeval/models/llms/anthropic_model.py` (lines 164/166 and 204/206). The open PR for #2913 only touches `deepeval/anthropic/extractors.py`, so the judge wrapper keeps the bug.

## To Reproduce

`AnthropicModel` forwards `generation_kwargs` straight into `messages.create(...)`, so enabling thinking is a supported configuration:

```python

model = AnthropicModel(
    model="claude-3-7-sonnet-latest",
    generation_kwargs={"thinking": {"type": "enabled", "budget_tokens": 1024}},
)
model.generate("What is 2 + 2?")
```

Output:

```
AttributeError: 'ThinkingBlock' object has no attribute 'text'
```

You can confirm the shape without an API call, straight from the SDK types:

```python
from anthropic.types import Message, TextBlock, ThinkingBlock, Usage

msg = Message(
    id="msg_test", model="claude-3-7-sonnet-latest", role="assistant",
    stop_reason="end_turn", type="message",
    usage=Usage(input_tokens=10, output_tokens=20),
    content=[
        ThinkingBlock(type="thinking", thinking="reasoning...", signature="sig"),
        TextBlock(type="text", text='{"score": 0.9}'),
    ],
)
msg.content[0].text  # AttributeError
```

## Expected behavior

`generate()` returns the text block's content (`{"score": 0.9}` above) regardless of where it sits in the content list. If the response has no text block at all, a clear `DeepEvalError` beats a raw `AttributeError`.

## Environment

- deepeval: current `main` (4.1.3)
- anthropic SDK: 0.118.0
- Python 3.11, macOS

Happy to submit the fix, I have a patch with regression tests ready.

bharadwaj-pendyala:
Fixes #2946

## Summary

- `AnthropicModel.generate()` and `a_generate()` now pick the first `text` block from `message.content` instead of assuming `content[0]` is it.
- Raises a clear `DeepEvalError` (listing the block types received) when a response has no text block, instead of a raw `AttributeError`.
- Adds sync and async regression tests built from real `anthropic.types` objects with a thinking-first content list.

## Why

Enabling extended thinking through `generation_kwargs={"thinking": {...}}` makes the API return `content=[ThinkingBlock, TextBlock]`. `content[0].text` then raises `AttributeError: 'ThinkingBlock' object has no attribute 'text'` and every metric using that judge fails. Same `content[0]` assumption as #2913, different file: that one is the tracing integration, this one is the judge model wrapper.

## Test plan

- [x] New tests fail on `main` with `AttributeError: 'ThinkingBlock' object has no attribute 'text'`, pass with the fix
- [x] `pytest tests/test_core/test_models/` passes (211 tests)
- [x] `black --check` clean on both files

### DIFF
--- a/deepeval/models/llms/anthropic_model.py
+++ b/deepeval/models/llms/anthropic_model.py
@@ -160,10 +160,11 @@ class AnthropicModel(DeepEvalBaseLLM):
         cost = self.calculate_cost(
             message.usage.input_tokens, message.usage.output_tokens
         )
+        text = self._extract_text(message)
         if schema is None:
-            return message.content[0].text, cost
+            return text, cost
         else:
-            json_output = trim_and_load_json(message.content[0].text)
+            json_output = trim_and_load_json(text)
             return schema.model_validate(json_output), cost
 
     @retry_anthropic
@@ -200,12 +201,24 @@ class AnthropicModel(DeepEvalBaseLLM):
         cost = self.calculate_cost(
             message.usage.input_tokens, message.usage.output_tokens
         )
+        text = self._extract_text(message)
         if schema is None:
-            return message.content[0].text, cost
+            return text, cost
         else:
-            json_output = trim_and_load_json(message.content[0].text)
+            json_output = trim_and_load_json(text)
 
             return schema.model_validate(json_output), cost
+
+    def _extract_text(self, message) -> str:
+        # With extended thinking enabled, content starts with thinking blocks;
+        # the text block is not guaranteed to be first.
+        for block in message.content:
+            if getattr(block, "type", None) == "text":
+                return block.text
+        raise DeepEvalError(
+            f"Anthropic response from `{self.name}` contained no text block "
+            f"(content types: {[getattr(b, 'type', '?') for b in message.content]})."
+        )
 
     def generate_content(self, multimodal_input: List[Union[str, MLLMImage]]):
         content = []
--- a/deepeval/models/llms/anthropic_model.py
+++ b/deepeval/models/llms/anthropic_model.py
@@ -163,10 +163,11 @@ class AnthropicModel(DeepEvalBaseLLM):
         cost = self.calculate_cost(
             message.usage.input_tokens, message.usage.output_tokens
         )
+        text = self._extract_text(message)
         if schema is None:
-            return message.content[0].text, cost
+            return text, cost
         else:
-            json_output = trim_and_load_json(message.content[0].text)
+            json_output = trim_and_load_json(text)
             return schema.model_validate(json_output), cost
 
     @retry_anthropic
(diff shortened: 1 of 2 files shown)


bharadwaj-pendyala:
Following up after 12 days. One thing that may be putting you off this: the red `lint (ubuntu-latest)` does not come from this change.

black reports both files the PR touches as clean, and fails on files it never went near:

```
deepeval/models/llms/anthropic_model.py already well formatted, good job.
tests/test_core/test_models/test_anthropic_model.py already well formatted, good job.
Oh no! 50 files would be reformatted, 867 files would be left unchanged.
```

Those 50 are pre-existing on `main`. The same check is red on #2990, #2979 and #2941, all merged, so it does not look like a gate you apply.

Nothing has touched `anthropic_model.py` or its test since I opened this, so the diff still applies. The branch is 59 commits behind `main` with no conflicts. Happy to rebase if you want it current.

The bug in one line: with extended thinking on, Anthropic returns `content=[ThinkingBlock, TextBlock]`, and `content[0].text` raises `AttributeError: 'ThinkingBlock' object has no attribute 'text'`, so every metric using that judge fails. Same `content[0]` assumption you already fixed in #2913 for the tracing integration, just in the judge model wrapper.

bharadwaj-pendyala:
Rebased onto `main` at d2de18c and pushed two changes.

`lint (ubuntu-latest)` was red on files this branch never touched, so I rebased. It passes now, and so does every other check except Vercel, which fails with "Authorization required to deploy" on any fork branch.

I also took @ebarkhordar's point from #2946 about the schema branch and found the tests did not actually cover it. Both original tests called `generate()` with `schema=None`, so a fix that only repaired the plain-text return would have passed them while metric scoring stayed broken. There are now two more tests that go through `trim_and_load_json` with a thinking-first response and assert the parsed model comes back:

```
test_anthropic_model_generate_parses_schema_when_thinking_block_first
test_anthropic_model_a_generate_parses_schema_when_thinking_block_first
```

All four fail on `main` with `AttributeError: 'ThinkingBlock' object has no attribute 'text'` and pass here. Full file: 14 passed.

Second change is smaller. The thinking tests built the model with `budget_tokens=1024` and no `max_tokens`, which falls back to the 1024 default. The SDK docstring for `ThinkingConfigEnabledParam.budget_tokens` says it "Must be >=1024 and less than `max_tokens`", so that pair is a request the API would reject. The tests now pass `max_tokens=2048`, which is a configuration a real caller can send.

On #3043: I simulated the merge locally and `deepeval/models/llms/anthropic_model.py` merges clean, so the two fixes do compose. `tests/test_core/test_models/test_anthropic_model.py` conflicts because both PRs append after line 313. Whichever lands second is a trivial fixup, and I am happy to rebase on top of #3043 if you would rather take that one first.

bharadwaj-pendyala:
Superseded by 3b8f1707d, closing.

That commit landed the same by-type scan in `AnthropicModel._extract_text`, and its version is the better one. It guards `content` being absent as well as empty, and it arrived alongside the `thinking` config work, which is the feature that makes this path reachable without a raw `generation_kwargs` override.

Verified on `main` at a2e0d4c that the crash from #2946 is gone:

```python
msg = SimpleNamespace(content=[
    SimpleNamespace(type="thinking", thinking="2+2..."),
    SimpleNamespace(type="text", text="4"),
])
AnthropicModel._extract_text(msg)  # '4'
```

`content[0].text` on that same message is the `AttributeError` the issue reported.

`tests/test_core/test_models/test_anthropic_model.py` passes 19/19 on `main`, and the two cases this PR added are covered there by `test_anthropic_skips_leading_thinking_block` and `test_anthropic_raises_when_response_has_no_text_block`. Nothing here is left to merge.

PR #2949, opened 2026-07-24
fix(models): send temperature on non-schema path for DeepSeek and Kimi

sanidhyasin:
## Summary

`DeepSeekModel` and `KimiModel` only forward the configured `temperature` on their **schema / JSON** generation path. On the plain-text path — `generate()`/`a_generate()` called without a `schema` (and, for Kimi, also when `supports_json_mode()` is `False`) — the `chat.completions.create()` call omits `temperature`, so the request silently runs at the provider's server-side default instead of the value the user set.

This makes `DeepSeekModel(temperature=0.0)` / `KimiModel(temperature=0.0)` behave inconsistently: deterministic when a schema is passed, non-deterministic for a plain prompt.

The sibling `GrokModel` already applies `temperature` uniformly (set once on `chat.create(...)`, used by both paths), and both DeepSeek and Kimi already send `temperature` unconditionally on their schema path — so aligning the plain path introduces no new behavior for any model.

## Changes

- Add `temperature=self.temperature` to the non-schema `create()` calls in `DeepSeekModel.generate`/`a_generate` and `KimiModel.generate`/`a_generate` (4 call sites).
- Regression tests (sync + async, both models) that mock the client and assert `temperature` is forwarded on the no-schema path.

## Testing

```
pytest tests/test_core/test_models/test_deepseek_model.py tests/test_core/test_models/test_kimi_model.py
```

All pass; the new tests fail on `main` (the kwarg is absent) and pass with the fix. `black --line-length 80` clean.

### DIFF
--- a/deepeval/models/llms/deepseek_model.py
+++ b/deepeval/models/llms/deepseek_model.py
@@ -128,6 +128,7 @@ class DeepSeekModel(DeepEvalBaseLLM):
             completion = client.chat.completions.create(
                 model=self.name,
                 messages=[{"role": "user", "content": prompt}],
+                temperature=self.temperature,
                 **self.generation_kwargs,
             )
             output = completion.choices[0].message.content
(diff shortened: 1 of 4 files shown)


sanidhyasin:
@penguine-ip flagging this one for review when you have bandwidth, happy to adjust if you'd like it scoped differently.

PR #2950, opened 2026-07-24
fix(openai): apply prompt-caching discount to GPTModel cost calculation

Aftabbs:
## Problem

`GPTModel.calculate_cost` bills **all** input tokens at the regular `input_price`, even when OpenAI served a cached prefix at a discounted rate.

OpenAI's Chat Completions API returns `usage.prompt_tokens_details.cached_tokens` whenever prompt caching kicks in. Those tokens are billed at ~50–75 % of the standard input price depending on the model (e.g. $1.25/M instead of $2.50/M for gpt-4o). Without reading that field, cost tracking over-reports spend for any evaluation run that hits the cache.

## Changes

| File | What changed |
|---|---|
| `deepeval/models/base_model.py` | Added `cache_read_input_price: Optional[float]` to `DeepEvalModelData` |
| `deepeval/models/llms/constants.py` | Added `cache_read_input_price` for gpt-4o family, gpt-4o-mini, gpt-4.1 / mini / nano, o1, o3-mini, o4-mini |
| `deepeval/models/llms/openai_model.py` | Added `_cached_tokens_from_usage()` helper; updated `calculate_cost` signature; updated all 8 call sites |
| `deepeval/models/utils.py` | Added `cached_tokens` attribute to `EvaluationCost` |
| `tests/…/test_openai_model.py` | 12 new unit tests |

## Pricing added (per OpenAI pricing page)

| Model | Input | Cached input |
|---|---|---|
| gpt-4o (all dated variants) | $2.50/M | $1.25/M |
| gpt-4o-mini (all dated variants) | $0.150/M | $0.075/M |
| gpt-4.1 | $2.00/M | $0.50/M |
| gpt-4.1-mini | $0.40/M | $0.10/M |
| gpt-4.1-nano | $0.10/M | $0.025/M |
| o1 / o1-preview / o1-2024-12-17 | $15.00/M | $7.50/M |
| o3-mini | $1.10/M | $0.55/M |
| o4-mini | $1.10/M | $0.275/M |

## Backwards compatibility

- `calculate_cost(input_tokens, output_tokens)` signature still works; `cached_tokens=0` is the default.
- `EvaluationCost` still subclasses `float`; all existing callers are unaffected.
- Models without `cache_read_input_price` continue to use full input pricing regardless of cached tokens.

## Test plan

- `TestCachedTokensFromUsage` — helper extracts correctly or returns 0
- `TestCalculateCostWithCaching` — discount applied when field set; falls back when not; EvaluationCost carries `cached_tokens`
- `TestGeneratePassesCachedTokens` — end-to-end: `generate()` reads `prompt_tokens_details`, passes to `calculate_cost`, reflects in returned cost

```
python -c "
from types import SimpleNamespace
from deepeval.models.llms.openai_model import _cached_tokens_from_usage, GPTModel
import os; os.environ['OPENAI_API_KEY'] = 'test-key'
model = GPTModel(model='gpt-4o')
cost = model.calculate_cost(1000, 500, cached_tokens=400)
# 600 * 2.50/1e6 + 400 * 1.25/1e6 + 500 * 10.00/1e6 = 0.007000
print(f'Cost with 400 cached: \${float(cost):.6f}')
cost_full = model.calculate_cost(1000, 500)
print(f'Cost without cache:  \${float(cost_full):.6f}')
assert float(cost) < float(cost_full)
print('OK')
"
```

### DIFF
--- a/deepeval/models/llms/constants.py
+++ b/deepeval/models/llms/constants.py
@@ -333,6 +333,7 @@ OPENAI_MODELS_DATA = ModelDataRegistry(
             supports_temperature=False,
             input_price=1.25 / 1e6,
             output_price=10.00 / 1e6,
+            cache_read_input_price=0.125 / 1e6,
         ),
         "gpt-5-2025-08-07": make_model_data(
             supports_log_probs=False,
@@ -342,6 +343,7 @@ OPENAI_MODELS_DATA = ModelDataRegistry(
             supports_temperature=False,
             input_price=1.25 / 1e6,
             output_price=10.00 / 1e6,
+            cache_read_input_price=0.125 / 1e6,
         ),
         "gpt-5-mini": make_model_data(
             supports_log_probs=False,
@@ -351,6 +353,7 @@ OPENAI_MODELS_DATA = ModelDataRegistry(
             supports_temperature=False,
             input_price=0.25 / 1e6,
             output_price=2.00 / 1e6,
+            cache_read_input_price=0.025 / 1e6,
         ),
         "gpt-5-mini-2025-08-07": make_model_data(
             supports_log_probs=False,
@@ -360,6 +363,7 @@ OPENAI_MODELS_DATA = ModelDataRegistry(
             supports_temperature=False,
             input_price=0.25 / 1e6,
             output_price=2.00 / 1e6,
+            cache_read_input_price=0.025 / 1e6,
         ),
         "gpt-5-nano": make_model_data(
             supports_log_probs=False,
@@ -369,6 +373,7 @@ OPENAI_MODELS_DATA = ModelDataRegistry(
             supports_temperature=False,
             input_price=0.05 / 1e6,
             output_price=0.40 / 1e6,
+            cache_read_input_price=0.005 / 1e6,
         ),
         "gpt-5-nano-2025-08-07": make_model_data(
             supports_log_probs=False,
@@ -378,6 +383,7 @@ OPENAI_MODELS_DATA = ModelDataRegistry(
             supports_temperature=False,
             input_price=0.05 / 1e6,
             output_price=0.40 / 1e6,
+            cache_read_input_price=0.005 / 1e6,
         ),
         "gpt-5-chat-latest": make_model_data(
             supports_log_probs=False,
@@ -386,6 +392,7 @@ OPENAI_MODELS_DATA = ModelDataRegistry(
             supports_json=False,
             input_price=1.25 / 1e6,
             output_price=10.00 / 1e6,
+            cache_read_input_price=0.125 / 1e6,
         ),
         "gpt-5.1": make_model_data(
             supports_log_probs=False,
@@ -395,6 +402,7 @@ OPENAI_MODELS_DATA = ModelDataRegistry(
             supports_temperature=False,
             input_price=1.25 / 1e6,
             output_price=10.00 / 1e6,
+            cache_read_input_price=0.125 / 1e6,
         ),
         "gpt-5.2": make_model_data(
             supports_log_probs=False,
@@ -404,6 +412,7 @@ OPENAI_MODELS_DATA = ModelDataRegistry(
             supports_temperature=False,
             input_price=1.75 / 1e6,
             output_price=14.00 / 1e6,
+            cache_read_input_price=0.175 / 1e6,
         ),
         "gpt-5.4": make_model_data(
             supports_log_probs=True,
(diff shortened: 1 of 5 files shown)


deepeval/models/llms/constants.py
```diff
@@ -307,6 +323,7 @@ def make_model_data(**kwargs: Any) -> ModelDataFactory:
             supports_temperature=False,
             input_price=1.10 / 1e6,
             output_price=4.40 / 1e6,
+            cache_read_input_price=0.275 / 1e6,
```
gnanirahulnutakki, line 326: [P1] Cover the default GPT-5.4 model

The cache-price additions stop immediately before the GPT-5 entries, although `DEFAULT_GPT_MODEL` is currently `gpt-5.4`. On this exact head, `GPTModel()` therefore has `cache_read_input_price is None`, and `calculate_cost(1_000_000, 0, cached_tokens=1_000_000)` returns `$2.50`; OpenAI documents GPT-5.4 cached input at `$0.25/M`, so the default evaluation path still over-reports a fully cached prompt by 10×: https://developers.openai.com/api/docs/models/gpt-5.4. The existing GPT-5, GPT-5.2, and GPT-5.4-mini entries also have documented cached rates but retain the full-price fallback. Please populate the supported GPT-5-family entries and add at least a default-model regression case.

deepeval/models/llms/openai_model.py
```diff
@@ -33,6 +33,11 @@
 retry_openai = create_retry_decorator(PS.OPENAI)
 
 
+def _cached_tokens_from_usage(usage: Any) -> int:
+    details = getattr(usage, "prompt_tokens_details", None)
+    return getattr(details, "cached_tokens", 0) or 0
```
gnanirahulnutakki: [P1] Keep existing completion paths from passing a `Mock` as a token count

This helper does not return `0` for the repository's existing mocked usage objects: `Mock` synthesizes `prompt_tokens_details.cached_tokens` as another truthy `Mock`, which then reaches `input_tokens - cached_tokens` and raises `TypeError`. On this exact head the affected test file is `4 failed, 48 passed`; the four failures cover sync generation, structured generation, async generation, and raw responses, and CI reports the same root cause. Please either make the extractor return only a real integer (treating other values as absent) or explicitly model `prompt_tokens_details=None` in every existing completion fixture, with the full file green—not only the newly added cache tests.

gnanirahulnutakki (commented):
Reviewed exact head `a14a307c0449a40dfefcf8d4e9aaa13db4062065` against OpenAI's current model pricing and the affected model test suite. The 15 newly added cache tests pass, but the full file regresses (`4 failed, 48 passed`), and the current/default GPT-5 family still falls back to full input pricing. I left both reproducible blockers inline. The changed test file also fails the repository-pinned Black check, which is already visible in CI, so I have not duplicated that as another thread.

AI-assisted review; I verified each finding locally and against the exact PR head.

Aftabbs:
The lint CI failure is unrelated to this PR — it flags formatting issues in pre-existing files (`deepeval/cli/inspect.py`, `deepeval/inspect/widgets/*.py`). Our changed files all pass `black --line-length 80`. Happy to rebase once main is green, or the CI can be ignored for those paths if preferred.

gnanirahulnutakki (commented):
Re-reviewed on `a7a4a718`. The new integer guard fixes the existing mocked-completion regression: the full affected test file is now green locally (`52 passed`), and Black passes on both changed files. The separate GPT-5.4 pricing finding remains unresolved: `DEFAULT_GPT_MODEL` is still `gpt-5.4`, its `cache_read_input_price` is still `None`, and a fully cached 1M-token input still calculates as `$2.50`; the official cached-input rate remains `$0.25/M` (https://developers.openai.com/api/docs/models/gpt-5.4). I did not add another inline finding because the existing GPT-5.4 thread already covers the required change.

deepeval/models/llms/constants.py
gnanirahulnutakki: Re-tested exact head `a2758d42a170426ae9f2363a4a23bb057e17d378`. The new standard cached-input constants fix the original default-model case below the long-context boundary, but one billing error remains: OpenAI documents GPT-5.4 and GPT-5.5 requests above 272K input tokens at 2x input and 1.5x output for the full request. `GPTModel(api_key="test-key").calculate_cost(300_000, 0, cached_tokens=300_000)` returns `$0.075` here; the documented cached-input result is `$0.15` (`300,000 x $0.25/M x 2`). The same reproduces on a synthetic merge with current main, while the full affected test file passes (`52 passed`). Please model the threshold and multipliers for GPT-5.4/GPT-5.5 snapshots and add default-model regressions on both sides of 272K: https://developers.openai.com/api/docs/models/gpt-5.4

PR #2953, opened 2026-07-24
docs(ragas): correct RagasMetric default threshold to 0.3

dfedoryshchev:
The `RagasMetric` parameter list documents the default `threshold` as `0.5`, but `RagasMetric.__init__` in `deepeval/metrics/ragas.py` defaults it to `0.3`. This updates the docs to match the source.


dfedoryshchev:
Hi, anything you would like changed on this one? It has been quiet a while. The test matrix and the framework integration jobs are green. Happy to rebase or reword if that helps.

PR #2926, opened 2026-07-19
fix(dataset): serialize tool call enums in JSON mode

nightcityblade:
## What Problem This Solves

Dataset saves fail when a `ToolCall` contains its default `ToolCallType` enum because Python-mode Pydantic dumps leave the enum object for `json.dumps()`.

## Why This Change Was Made

Serialize nested tool models with Pydantic's JSON mode in conversational and single-turn JSON, JSONL, and CSV paths. This preserves aliases while converting enum values to JSON-safe strings. The regression assertions verify that `ToolCallType.FUNCTION` is emitted as `"FUNCTION"`.

## User Impact

Users can save datasets containing tool calls without `Object of type ToolCallType is not JSON serializable` errors.

## Evidence

- `pytest -q tests/test_core/test_datasets/test_dataset.py` — 18 passed
- `black --check deepeval/dataset/utils.py deepeval/dataset/dataset.py tests/test_core/test_datasets/test_dataset.py` — passed
- `ruff check --ignore F401 deepeval/dataset/utils.py deepeval/dataset/dataset.py tests/test_core/test_datasets/test_dataset.py` — passed (`dataset.py` has unrelated pre-existing F401 findings)

Fixes confident-ai/deepeval#2902

PR #2954, opened 2026-07-25
fix(metrics): preserve whitespace in exact and pattern matching

CTWalk:
## Summary

- Compare the supplied strings as given in `ExactMatchMetric`.
- Compile the supplied regex unchanged and match it against the complete `actual_output` in `PatternMatchMetric`.
- Add keyless regression tests for the whitespace and case behavior.
- Correct an adjacent `ignore_case=True` documentation typo.

Rebased onto current `main`. **No existing test is modified or removed** — the new tests are a separate file, and the four existing tests in `tests/test_metrics/test_exact_match_metric.py` and `test_pattern_match_metric.py` still pass unchanged under this change.

## Problem

`ExactMatchMetric` is documented as a strict, whitespace-sensitive string equality check, but it strips both values:

```python
expected = test_case.expected_output.strip()
actual = test_case.actual_output.strip()
```

So `actual_output="act "` passes against `expected_output="act"`.

`PatternMatchMetric` strips both its pattern and the actual output before `re.fullmatch`, so `pattern="act"` passes on `actual_output=" act "` even though the docs say the regex applies to the entire output.

Reproduced on `0d100e37d` with no API key:

```text
ExactMatch:   actual="act ",  expected="act"  -> score 1.0
ExactMatch:   actual=" act",  expected="act"  -> score 1.0
PatternMatch: pattern="act",  actual=" act "  -> score 1.0
```

This hides representation defects in enum, classification, protocol-token, and structured-output tests — the cases the docs recommend deterministic matching for.

## Behavior after this change

- `"act"` exactly equals `"act"`; `"act "` does not.
- regex `"act"` does not full-match `" act "`; regex `r"\s*act\s*"` does.
- callers can still normalize before evaluation when that is their intended contract.

## Compatibility

**This changes results for callers relying on the current undocumented trimming.** The most likely breakage is a trailing newline: a suite comparing `"yes\n"` against `"yes"` passes today and fails after this change.

The change restores the published whitespace-sensitive and `re.fullmatch` contracts, and tolerant behavior stays expressible through caller preprocessing or an explicit `r"\s*...\s*"` regex.

If that trade isn't one you want to make, two smaller positions are available and I'm happy to cut this down to either:

1. **Output-side only** — keep the constructor's `pattern.strip()` (patterns come from source code, where a stray space is likelier a typo than intent) and remove only the `actual_output` strip.
2. **Docs only** — drop the behavior change and just document the trimming as intended.

Say which contract you want and I'll adjust the tests to match.

## Scope

No thresholds, scoring formulas, public types, or model-backed metrics change. This intentionally adds no normalization option — callers can normalize before evaluation or express accepted whitespace in the regex.

## Verification

- [x] `pytest tests/test_metrics/test_match_whitespace.py` — 13 passed with `OPENAI_API_KEY` unset; 7 of the 13 fail on unpatched `main`, so they're real regression tests
- [x] the four pre-existing tests in the two matcher test files pass unchanged under this change (verified with the key set, so the `skipif` doesn't hide them)
- [x] `black --check` on all changed files — clean

One note on what I did not verify: `black --check .` across the repo reports pre-existing files needing reformatting on unmodified `main`, unrelated to this PR, so I checked only the files this PR touches. The `lint` job appears red for that reason on other open PRs too. I left the repo-wide reformat out deliberately to keep this diff at 4 files — happy to add it if you'd like.

### DIFF
--- a/deepeval/metrics/exact_match/exact_match.py
+++ b/deepeval/metrics/exact_match/exact_match.py
@@ -45,8 +45,8 @@ class ExactMatchMetric(BaseMetric):
         with metric_progress_indicator(
             self, _show_indicator=_show_indicator, _in_component=_in_component
         ):
-            expected = test_case.expected_output.strip()
-            actual = test_case.actual_output.strip()
+            expected = test_case.expected_output
+            actual = test_case.actual_output
 
             if expected == actual:
                 self.score = self.precision = self.recall = self.f1 = 1.0
--- a/deepeval/metrics/pattern_match/pattern_match.py
+++ b/deepeval/metrics/pattern_match/pattern_match.py
@@ -24,7 +24,7 @@ class PatternMatchMetric(BaseMetric):
         verbose_mode: bool = False,
         flaky: bool = False,
     ):
-        self.pattern = pattern.strip()
+        self.pattern = pattern
         self.ignore_case = ignore_case
         self.verbose_mode = verbose_mode
         self.flaky = flaky
@@ -55,7 +55,7 @@ class PatternMatchMetric(BaseMetric):
         with metric_progress_indicator(
             self, _show_indicator=_show_indicator, _in_component=_in_component
         ):
-            actual = test_case.actual_output.strip()
+            actual = test_case.actual_output
             full_match = self._compiled_pattern.fullmatch(actual)
 
             self.score = 1.0 if full_match else 0.0
--- a/docs/content/docs/(non-llm)/metrics-pattern-match.mdx
+++ b/docs/content/docs/(non-llm)/metrics-pattern-match.mdx
@@ -88,7 +88,7 @@ await evaluate([testCase], [metric]);
 There is **ONE** mandatory and **FOUR** optional parameters when creating a `PatternMatchMetric`:
 
 - `pattern`: a string representing the regular expression pattern that the <Term py="actual_output" ts="actualOutput"/> must match.
-- [Optional] <Term py="ignore_case" ts="ignoreCase"/>: a boolean which when set to <Term py="True" ts="true"/>, performs case-sensitive pattern matching. Defaulted to <Term py="False" ts="false"/>.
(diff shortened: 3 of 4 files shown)


CTWalk:
Quick note on the red `lint` check. It looks pre-existing rather than something this PR introduced.

The [job](https://github.com/confident-ai/deepeval/actions/runs/30142306361/job/89637818009) reports 50 files needing reformat, and none of them are the 5 this PR touches; both changed test files show up in the log as "already well formatted, good job". The same 50-file result reproduces locally on unmodified `main`, and `lint` is currently red on #2942, #2930 (the present `main` HEAD), and #2926 as well, so it reads like drift between the committed formatting and `psf/black@stable`.

Happy to add a `black .` commit if you'd like that cleaned up, though I left it out deliberately so the diff stays at the 5 files under review. The Vercel check is just a fork not being authorized to deploy the docs preview.

The `test` jobs and the integration jobs are green.

CTWalk:
Rebased onto current `main` (`0d100e37d`) and force-pushed, the branch was conflicting against the `flaky` flag and docs changes that landed after it was opened.

I also corrected a scope mistake in the original push. That version rewrote `tests/test_metrics/test_exact_match_metric.py` and `test_pattern_match_metric.py`, removing about 125 lines of existing tests, which is not something a whitespace fix should be doing. Both files are now **untouched**, and the new cases live in a separate `tests/test_metrics/test_match_whitespace.py`. I put them in a new file rather than the existing ones because those modules carry a top-level `OPENAI_API_KEY` skipif that would have skipped tests which need no key.

To confirm the removal was never necessary: the four pre-existing tests pass unchanged against this change. I ran them with a key set so the skipif didn't hide the result.

The diff is now 4 source lines, 1 docs line, and one new test file with 65 insertions, 5 deletions.

My earlier note about `lint` still applies, just with "4 files" rather than 5: the job's 50-file reformat list reproduces on unmodified `main` and doesn't include anything this PR touches. `black --check` is clean on all 4 changed files.

PR #2963, opened 2026-07-27
fix: correct local embedding API key enum value

### LINKED ISSUE #2904: LOCAL_EMBEDDING_API_KEY enum value is defined as a tuple instead of a string
### Summary

`LOCAL_EMBEDDING_API_KEY` is declared as

```python
("LOCAL_EMBEDDING_API_KEY",)
```

instead of

```python
"LOCAL_EMBEDDING_API_KEY"
```

### Impact

This causes incorrect JSON serialization and lookup failures.

### Expected Behavior

Every enum member should contain a string.

### Proposed Fix

Replace

```python
("LOCAL_EMBEDDING_API_KEY",)
```

with

```python
"LOCAL_EMBEDDING_API_KEY"
```

### Files

```
deepeval/key_handler.py
```

jatinder14:
## Summary
Corrects `LOCAL_EMBEDDING_API_KEY` so legacy JSON configuration can map it to the settings field.

| File | Change |
|------|--------|
| `deepeval/key_handler.py` | Replaces the singleton tuple enum value with a string. |
| `tests/test_core/test_config/test_settings.py` | Adds legacy-keyfile regression coverage for the local embedding key. |

Closes #2904

## Tests
- `.venv/bin/python -m pytest tests/test_core/test_config/test_settings.py -q` (54 passed)
- `.venv/bin/python -m black --check deepeval/key_handler.py tests/test_core/test_config/test_settings.py`

## Rollback
Revert commit `a87f37858`.


jatinder-mahajan-certifyos:
@penguine-ip would you mind taking a look, or pointing me at whoever owns the key handling?

`EmbeddingKeyValues.LOCAL_EMBEDDING_API_KEY` has a trailing comma, so its value is the tuple `("LOCAL_EMBEDDING_API_KEY",)` rather than the string every lookup expects. One-character fix plus a regression test. Fixes #2904.

I rebuilt the branch onto current `main` earlier today, so it's now 2 files and no conflicts — an earlier merge had inflated the diff. There's a sibling fix in #2964 (unknown model cost in the synthesizer) if it's easier to look at both together.

PR #2964, opened 2026-07-27
fix(synthesizer): handle unknown model costs

### LINKED ISSUE #2884: Synthesizer silently produces empty goldens with unregistered models (e.g. deepseek-v4-flash/v4-pro)

jatinder14:
## Summary
- avoid adding `None` when a native model has no registered pricing
- preserve context scoring and golden generation for unknown-cost models
- add synchronous and asynchronous regressions for the unknown-cost path

Fixes #2884

## Test plan
- [x] `uv run --no-project ... pytest tests/test_core/test_synthesizer/test_context_generator.py -q` (6 passed, 7 skipped)
- [x] `python3 -m compileall -q deepeval/synthesizer/chunking/context_generator.py tests/test_core/test_synthesizer/test_context_generator.py`

### DIFF
--- a/deepeval/synthesizer/chunking/context_generator.py
+++ b/deepeval/synthesizer/chunking/context_generator.py
@@ -837,7 +837,8 @@ class ContextGenerator:
         prompt = FilterTemplate.evaluate_context(chunk)
         if self.using_native_model:
             res, cost = self.model.generate(prompt, schema=ContextScore)
-            self.total_cost += cost
+            if cost is not None:
+                self.total_cost += cost
             return (res.clarity + res.depth + res.structure + res.relevance) / 4
         else:
             try:
@@ -862,7 +863,8 @@ class ContextGenerator:
         prompt = FilterTemplate.evaluate_context(chunk)
         if self.using_native_model:
             res, cost = await self.model.a_generate(prompt, schema=ContextScore)
-            self.total_cost += cost
+            if cost is not None:
+                self.total_cost += cost
             return (res.clarity + res.depth + res.structure + res.relevance) / 4
         else:
 
--- a/tests/test_core/test_synthesizer/test_context_generator.py
+++ b/tests/test_core/test_synthesizer/test_context_generator.py
@@ -4,7 +4,10 @@ import os
 from itertools import chain
 from types import SimpleNamespace
 
-from deepeval.synthesizer.chunking.context_generator import ContextGenerator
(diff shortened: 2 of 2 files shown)


jatinder-mahajan-certifyos:
@kritinv flagging this one to you since you've done recent work in the synthesizer.

`ContextGenerator.evaluate_chunk` / `a_evaluate_chunk` do `self.total_cost += cost`, but a native model can return `None` for cost, which raises `TypeError`. This guards the accumulation. Fixes #2884.

Rebuilt onto current `main` earlier today — 2 files, no conflicts (an earlier merge had ballooned the diff, and it also carried a stray `status.md` which is now gone). Sibling fix in #2963 if useful.

PR #2965, opened 2026-07-27
fix(metrics): normalize ConversationalGEval score against the rubric range

arthi-arumugam-git:
### CI note

Regenerating the templates bundle touches `typescript/src/templates/metrics/templates.json`, which triggers the TypeScript Lint / TypeScript Tests workflows. Those fail at `npm ci` with `Missing: ai@7.0.37 from lock file`: `typescript/package-lock.json` is out of sync with `package.json` on `main`, and both workflows already fail on `main` for the same reason (TypeScript Lint run 30238259793, TypeScript Tests run 30238259784, both on `main` before this branch). Nothing in this PR touches the TypeScript sources or lockfile. Happy to drop the TS bundle from the diff if you'd rather regenerate it separately.


arthi-arumugam-git:
The checks on this are red, but none of the red is coming from this change, and it seemed more useful to say why than to leave it looking broken.

**lint.** Prettier fails on `typescript/src/evaluate/evaluate.ts`. This branch does not touch that file, and the blob is the same one that is on main right now:

```
gh api repos/confident-ai/deepeval/contents/typescript/src/evaluate/evaluate.ts --jq .sha
d2caefdd3f7f14a3a56baa0a6dcd8052549ee762
```

The copy on this branch has that identical sha, so the formatting issue is already on main and any PR opened today will inherit it.

**Vercel.** Reports "Authorization required to deploy", which is the fork permission prompt rather than a build failure.

**openai, openai-agents, langchain, mastra, ai-sdk, e2e.** These fail inside `node_modules/openai` with API status errors raised through `src/openai/patch.ts`, so they are making live API calls. A pull request from a fork does not receive the repository secrets, so those jobs cannot pass from here whatever the diff contains.

The change itself is Python, plus a test at `tests/test_metrics/test_conversational_g_eval_rubric_score_range.py` that fails on main and passes with the fix, so it can be checked without any of the above running.

Happy to rebase onto current main if that would make the signal cleaner, and equally happy to leave it alone if the CI state is already known.

arthi-arumugam-git:
The red checks here do not look like they come from this change. Two separate causes, and both hit any fork PR that touches the `typescript/` tree.

Every Python check on this PR passes: `lint (ubuntu-latest)`, all six `test` jobs, `openai`, `openai-agents`, `langchain`, `agentcore`, `crewai`, `llama-index`, `openinference`, `pydantic-ai`, `strands`, `pr-title-check`. The failures are all on the TypeScript side. This is also the only one of my four open PRs that runs those jobs at all, because it is the only one touching `typescript/`. #2966, #2967 and #2968 fail nothing except `Vercel`.

**`lint`.** It fails on prettier, and the file it names is `src/evaluate/evaluate.ts`. This PR changes no `.ts` file, only `typescript/src/templates/metrics/templates.json`, and the prettier step checks `src/**/*.ts` and `test/**/*.ts`, so a JSON file cannot be what it is complaining about. The same job also fails on `main`: runs 32944965842 (26 Aug), 32781935235 and 32779740828 (24 Aug), same "Code style issues found" message, 2 files there against 1 here.

**`ai-sdk`, `e2e`, `langchain`, `mastra`, `openai`, `openai-agents`.** The provider keys are empty. The `openai` job log has `OPENAI_API_KEY:` with nothing after it, and then all 11 tests fail with `401 You didn't provide an API key`. Fork pull requests do not get repository secrets, so I do not think I can turn those green from my side.

`Vercel` says "Authorization required to deploy", and that one is failing on all four of my PRs including the three with no TypeScript changes.

On the TS file itself: I changed it only to keep the mirrored template in step with the Python one, since both interpolate the same rubric range. If you would rather this PR did not pull in the TS suite, I can drop that file and send it separately, or leave the two templates out of sync for now. Whichever is easier for you.

PR #2966, opened 2026-07-27
fix(benchmarks): correct the accuracy denominators in EquityMedQA and GSM8K

arthi-arumugam-git:
Two benchmarks report an accuracy divided by a denominator that does not match the goldens they actually scored. Both fail silently -- no error, just a wrong number.

## EquityMedQA

`EquityMedQA.evaluate()` scores only the first 10 goldens of each task:

```python
for golden in tqdm(goldens[:10], desc=f"Processing {task.value}"):
```

but sets the denominator from the full task:

```python
task_total_predictions = len(goldens)
overall_total_predictions += len(goldens)
```

So the reported accuracy is scaled by `10 / len(goldens)`. A model that answers every scored golden correctly on a 40-golden task reports 0.25. On `fbrt_llm_661_sampled` (661 goldens) a perfect run reports 0.015. Only tasks with 10 or fewer goldens are correct today.

The `[:10]` cap was also hard-coded, so there was no way to run the full benchmark.

Fix: divide by the goldens actually scored, and expose the cap as `n_problems_per_task`, the same argument BBQ, MMLU, HellaSwag, DROP, SQuAD, LogiQA, MathQA and TruthfulQA already take. The default is 10, so an existing run scores exactly the same goldens as before and only the denominator changes. `n_problems_per_task=None` scores the whole task.

## GSM8K

`GSM8K.evaluate()` divides by the requested `n_problems` rather than the goldens actually loaded:

```python
overall_total_predictions = self.n_problems
goldens = self.load_benchmark_dataset()[: self.n_problems]
```

and unlike BoolQ, LAMBADA, Winogrande and ARC it has no bound on `n_problems`. `GSM8K(n_problems=5000)` is accepted, runs the 1319 problems that exist, and deflates the accuracy by `1319/5000` -- a perfect model reports 0.264.

Fix: add the bound assert the sibling benchmarks all have, and use `len(goldens)` as the denominator. With `n_problems <= 1319` the two are equal, so runs that were already correct are unaffected.

## Tests

`tests/test_benchmarks/test_benchmark_accuracy_denominator.py`. The dataset and the predictions are scripted fakes, so the tests need no model, network or API key.

On main, 5 of the 8 fail:

```
$ python -m pytest tests/test_benchmarks/test_benchmark_accuracy_denominator.py -q

___________ test_equity_med_qa_accuracy_is_over_the_goldens_actually_scored ____________

    benchmark = _FakeEquityMedQA(n_goldens=40)
[...]

PR #2967, opened 2026-07-27
fix(benchmarks): stop HumanEval from collapsing pass@k into pass@n

arthi-arumugam-git:
`HumanEval.evaluate()` folds the pass@k score through a truthiness check:

```python
prediction, score = self.predict(model, task, golden, k).values()
if score:
    task_correct = 1
    overall_correct_predictions += 1
```

pass@k is a probability in `[0, 1]`, not a pass/fail flag. `Scorer.pass_at_k(n, c, k)` returns a non-zero value whenever `c >= 1`, i.e. whenever at least one of the `n` samples passes. So `if score:` counts every such task as a full pass, which is pass@n -- regardless of the `k` that was requested. The `k` argument to `evaluate()` has no effect on the reported number.

With the default `n=200` and `k=1`, a task where 1 of 200 samples passes has `pass@1 = 0.005` but is reported as `1.0`. A model that happens to solve each task once in 200 attempts scores 100% on HumanEval.

Fix: accumulate the score itself. `overall_accuracy` becomes the mean pass@k over the tasks and the per-task score becomes that task's pass@k, which is what the standard metric is.

This changes the reported number for `k < n`, which is the case that was wrong. For `k == n`, `pass_at_k` returns exactly 1.0 for any task with a passing sample and 0.0 otherwise, so runs at `k == n` report the same number as before.

## Tests

`tests/test_benchmarks/test_human_eval_pass_at_k.py`. `evaluate()` is driven with scripted `(n, c)` pairs, so no model, network or API key is needed. The last test pins the `k == n` case that must not change.

On main, 3 of the 7 fail:

```
$ python -m pytest tests/test_benchmarks/test_human_eval_pass_at_k.py -q

_______ test_pass_at_1_of_one_correct_sample_in_two_hundred_is_not_a_full_pass _______

    benchmark = _FakeHumanEval(n=200, c=1)
    result = benchmark.evaluate(model=None, k=1)

>   assert result.overall_accuracy == pytest.approx(0.005)
E   assert 1.0 == 0.005 +- 5.0e-09
E     comparison failed
E     Obtained: 1.0
E     Expected: 0.005 +- 5.0e-09

________________ test_overall_accuracy_is_the_mean_pass_at_k[2-1-0.02] ________________

c = 2, k = 1, expected = 0.02

    benchmark = _FakeHumanEval(n=100, c=c, n_tasks=4)
    result = benchmark.evaluate(model=None, k=k)

>   assert result.overall_accuracy == pytest.approx(expected)
E   assert 1.0 == 0.02 +- 2.0e-08
E     comparison failed
E     Obtained: 1.0
E     Expected: 0.02 +- 2.0e-08

________________ test_overall_accuracy_is_the_mean_pass_at_k[10-1-0.1] ________________

c = 10, k = 1, expected = 0.1

    benchmark = _FakeHumanEval(n=100, c=c, n_tasks=4)
    result = benchmark.evaluate(model=None, k=k)

>   assert result.overall_accuracy == pytest.approx(expected)
E   assert 1.0 == 0.1 +- 1.0e-07
E     comparison failed
E     Obtained: 1.0
E     Expected: 0.1 +- 1.0e-07

3 failed, 4 passed in 12.21s
```

With the fix, `tests/test_benchmarks/` is 20 passed. `black` was run over the changed files.

Related but independent of #2965 (`ConversationalGEval` rubric range) and #2966 (`EquityMedQA`/`GSM8K` accuracy denominators). No shared code or branches.

### DIFF
--- a/deepeval/benchmarks/human_eval/human_eval.py
+++ b/deepeval/benchmarks/human_eval/human_eval.py
@@ -122,22 +122,20 @@ class HumanEval(DeepEvalBaseBenchmark):
 
             for task in self.tasks:
                 golden: Golden = self.load_benchmark_dataset(task)
-                task_correct = 0
                 overall_total_predictions += 1
 
-                # Calculate task accuracy
+                # Calculate task accuracy. pass@k is a probability in [0, 1],
+                # not a pass/fail flag, so it is accumulated as-is.
                 prediction, score = self.predict(
                     model, task, golden, k
                 ).values()
-                if score:
-                    task_correct = 1
-                    overall_correct_predictions += 1
+                overall_correct_predictions += score
                 predictions_row.append(
                     (
                         task.value,
                         golden.input,
                         prediction,
-                        task_correct,
+                        score,
                         golden.expected_output,
                         score,
                     )
@@ -146,10 +144,8 @@ class HumanEval(DeepEvalBaseBenchmark):
                     self.print_verbose_logs(
                         task.value, golden.input, prediction, score
                     )
-                print(
-                    f"HumanEval Task Accuracy (task={task.value}): {task_correct}"
-                )
-                scores_row.append((task.value, task_correct))
+                print(f"HumanEval Task Accuracy (task={task.value}): {score}")
+                scores_row.append((task.value, score))
 
             # Calculate overall accuracy
             overall_accuracy = (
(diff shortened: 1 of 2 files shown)


arthi-arumugam-git:
Gentle ping. This one is small and still merges cleanly.

The bug in one line: HumanEval's pass@k is computed over all n samples rather than the k requested, so a benchmark run reports pass@n while labelling it pass@k. The number comes out too high and nothing raises, which is the kind of defect that quietly inflates a published comparison.

The Vercel check failure is a preview deploy unrelated to this change. Happy to rebase if it has drifted.

Issue #2906, opened 2026-07-16
Windows-specific test failures due to path expansion and encoding assumptions


sahniaditya007:
### Summary

Two tests fail on Windows because of platform-specific assumptions.

### Problems

1.

`expanduser()` relies on `USERPROFILE` on Windows.

Only `HOME` is mocked.

2.

HTML export tests read UTF-8 files without specifying encoding.

This can raise

```
UnicodeDecodeError
```

on Windows.

### Suggested Fixes

Mock both

```
HOME
USERPROFILE
```

and use

```python
read_text(encoding="utf-8")
```

### Files

```
tests/test_core/test_config/test_settings.py

tests/test_core/test_evaluation/test_console_report.py
```

PR #2976, opened 2026-07-27
fix: use UTF-8 for prompt template and .env file IO

### LINKED ISSUE #2975: Prompt.load and the CLI dotenv writer use the Windows ANSI codepage, corrupting non-ASCII content
Two file IO paths use the platform default encoding, which on Windows is the ANSI codepage (cp1252 on Western installs, cp932/936/949 on Japanese/Chinese/Korean ones), never UTF-8. Both are reachable from documented entry points. Neither can show up on your CI since ubuntu defaults to UTF-8.

**1. `Prompt.load()` silently corrupts text templates**

`deepeval/prompt/prompt.py` opens template files with `open(file_path, "r")`, no encoding. Loading a UTF-8 `.txt` template containing

```
Don’t guess — answer in 中文 (你好).
```

on cp1252 Windows raises nothing and produces

```
Donâ€™t guess â€” answer in ä¸­æ–‡ (ä½ å¥½).
```

which is then stored on `text_template` and sent to the model. Curly quotes and em dashes are what Word, Google Docs and most chat apps insert by default, so ordinary English templates hit this too, not just non-English ones.

JSON templates fail louder, since the mangled bytes usually break decoding first:

```
>>> prompt.load("template.json")
UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d in position 141
```

**2. The CLI dotenv writer and the dotenv reader disagree about encoding**

`deepeval/cli/dotenv_handler.py` reads and writes `.env` files with `Path.read_text()` / `write_text()`, no encoding, so ANSI. But the read path for the same files (`deepeval/config/utils.py:150`) goes through python-dotenv, which defaults to UTF-8. So on Windows, `deepeval set-* --save` writes a file the rest of deepeval cannot reliably read:

```
> deepeval set-azure-openai --deployment "Ángel-gpt4" ... --save dotenv
# .env.local now contains 0xC1 (cp1252) where UTF-8 is expected
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xc1 in position 17
```

And because `upsert()` rewrites the entire file, any non-ASCII the user already keeps in a shared `.env.local` (values for their own app, comments) is also at risk: read as ANSI, written back as ANSI, or a hard UnicodeEncodeError on codepages that cannot represent the characters at all.

Repros above are from cp1252 Windows, current main. The fix is `encoding="utf-8"` on the one `open()` and the four `read_text`/`write_text` calls; PR with regression tests is up alongside this issue.

hakoke:
Fixes #2975.

`Prompt.load()` opened template files, and `DotenvHandler` read and wrote `.env` files, with no encoding argument, so on Windows both used the ANSI codepage instead of UTF-8. A UTF-8 text template with curly quotes loads as mojibake with no error and gets sent to the model; JSON templates with CJK content raise UnicodeDecodeError; and the dotenv writer produced ANSI files that the UTF-8 read path (`deepeval/config/utils.py:150`, python-dotenv) then fails to parse back, including values it never touched, since `upsert()` rewrites the whole file.

The change is `encoding="utf-8"` on the one `open()` in `Prompt.load` and the four `read_text`/`write_text` calls in `DotenvHandler`. The prompt cache files were left alone on purpose: they go through `json.dump` with `ensure_ascii=True`, so they are ASCII by construction and safe under any codepage.

Tests: three new loads in `test_load.py` covering the silent-mojibake and UnicodeDecodeError cases, and a new `test_dotenv_handler.py` covering upsert/unset round-trips with non-ASCII values next to untouched keys. All seven fail on the unfixed code on cp1252 Windows and pass with the fix. `tests/test_core` minus the CI-ignored dirs: 1174 passed, with 2 pre-existing Windows failures unrelated to this change (one is the expanduser test from #2906). `black --check` clean at line length 80. No behavior change on platforms whose default encoding is already UTF-8.

### DIFF
--- a/deepeval/cli/dotenv_handler.py
+++ b/deepeval/cli/dotenv_handler.py
@@ -25,7 +25,11 @@ class DotenvHandler:
         Idempotently set/replace keys in a dotenv file. Preserves comments/order.
         Creates file if missing. Sets file mode to 0600 when possible.
         """
-        lines = self.path.read_text().splitlines() if self.path.exists() else []
+        lines = (
+            self.path.read_text(encoding="utf-8").splitlines()
+            if self.path.exists()
+            else []
+        )
         seen = set()
 
         # replace existing keys in-place
@@ -50,7 +54,9 @@ class DotenvHandler:
             lines.extend(to_append)
 
         self.path.parent.mkdir(parents=True, exist_ok=True)
-        self.path.write_text("\n".join(lines) + ("\n" if lines else ""))
+        self.path.write_text(
+            "\n".join(lines) + ("\n" if lines else ""), encoding="utf-8"
+        )
         try:
             self.path.chmod(stat.S_IRUSR | stat.S_IWUSR)  # 0600
         except Exception:
@@ -60,7 +66,7 @@ class DotenvHandler:
         """Remove keys from dotenv file, but leave comments and other lines untouched."""
         if not self.path.exists():
             return
-        lines = self.path.read_text().splitlines()
+        lines = self.path.read_text(encoding="utf-8").splitlines()
         out = []
         keys = set(keys)
         for line in lines:
@@ -68,4 +74,6 @@ class DotenvHandler:
             if m and m.group(1) in keys:
                 continue
             out.append(line)
-        self.path.write_text("\n".join(out) + ("\n" if out else ""))
+        self.path.write_text(
+            "\n".join(out) + ("\n" if out else ""), encoding="utf-8"
+        )
--- a/deepeval/prompt/prompt.py
+++ b/deepeval/prompt/prompt.py
@@ -202,7 +202,7 @@ class Prompt:
 
         file_name = os.path.basename(file_path).split(".")[0]
         self.alias = file_name
-        with open(file_path, "r") as f:
+        with open(file_path, "r", encoding="utf-8") as f:
             content = f.read()
         try:
             data = json.loads(content)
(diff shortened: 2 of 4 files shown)


spacesheepinternet:
Verified on Windows 11, Python 3.12.7, `locale.getpreferredencoding(False)` = `cp1252`.

I reproduced both paths independently before applying the patch:

- `Prompt.load()` on a UTF-8 `.txt` template containing `Don't guess — answer in 中文 (你好).` returns `Donâ€™t guess â€" answer in ä¸­æ–‡ (ä½ å¥½).` and stores it on `text_template`. The `.json` path corrupted silently for my fixture rather than raising, so the failure isn't always loud.
- `DotenvHandler.upsert({"AZURE_DEPLOYMENT": "Ángel-gpt4"})` writes `b'AZURE_DEPLOYMENT=\xc1ngel-gpt4'`, and `dotenv_values()` on that file raises `UnicodeDecodeError: 'utf-8' codec can't decode byte 0xc1 in position 17` — the same byte and position as the report.
- An existing `.env.local` containing Japanese makes `upsert()` raise `UnicodeDecodeError: 'charmap' codec can't decode byte 0x81` before it writes anything.

With the patch applied, all three behave correctly.

On the tests in this PR: **18 pass** on the branch. Reverting only `dotenv_handler.py` and `prompt.py` to `main` while keeping the new tests, **7 of them fail** — so they're genuine regression tests, not vacuous ones.

One nuance worth recording: in the dotenv path, a value containing only curly quotes or em dashes round-trips intact on cp1252, because those byte values map back through the codepage. It's the characters cp1252 cannot represent (CJK, `Á`) that break it. `Prompt.load` corrupts curly quotes regardless.

PR #2991, opened 2026-07-31
fix: strip <think> tags in trimAndLoadJson for reasoning models

chenrocky:
## Summary

Reasoning models (DeepSeek-R1, QwQ, Nemotron, MiniMax-M2, etc.) emit `<think>...</think>` blocks that can contain braces, causing `trimAndLoadJson` / `trim_and_load_json` to extract the wrong JSON substring and fail with "Evaluation LLM outputted an invalid JSON."

This is an alternative implementation to #2410 that addresses all 4 points from the reviewer's feedback on that PR.

## Changes

- Adds `<think>` tag stripping as a **fallback** in all three copies of the JSON parsing function — only runs after direct `json.loads` fails, so valid JSON with literal `<think>` in string values is never corrupted
- Handles the **closing-tag-only** variant (`</think>` without opening tag), which occurs when the chat template injects the opening `<think>`
- Fixes all three copies: `deepeval/metrics/utils.py`, `deepeval/models/llms/utils.py`, `deepeval/dataset/utils.py`
- Also upgrades `models/llms/utils.py` to the two-phase "parse first, clean up on failure" approach from #2701, which was missed in that PR

## Tests

- 27/27 tests pass in `tests/test_core/test_trim_and_load_json.py`
- Added 5 new test cases: think block with braces, literal `<think>` in JSON strings preserved, closing-tag-only variant, brace-free think block, nested JSON in think block
- Extended test coverage to include `models/llms/utils.py` (previously untested)
- All 3 function copies tested via parametrize

## Context

- Related: #2410 (original PR, reviewer feedback addressed here)
- The existing `find("{")` / `rfind("}")` extraction already handles the common case where a brace-free think block precedes the JSON. This fix adds value when the think block itself contains braces.
- Verified against real evaluation workloads with DeepSeek-R1 and MiniMax-M2 as judge models — reduced GEval errors from 9 to 0 for DeepSeek-R1.

### DIFF
--- a/deepeval/dataset/utils.py
+++ b/deepeval/dataset/utils.py
@@ -202,9 +202,7 @@ def trimAndLoadJson(input_string: str) -> Any:
                 return json.loads(jsonStr2)
             except json.JSONDecodeError:
                 try:
-                    return json.loads(
-                        re.sub(r",\s*([\]}])", r"\1", jsonStr2)
-                    )
+                    return json.loads(re.sub(r",\s*([\]}])", r"\1", jsonStr2))
                 except json.JSONDecodeError:
                     pass
 
--- a/deepeval/dataset/utils.py
+++ b/deepeval/dataset/utils.py
@@ -160,20 +160,55 @@ def convert_convo_goldens_to_convo_test_cases(
     return test_cases
 
 
+def _strip_think_tags(text: str) -> str:
+    """Strip reasoning/thinking blocks emitted by models like DeepSeek-R1, QwQ,
+    and Nemotron.  Handles both fully-paired ``<think>...</think>`` and the
+    closing-tag-only variant where the chat template injects the opening tag so
+    the model only emits ``</think>``."""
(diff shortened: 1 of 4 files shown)


chenrocky:
Hi @penguine-ip — this PR is an alternative implementation to #2410, addressing all 4 points from your review there:

1. Rebased on latest main (post-#2701 two-phase parse approach)
2. `<think>` stripping only in the fallback path — valid JSON with literal `<think>` in string values is never corrupted
3. Handles the closing-tag-only variant (`</think>` without opening tag)
4. Tests added covering: think block with braces, literal `<think>` in strings preserved, closing-tag-only, and nested JSON in think block

Also fixes `models/llms/utils.py` which was missed by #2701 (still had the old unconditional-regex approach).

The lint CI failure is pre-existing (`llm_test_case.py` formatting) — not from this PR.

chenrocky:
@penguine-ip friendly ping — just checking if you've had a chance to look at this. Happy to address any feedback.

PR #2995, opened 2026-08-02
build(typescript): unbreak npm ci by aligning the ai devDependency with the mastra peer range

arthi-arumugam-git:
`npm ci` has failed on every branch of this repo, including `main`, since at least 21 July. It takes both **TypeScript Lint** and **TypeScript Tests** down at the install step, so neither has run to completion on any PR in that window.

```
npm error code EUSAGE
npm error `npm ci` can only install packages when your package.json and
npm error package-lock.json ... are in sync.
npm error Missing: ai@7.0.48 from lock file
npm error Missing: @ai-sdk/gateway@4.0.37 from lock file
npm error Missing: @ai-sdk/provider@4.0.4 from lock file
```

## The version number is the clue

CI asked for `ai@7.0.37` on 27 July and asks for `7.0.48` today. The required version tracks whatever npm published most recently, which means something is re-resolving at install time rather than reading the lock.

## Root cause

`@mastra/core` nests a package `chat` that declares:

```json
"peerDependencies": { "ai": "^6.0.182 || ^7.0.0" }
```

The lockfile pins `ai` at **6.0.97**. That satisfies neither branch of the range: it is below `^6.0.182` and it is not `^7`. So npm resolves the peer by reaching for the newest `ai@7`, then fails because that version is not in the lock.

It reproduces only on **npm 10**, which is what `node:22` ships and what these workflows run in. npm 11 changed peer resolution and prefers the `^6.0.97` devDependency, so this passes on a modern local machine and fails in CI. That gap is why it has gone unnoticed.

## The fix

Bump the `ai` devDependency from `^6.0.97` to `^6.0.182`, so the root `ai` falls inside the range `chat` asks for and one copy satisfies the whole tree. The lock regenerates to `6.0.240`.

I also tried simply bounding the `ai` peer range to `<7.0.0`. That does **not** work, because the conflict is with a nested peer requirement, not the root one. Worth recording so nobody repeats it.

## Verification

Run with npm 10 to match the container:

| | before | after |
|---|---|---|
| `npx npm@10 ci` | `EUSAGE`, missing `ai@7.0.48` | **exit 0**, added 978 packages |
| `npm run lint -- .` | never reached | **exit 0**, 0 errors, 2 pre-existing warnings |

## What this does not fix

This unblocks the install step only. Being explicit so the result is not oversold:

- **TypeScript Lint** now gets past `npm ci` and reaches the prettier step, where `npx prettier --check "src/**/*.ts" "test/**/*.ts"` reports **66 files** with style issues. ESLint itself passes (0 errors, 2 pre-existing warnings). That drift is untouched by this PR, which changes only `package.json` and `package-lock.json`, neither of which prettier checks. Happy to send it as a separate mechanical `npm run format` PR if you want it.
- **TypeScript Tests** needs `OPENAI_API_KEY` and `CONFIDENT_API_KEY`, which GitHub does not pass to forked PRs, so those jobs will still not go green from a fork regardless.

## Related, if useful

Four peer ranges in `typescript/package.json` have no upper bound: `ai`, `@google/genai`, `@anthropic-ai/sdk`, `@aws-sdk/client-bedrock-runtime` and `ollama`. The `@mastra/*` entries next to them are bounded (`>=1.16.0-0 <2.0.0-0`), which is the shape that avoids this class of problem. `@google/genai` looks like the next candidate: its peer is `>=1.0.0` while the devDependency is `^2.8.0`. Not touching those here, since this PR should stay minimal.

### DIFF
--- a/typescript/package-lock.json
+++ b/typescript/package-lock.json
@@ -52,7 +52,7 @@
         "@types/nunjucks": "^3.2.6",
         "@types/papaparse": "^5.3.16",
         "@types/uuid": "^9.0.0",
-        "ai": "^6.0.97",
+        "ai": "^6.0.182",
         "babel-jest": "^29.7.0",
         "eslint": "^9.38.0",
         "eslint-config-prettier": "^10.1.8",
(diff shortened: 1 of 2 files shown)


arthi-arumugam-git:
Quick note on the red checks here, because they are misleading at a glance and I would rather explain them than have someone assume the worst.

**The three failures are not caused by this change, and two of them are the change working.**

Before this PR, `lint`, `test` and `test-mastra` all died at `npm ci` with `EUSAGE`, so nothing downstream ever ran. They now get all the way through install and fail later, for two separate reasons:

**`test` and `test-mastra`** fail on missing secrets: `Confident AI API key not found`, and `Could not find API key process.env.OPENAI_API_KEY for model id openai/gpt-4o-mini`. This PR comes from a fork, and fork PRs do not get repository secrets. Nothing I can do about that from here, and it is not specific to this change.

**`lint`** fails on this:

```
typescript/src/tracing/tracing.ts
  704:15  warning  'confidentApiKey' is assigned a value but never used.
                   Allowed unused vars must match /^_/u  @typescript-eslint/no-unused-vars

2 problems (0 errors, 2 warnings)
```

Zero errors, two warnings, and the job exits 1 because warnings are treated as failures. `tracing.ts` is not a file this PR touches: the diff is `typescript/package.json` and `typescript/package-lock.json` only. The warning was already there. It just could not be reported while `npm ci` was failing first, so this is pre-existing lint debt becoming visible rather than anything new.

Happy to rename it to `_confidentApiKey` in a separate commit here if you want `lint` green in one go, but I did not want to quietly widen a dependency PR into a source change without asking.

The install fix itself is verified in your own CI on this branch: `npm ci` reports `added 978 packages in 19s`, where every branch since 21 July errored out.

arthi-arumugam-git:
Closing this as obsolete. npm ci was genuinely failing on main when I opened this, at the install step across all three jobs. It installs cleanly on main now, so this no longer fixes anything and the lockfile here would only add churn.

Worth noting for anyone who lands here from a red CI run: the lint job still fails on main, but the failing step is Run Prettier (check) rather than install, which is a different problem from the one this PR was about.

PR #2997, opened 2026-08-02
feat: add DEEPEVAL_DISABLE_PROMOTION to silence Confident AI promo output

### LINKED ISSUE #1417: Don't push for confidently.ai platform, or give way to silence it
**Describe the bug**
Running evals without doing a `deepeval login` results a console message being printed. 
https://github.com/confident-ai/deepeval/blob/main/deepeval/test_run/test_run.py#L744

I would like to make sure this library's stance is to be less opinionated about cloud platforms. Those of us at large companies should be not be shipping data outside our compute boundaries. 

**To Reproduce**
1. Run any eval without logging in

**Expected behavior**
If toggled, no references to confidently.ai, respecting local contexts better

namankhurpia:
Adds an opt-out env var so evals run without 'deepeval login' can suppress the Confident AI promotional console output (post-eval nudge, arena nudge, and the failed-case note row). Defaults to showing, preserving existing behavior. Closes #1417.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -88,6 +88,9 @@ CONFIDENT_API_KEY=  # Confident AI cloud
 # DEEPEVAL_TELEMETRY_OPT_OUT=1
 # ERROR_REPORTING=1
 
+# Suppress Confident AI promotional console output after evals (when not logged in)
+# DEEPEVAL_DISABLE_PROMOTION=1
+
 # ===================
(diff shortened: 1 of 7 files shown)


namankhurpia:
@penguine-ip  why was this closed ? 

please close the issue as well #1417

PR #2999, opened 2026-08-03
fix(optimizer): give each PromptOptimizer its own algorithm instance

### LINKED ISSUE #2998: PromptOptimizer's default GEPA() is a shared instance - a second optimizer silently hijacks the first's model_callback and metrics
`PromptOptimizer.__init__` uses a mutable default argument for `algorithm` (`deepeval/optimizer/prompt_optimizer.py:57`):

```python
algorithm: Union[GEPA, MIPROV2, COPRO, SIMBA] = GEPA(),
```

`GEPA()` is evaluated once at import, so every `PromptOptimizer` created without an explicit `algorithm=` shares a single algorithm object. `_configure_algorithm()` (lines 126-152) then mutates that shared object, assigning `scorer`, `_rewriter`, `status_callback`, `step_callback` and `optimizer_model` onto it.

Constructing a second optimizer therefore silently rewires the first one.

### Repro (no API key needed)

```python
from deepeval.metrics import BaseMetric
from deepeval.optimizer.prompt_optimizer import PromptOptimizer


class StubLLM(DeepEvalBaseLLM):
    def load_model(self): return None
    def generate(self, *a, **k): return ""
    async def a_generate(self, *a, **k): return ""
    def get_model_name(self): return "stub"


class StubMetric(BaseMetric):
    def measure(self, tc, *a, **k): self.score = 1.0; return self.score
    async def a_measure(self, tc, *a, **k): return self.measure(tc)
    def is_successful(self): return True


def cb1(prompt, golden): return "from-callback-1"
def cb2(prompt, golden): return "from-callback-2"

o1 = PromptOptimizer(model_callback=cb1, metrics=[StubMetric()], optimizer_model=StubLLM())
o2 = PromptOptimizer(model_callback=cb2, metrics=[StubMetric()], optimizer_model=StubLLM())

print(o1.algorithm is o2.algorithm)                  # expected False
print(o1.model_callback.__name__)                    # cb1
print(o1.algorithm.scorer.model_callback.__name__)   # expected cb1
print(o1.algorithm.status_callback.__self__ is o2)   # expected False
```

Actual output on `main` (blob `e5cac3e5b` for this file, current at `abce857e5`):

```
True
cb1
cb2
True
```

So `o1` — built with `cb1` — will run `cb2`.

### Impact

`GEPA.execute()` / `a_execute()` run through `self.scorer`, so:

```python
optimizer_a = PromptOptimizer(model_callback=app_a, metrics=[MetricA()])
optimizer_b = PromptOptimizer(model_callback=app_b, metrics=[MetricB()])

best_a = optimizer_a.optimize(prompt=prompt_a, goldens=goldens_a)
```

`best_a` is optimized against **app B** and scored with **MetricB** — no exception, no warning, and a perfectly plausible-looking result. Progress output is routed to the wrong optimizer as well, and because `GEPA.execute()` calls `reset_state()` on entry, two optimizers running concurrently wipe each other's run state.

This affects every documented usage, since no example passes `algorithm=` (`docs/content/docs/prompt-optimization-introduction.mdx:37`, `:83`, `:133`). It also hits the common patterns of building optimizers in a loop or in a pytest fixture.

### Scope

I swept all 503 files in the package with `ast` for defaults that are constructor calls. Shared default *config* instances are an established idiom here (`= AsyncConfig()`, `= DisplayConfig()`, `= ErrorConfig()`, `= CacheConfig()` in `evaluate/evaluate.py`, `evaluate/compare.py`, `evaluate/execute/e2e.py`, `evaluate/execute/loop.py`) — those are read-only and hold no per-run state, so they are harmless. `prompt_optimizer.py:57` is the only shared default that gets mutated and carries per-run state, so the fix is one call site.

### Proposed fix

```python
algorithm: Optional[Union[GEPA, MIPROV2, COPRO, SIMBA]] = None,
...
self.algorithm = algorithm if algorithm is not None else GEPA()
```

`Optional` is already imported. Behaviour for existing callers is unchanged — the default is still GEPA, just a fresh instance per optimizer — so the docs stay accurate.

Opened #2999 with the fix and a regression test.

shalinis97:
Closes #2998

### Problem

`PromptOptimizer.__init__` used a mutable default argument (`deepeval/optimizer/prompt_optimizer.py:57`):

```python
algorithm: Union[GEPA, MIPROV2, COPRO, SIMBA] = GEPA(),
```

`GEPA()` is evaluated once at import, so every optimizer built without an explicit `algorithm=` shared a single algorithm object — and `_configure_algorithm()` mutates it (`scorer`, `_rewriter`, `status_callback`, `step_callback`, `optimizer_model`).

Constructing a second optimizer silently rewired the first. Since `execute()` runs through `self.scorer`, `optimizer_a.optimize(...)` would call **optimizer B's** `model_callback` and score with **B's** metrics — no error, no warning, just a confidently wrong optimized prompt. Every documented example is affected, as none pass `algorithm=`.

Before:

```
o1.algorithm is o2.algorithm             : True
o1.model_callback                        : cb1
o1.algorithm.scorer.model_callback       : cb2     <-- wrong
o1.algorithm.status_callback.__self__ is o2: True  <-- wrong
```

After:

```
o1.algorithm is o2.algorithm             : False
o1.model_callback                        : cb1
o1.algorithm.scorer.model_callback       : cb1
o1.algorithm.status_callback.__self__ is o2: False
```

### Fix

Default `algorithm` to `None` and construct a fresh `GEPA()` per instance. Behaviour for existing callers is unchanged — the default is still GEPA — so the docs stay accurate and there is no migration.

I deliberately left `async_config` / `display_config` alone: shared default *config* instances are an established idiom across `evaluate/evaluate.py`, `evaluate/compare.py`, `evaluate/execute/e2e.py` and `evaluate/execute/loop.py`, and those objects are read-only with no per-run state. An `ast` sweep of all 503 package files confirmed `prompt_optimizer.py:57` is the only shared default that is mutated and carries per-run state.

### Tests

Added `tests/test_core/test_optimization/test_prompt_optimizer_defaults.py` covering:

- two optimizers built without `algorithm=` get distinct algorithm and scorer objects
- building a second optimizer does not rewire the first's `model_callback` or status/step callbacks
- an explicitly passed `algorithm=` instance is still used as-is

The new file is intentionally not gated on `OPENAI_API_KEY` (unlike the sibling `test_prompt_optimizer.py`), since the bug is provable with no LLM calls — it uses a stub `DeepEvalBaseLLM` plus the existing `_DummyMetric` stub, so it actually runs in CI.

All three fail on `main` and pass with this change.

```
tests/test_core/test_optimization   81 passed, 7 skipped   (78 passed before this PR)
tests/test_core                   1245 passed, 56 skipped
black --check                     clean
```

### DIFF
--- a/deepeval/optimizer/prompt_optimizer.py
+++ b/deepeval/optimizer/prompt_optimizer.py
@@ -54,7 +54,7 @@ class PromptOptimizer:
         model_callback: ModelCallback,
         metrics: Union[List[BaseMetric], List[BaseConversationalMetric]],
         optimizer_model: Optional[Union[str, DeepEvalBaseLLM]] = None,
-        algorithm: Union[GEPA, MIPROV2, COPRO, SIMBA] = GEPA(),
+        algorithm: Optional[Union[GEPA, MIPROV2, COPRO, SIMBA]] = None,
         async_config: Optional[AsyncConfig] = AsyncConfig(),
         display_config: Optional[DisplayConfig] = DisplayConfig(),
     ):
@@ -71,7 +71,10 @@ class PromptOptimizer:
 
         self.async_config = async_config
         self.display_config = display_config
-        self.algorithm = algorithm
+        # Defaulted here rather than in the signature: `_configure_algorithm`
+        # mutates the algorithm, so a default instance would be shared by every
+        # PromptOptimizer built without an explicit `algorithm`.
+        self.algorithm = algorithm if algorithm is not None else GEPA()
         self.optimization_report = None
         self._configure_algorithm()
 
--- a/deepeval/test_case/llm_test_case.py
+++ b/deepeval/test_case/llm_test_case.py
@@ -257,7 +257,7 @@ class ToolCall(BaseModel):
         validation_alias=AliasChoices("inputParameters", "input_parameters"),
     )
 
-    @field_serializer('type')
+    @field_serializer("type")
     def serialize_type(self, value: ToolCallType) -> str:
         return value.value
 
--- a/tests/test_core/test_optimization/test_prompt_optimizer_defaults.py
+++ b/tests/test_core/test_optimization/test_prompt_optimizer_defaults.py
@@ -1,0 +1,82 @@
+from deepeval.models.base_model import DeepEvalBaseLLM
+from deepeval.optimizer.algorithms import COPRO
+from deepeval.optimizer.configs import DisplayConfig
+from deepeval.optimizer.prompt_optimizer import PromptOptimizer
+from deepeval.optimizer.scorer import Scorer
+
+from tests.test_core.stubs import _DummyMetric
+
+
+class _StubLLM(DeepEvalBaseLLM):
+    """Keeps `initialize_model` off the GPTModel path so no API key is needed."""
+
+    def load_model(self):
+        return self
+
+    def generate(self, *args, **kwargs):
+        return ""
+
+    async def a_generate(self, *args, **kwargs):
+        return ""
+
+    def get_model_name(self):
+        return "stub"
+
+
+def _callback_one(prompt, golden):
+    return "from-callback-one"
+
+
+def _callback_two(prompt, golden):
+    return "from-callback-two"
+
+
+def _build(model_callback):
+    return PromptOptimizer(
+        model_callback=model_callback,
+        metrics=[_DummyMetric()],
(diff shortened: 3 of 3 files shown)


shalinis97:
@penguine-ip , please review pr and merge it.

PR #3001, opened 2026-08-04
fix(g_eval): read the score token from the end of the logprobs (#3000)

### LINKED ISSUE #3000: GEval logprobs fails if reason contains score token.

```
        break
```

Anai-Guo:
Fixes #3000.

### Problem

`ReasonScore` declares `reason` before `score`, so the model emits the reasoning text first and the numeric verdict last. `calculate_weighted_summed_score` scans the token stream **forwards** for the first token equal to `str(raw_score)`:

```python
for token_logprobs in generated_logprobs:
```

If the reasoning happens to contain that same digit — which is common ("... so I rate this 8 out of 10") — the loop latches onto the token inside the reason and weighs *its* `top_logprobs` distribution. The returned score is then derived from an unrelated position in the output and no longer reflects the model's confidence in its verdict.

### Fix

Scan in reverse. The score token is the last thing the model writes, so the trailing match is the right one. One-line change plus a comment explaining why the direction matters.

### Verification

Extracted `calculate_weighted_summed_score` and exercised it against a stubbed `ChatCompletion` with two `"8"` tokens: an early one (from the reason) whose top-logprobs are a degenerate `{8: 1.0}`, and a trailing one (the real score) with `{8: 0.5, 9: 0.5}`.

| | reason contains the score digit | reason does not |
|---|---|---|
| before | `8.0` (wrong distribution) | `8.5` |
| after | `8.5` | `8.5` |

Both cases are now covered by unit tests in `tests/test_metrics/test_g_eval_utils.py` — the second one pins the unchanged behaviour so the reversal can't regress the ordinary path. `black --line-length 80` (25.1.0, per `pyproject.toml`) reports both touched files unchanged.

### DIFF
--- a/deepeval/metrics/conversational_g_eval/conversational_g_eval.py
+++ b/deepeval/metrics/conversational_g_eval/conversational_g_eval.py
@@ -390,9 +390,11 @@ class ConversationalGEval(BaseConversationalMetric):
         self, raw_score: int, raw_response: ChatCompletion
     ) -> Union[int, float]:
         generated_logprobs = raw_response.choices[0].logprobs.content
-        # First, locate the token that we care for logprobs, i.e., the token matching the score
+        # First, locate the token that we care for logprobs, i.e., the token matching the score.
+        # Scan backwards: the response schema emits "reason" before "score", so a digit that
+        # happens to appear in the reasoning text would otherwise shadow the actual score token.
         score_logprobs = None
-        for token_logprobs in generated_logprobs:
+        for token_logprobs in reversed(generated_logprobs):
             if token_logprobs.token == str(raw_score):
                 score_logprobs = token_logprobs
                 break
--- a/deepeval/metrics/g_eval/utils.py
+++ b/deepeval/metrics/g_eval/utils.py
@@ -338,9 +338,11 @@ def calculate_weighted_summed_score(
 ) -> Union[int, float]:
     try:
         generated_logprobs = raw_response.choices[0].logprobs.content
-        # First, locate the token that we care for logprobs, i.e., the token matching the score
+        # First, locate the token that we care for logprobs, i.e., the token matching the score.
+        # Scan backwards: the response schema emits "reason" before "score", so a digit that
+        # happens to appear in the reasoning text would otherwise shadow the actual score token.
         score_logprobs = None
-        for token_logprobs in generated_logprobs:
+        for token_logprobs in reversed(generated_logprobs):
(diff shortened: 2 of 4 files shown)


Anai-Guo:
CI note: the red `lint (ubuntu-latest)` is pre-existing drift on `main`, not from this PR.

Black flags exactly one file — `deepeval/test_case/llm_test_case.py` — which this PR doesn't touch:

```
would reformat .../deepeval/test_case/llm_test_case.py
1 file would be reformatted, 928 files would be left unchanged.
```

`black --line-length 80` (25.1.0, per `pyproject.toml`) leaves both files I changed unchanged. All `test` jobs and every integration job are green.

ebarkhordar (commented):
The reversal is right, and the same routine has a second copy it does not reach.
`ConversationalGEval.generate_weighted_summed_score`
(`deepeval/metrics/conversational_g_eval/conversational_g_eval.py:389` at this head) is a
line-for-line duplicate of `calculate_weighted_summed_score`, forwards scan included, at :395.
The conversational `ReasonScore` declares `reason` before `score` exactly as the g-eval one
does, so the shadowing case you describe applies to it unchanged, and both `evaluate` and
`_a_evaluate` reach it whenever `no_log_prob_support` is false (:373 and :306).

Feeding both functions the same stub your new test builds, an early `"8"` from the reason with
`{8: 1.0}` and a trailing `"8"` with `{8: 0.5, 9: 0.5}`, at fcd152d:

```
g_eval/utils.calculate_weighted_summed_score       -> 8.5
ConversationalGEval.generate_weighted_summed_score -> 8.0
```

The two copies have now drifted twice. The conversational one is also missing the
`sum_linear_probability == 0` fallback that `calculate_weighted_summed_score` picked up in
3a702cfd3, which is what open PR #2871 is porting across right now. Having the conversational
path call the shared helper would close both gaps at once and stop the third one.

None of this blocks the change here, which is correct on its own. Whether to widen the scope or
leave the sibling to #2871 is your call.

Anai-Guo:
@ebarkhordar thanks for the careful trace — you're right, the conversational copy had the same forward scan. I've applied the reversal fix to `ConversationalGEval.generate_weighted_summed_score` and added a parallel test that reproduces the 8.0→8.5 divergence you described.

I deliberately kept the change narrow to just the scan direction and left the `sum_linear_probability == 0` fallback to #2871, so the two PRs don't collide. Fully agree the long-term fix is to collapse both onto the shared helper once #2871 lands.

PR #3002, opened 2026-08-04
fix(g-eval): pick the trailing score token for weighted logprob score (#3000)

### LINKED ISSUE #3000: GEval logprobs fails if reason contains score token.

Anai-Guo:
## Summary

Fixes #3000.

`calculate_weighted_summed_score` locates the completion token whose text equals the raw score, then reconstructs a probability-weighted score from that token's `top_logprobs`. It searched **forward** and stopped at the first match:

```python
for token_logprobs in generated_logprobs:
        break
```

The model emits the completion as a JSON object `{"reason": ..., "score": N}` (see `ReasonScore` in `g_eval/schema.py` — `reason` first, `score` last), so the score digit is the **last** matching token. When the reasoning prose happens to contain the same digit as the final score (e.g. the reason mentions "8" and the score is also 8), the forward search latches onto the reasoning token's logprobs instead of the actual score token, corrupting the weighted score.

## Fix

Iterate from the end so the trailing score token wins (as suggested in the issue):

```python
for token_logprobs in reversed(generated_logprobs):
    ...
```

One-line behavioural change plus a clarifying comment.

## Test

Added a regression test in `tests/test_metrics/test_g_eval_utils.py` that builds a fake `logprobs.content` where the score digit appears earlier in the reasoning (with misleadingly certain logprobs) and again at the tail (split 8/7). With the old forward search the function returns `8.0`; with the fix it correctly returns `7.5`.

## Notes

Scope kept intentionally minimal to the file named in the issue. The same forward-search loop also exists inline in `deepeval/metrics/conversational_g_eval/conversational_g_eval.py::generate_weighted_summed_score`; happy to apply the identical fix there in a follow-up if you'd like it in the same PR.

### DIFF
--- a/deepeval/metrics/g_eval/utils.py
+++ b/deepeval/metrics/g_eval/utils.py
@@ -338,9 +338,13 @@ def calculate_weighted_summed_score(
 ) -> Union[int, float]:
     try:
         generated_logprobs = raw_response.choices[0].logprobs.content
-        # First, locate the token that we care for logprobs, i.e., the token matching the score
+        # First, locate the token that we care for logprobs, i.e., the token
+        # matching the score. Search from the end: the model emits the JSON as
+        # {"reason": ..., "score": N}, so the score is the last matching token.
+        # Iterating forward would wrongly latch onto an identical digit that
+        # happens to appear earlier in the reasoning text.
         score_logprobs = None
-        for token_logprobs in generated_logprobs:
+        for token_logprobs in reversed(generated_logprobs):
             if token_logprobs.token == str(raw_score):
                 score_logprobs = token_logprobs
                 break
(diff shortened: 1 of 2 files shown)


Anai-Guo:
Closing as a duplicate of #3001 — same fix (scan the logprobs in reverse so the score token isn't shadowed by an identical digit in the reasoning text). Consolidating review on #3001.

Anai-Guo:
Apologies — I opened this and #3001 for the same issue within a minute of each other. They carry the identical production fix (`reversed(generated_logprobs)` in `calculate_weighted_summed_score`); #3001 is the earlier one and has the broader regression test (it also covers the no-shadowing-token case). Closing this in favour of #3001 to keep the queue clean.

Issue #1137, opened 2024-11-03
Improving Answer Relevancy Template


dipanjanS:
This is just a request in terms of if we could make the answer relevancy a bit more strict or have a way to control it especially if we have 'idk' verdicts being generated which go into the reasoning prompt (based on the source code I checked out earlier)?

Example RAG generation:
![image](https://github.com/user-attachments/assets/8f1e5197-c5b7-4705-a562-4c0a2bf06972)

Eval Result:
![image](https://github.com/user-attachments/assets/ccd5c632-48ed-4ba3-bd3f-c5b8a417faea)

penguine-ip:
Hey @dipanjanS , the example you showed is actually what we intended to happen. A few months ago when the metric was being built this example when being confined to yes or no will actually output 'no'. I don't know about you but I think the verdict should be 'yes'. So what we did was we made 'idk' and 'yes' vs 'no' to make these statements acceptable, is this something you don't want to happen?

dipanjanS:
Hey @penguine-ip thanks for getting back, yes I didn't think that it would be a bug hence I put it more as a feature request. The problem here is, given a lot of this is of course based on relevancy different folks might have different opinion. 

In general what I felt was if the response is 'I don't know' I'm not sure if we can really treat the answer as relevant, example in case of the following one its easy to check for relevancy as we have some definite statements to compare with the input query.
 
![image](https://github.com/user-attachments/assets/e3faa65b-cedf-4deb-aa62-0ccaf9137148)

However when the RAG fails to answer because of lack of context or lack of issues of the LLM not being able to generate the response and says it doesn't know maybe we should treat that as a not a relevant answer w.r.t question as it is either the shortcoming of the retriever or generator leading to that. 

In my initial example earlier, if you check the reasoning above also where it says the response is completely relevant and addresses the question, that is kind of misleading as it doesn't really address the question. So my thought was rather than getting rid of 'idk' is there a way we could control it using a setting to either treat it as a 'no'or 'yes' or somewhere in between, then people can take a call and set it up as they wish to.

I mean I can kind of do it manually as I understand your source code (kudos on the excellent code btw its really easy to understand) but this kind of an argument or setting might be useful even for other metrics. Happy to hear your thoughts.

penguine-ip:
[...]

PR #3006, opened 2026-08-05
feat: optionally penalize ambiguous answer relevancy verdicts

Yashaswini1233:
## Summary

* Adds a `penalize_ambiguous_statements` option to `AnswerRelevancyMetric`.
* When enabled, `idk` verdicts are treated as irrelevant when calculating the answer relevancy score.
* Preserves the existing behavior by default with `penalize_ambiguous_statements=False`.
* Includes ambiguous verdicts in the generated reason when penalization is enabled.

## Motivation

Currently, `AnswerRelevancyMetric` counts every verdict other than `no` as relevant, which means an `idk` verdict contributes positively to the score.

This change provides an opt-in stricter scoring mode for users who want ambiguous statements to be penalized while remaining backward compatible with the existing behavior.

Addresses #1137.

## Testing

Added focused tests covering:

* Existing/default `idk` scoring behavior
* Penalized `idk` scoring
* Ambiguous-only verdicts
* Strict mode behavior
* Existing empty-output regression behavior

Validation performed locally:

* 8 targeted tests passed
* `black --check` passed
* Python compilation passed
* `git diff --check` passed

### DIFF
--- a/deepeval/metrics/answer_relevancy/answer_relevancy.py
+++ b/deepeval/metrics/answer_relevancy/answer_relevancy.py
@@ -37,6 +37,7 @@ class AnswerRelevancyMetric(BaseMetric):
         async_mode: bool = True,
         strict_mode: bool = False,
         verbose_mode: bool = False,
+        penalize_ambiguous_statements: bool = False,
         flaky: bool = False,
     ):
         self.threshold = 1 if strict_mode else threshold
@@ -46,6 +47,7 @@ class AnswerRelevancyMetric(BaseMetric):
         self.async_mode = async_mode
         self.strict_mode = strict_mode
         self.verbose_mode = verbose_mode
+        self.penalize_ambiguous_statements = penalize_ambiguous_statements
         self.flaky = flaky
 
     def measure(
@@ -163,8 +165,16 @@ class AnswerRelevancyMetric(BaseMetric):
 
         irrelevant_statements = []
         for verdict in self.verdicts:
-            if verdict.verdict.strip().lower() == "no":
+            normalized_verdict = verdict.verdict.strip().lower()
+
+            if normalized_verdict == "no":
                 irrelevant_statements.append(verdict.reason)
+
+            if (
+                normalized_verdict == "idk"
+                and self.penalize_ambiguous_statements
+            ):
+                irrelevant_statements.append(f"(Ambiguous) {verdict.reason}")
 
         prompt = self._get_prompt(
             "generate_reason",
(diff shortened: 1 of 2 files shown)


Yashaswini1233:
Note on CI: the lint check is currently failing on `deepeval/test_case/llm_test_case.py`, which is unchanged by this PR and identical to the version on `upstream/main`. I also reproduced the check locally with Black 25.12.0. The changes introduced in this PR are Black-formatted and the 8 targeted Answer Relevancy tests pass locally.

deepeval/metrics/answer_relevancy/answer_relevancy.py
```diff
@@ -295,16 +313,29 @@ async def _a_generate_statements(
             + [ele for ele in actual_output if isinstance(ele, MLLMImage)],
         )
 
+    def _calculate_score(self):
```
rajeshl8: Duplicate method definition. 
This PR adds a new _calculate_score method at line 313 that only contains the empty-verdicts guard (return 1), but the original full _calculate_score is still defined below it at ~line 325. In Python, when a method is defined twice in the same class body, the second definition silently overrides the first — so this new one is unreachable dead code and will never execute.

Yashaswini1233:
The duplicate _calculate_score() definition has been removed, leaving one complete scoring method with the empty-verdict guard intact. All 8 targeted tests pass locally. The remaining Black CI failure is on .scripts/release.py, which is unrelated to this PR. The Vercel check requires repository authorization.

deepeval/metrics/answer_relevancy/answer_relevancy.py
Yashaswini1233: Thanks for catching this. I removed the duplicate _calculate_score() definition and retained the empty-verdict guard in the single complete method. All 8 targeted tests pass locally.

rajeshl8: Thanks for addressing it. Confirmed — only one _calculate_score definition now. The penalization logic looks correct.

Yashaswini1233: Thanks for confirming! Glad the changes look good. Please let me know if there are any additional changes or checks needed from my side before this is ready to merge.

Yashaswini1233: Hi @maintainer, just following up on this PR when you get a chance. The requested changes and tests have been completed, and the targeted tests are passing. Please let me know if there’s anything else I should update to help move this toward review/merge. Thanks!

PR #3021, opened 2026-08-09
feat: report a confidence interval alongside the aggregate pass rate

ipezygj:
## What

The aggregate metrics table reports a pass rate as if it were an exact property of the system under test. It is an estimate from however many test cases happened to be in the run, and small runs are common.

This adds a 95% confidence interval next to the rate:

```
Before:  50.00% | passed=1 | failed=1
After:   50.00% (95% CI 9.5-90.5%) | passed=1 | failed=1

Before:  80.00% | passed=4 | failed=1
After:   80.00% (95% CI 37.6-96.4%) | passed=4 | failed=1

Before:  92.00% | passed=460 | failed=40
After:   92.00% (95% CI 89.3-94.1%) | passed=460 | failed=40
```

The last line is the point: with 500 cases the number means something, and the interval says so. With 5 cases it doesn't, and the interval says that too — which is otherwise invisible to someone reading a CI log or comparing two runs.

## Why Wilson and not the usual formula

The textbook normal (Wald) interval collapses to zero width at 0% and 100%, so a suite where everything passes would report `100.00% (95% CI 100.0-100.0%)` no matter how few cases ran. Those are the two values eval suites hit most often. The Wilson score interval is well behaved there (`10/10` gives 72.2%–100%) and has better coverage at small n generally.

## Changes

- `deepeval/evaluate/statistics.py` (new): `wilson_score_interval()` and `format_pass_rate_with_interval()`. Pure `math`, no new dependency, no scipy. The 95% normal quantile is a module constant so the result is deterministic.
- `deepeval/evaluate/console_report.py`: `_format_pass_rate()` uses it. That one function feeds both the console table and the markdown/HTML exports, so all three surfaces pick it up from a single change.
- `tests/test_core/test_evaluation/test_statistics.py` (new): 12 tests.
- `tests/test_core/test_evaluation/test_console_report.py`: the existing assertion pinned the exact rate string, so it is updated to the new one.

## Testing

`tests/test_core/test_evaluation/test_statistics.py` — 12 passed:

- reference value: 5/10 matches the published Wilson figure [0.2366, 0.7634]
- the interval brackets the point estimate across several sample sizes
- 0/10 and 10/10 still have width (the case Wald gets wrong)
- the interval narrows as n grows at a fixed proportion
- bounds stay inside [0, 1]
- invalid inputs (`total <= 0`, successes out of range) raise
- the rendered console cell contains the interval, and the no-verdicts case still renders `N/A`

Full-directory check: `tests/test_core/test_evaluation/` has 17 failures on master in my environment (local-store lock tests, missing optional deps) and the same 17 with this branch, with 12 added passes. I did not manage to get a fully green local suite, so the failures I am reporting as pre-existing were established by running the same directory with and without the change and diffing the failure lists.

`black` was run over the touched files.

## Notes

`deepeval/evaluate/utils.py` prints and writes pass rates in a couple of other places (`{metric}: {pass_rate:.2%} pass rate`). I left those alone to keep the diff focused, but they could take the same treatment if you want it — happy to add it in this PR or a follow-up.

### DIFF
--- a/deepeval/evaluate/statistics.py
+++ b/deepeval/evaluate/statistics.py
@@ -7,14 +7,28 @@ helpers put the sample size back into the reported number.
 """
 
 import math
+from statistics import NormalDist
 
-# Two-sided 95% standard-normal quantile. Hardcoded so this module stays
-# dependency-free (no scipy) and deterministic.
-Z_95 = 1.959963984540054
+_STANDARD_NORMAL = NormalDist()
+
+
+def z_for_confidence(confidence: float = 0.95) -> float:
+    """Two-sided standard-normal quantile for a confidence level.
+
+    `confidence=0.95` returns 1.9599639845..., the familiar 1.96. Derived from
+    the stdlib rather than hardcoded so the level a caller asks for and the
+    level the arithmetic uses cannot drift apart.
+
+    Raises:
+        ValueError: if `confidence` is not strictly between 0 and 1.
+    """
+    if not 0.0 < confidence < 1.0:
+        raise ValueError("confidence must be strictly between 0 and 1")
+    return _STANDARD_NORMAL.inv_cdf(1.0 - (1.0 - confidence) / 2.0)
 
 
 def wilson_score_interval(
-    successes: int, total: int, z: float = Z_95
+    successes: int, total: int, confidence: float = 0.95
 ) -> tuple[float, float]:
     """Wilson score interval for a binomial proportion, as fractions in [0, 1].
 
@@ -25,19 +39,21 @@ def wilson_score_interval(
     Args:
         successes: number of passing verdicts.
         total: number of verdicts (passes + fails).
-        z: standard-normal quantile; defaults to a two-sided 95% interval.
+        confidence: two-sided confidence level, e.g. 0.95 or 0.99.
 
     Returns:
         (low, high) as fractions of 1.
 
     Raises:
-        ValueError: if `total` is not positive or `successes` is out of range.
+        ValueError: if `total` is not positive, `successes` is out of range, or
+            `confidence` is not strictly between 0 and 1.
     """
     if total <= 0:
         raise ValueError("total must be positive to form an interval")
     if not 0 <= successes <= total:
         raise ValueError("successes must be between 0 and total")
 
+    z = z_for_confidence(confidence)
     p = successes / total
     z2 = z * z
     denominator = 1 + z2 / total
@@ -49,9 +65,14 @@ def wilson_score_interval(
 
 
 def format_pass_rate_with_interval(
-    successes: int, total: int, z: float = Z_95
+    successes: int, total: int, confidence: float = 0.95
 ) -> str:
-    """'80.00% (95% CI 37.6-96.4%)' — a pass rate that shows what the sample supports."""
+    """'80.00% (95% CI 37.6-96.4%)' — a pass rate that shows what the sample supports.
+
+    The label reports the level the interval was actually computed at, so
+    `confidence=0.99` prints "99% CI" rather than a stale "95%".
+    """
     rate = (successes / total) * 100
-    low, high = wilson_score_interval(successes, total, z)
-    return f"{rate:.2f}% (95% CI {low * 100:.1f}-{high * 100:.1f}%)"
+    low, high = wilson_score_interval(successes, total, confidence)
+    label = f"{confidence * 100:g}%"
+    return f"{rate:.2f}% ({label} CI {low * 100:.1f}-{high * 100:.1f}%)"
--- a/deepeval/evaluate/statistics.py
+++ b/deepeval/evaluate/statistics.py
@@ -1,0 +1,57 @@
+"""Sampling-uncertainty helpers for aggregate evaluation results.
+
+A pass rate is an estimate from a finite number of test cases, not an exact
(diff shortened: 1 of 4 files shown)


ipezygj:
Heads-up on the red `lint` check here: it isn't from this branch. `black --check` fails on `main` too, on `deepeval/test_case/llm_test_case.py` (a single-quoted decorator argument from #3003), so it currently fails every open PR's lint run. I verified it by running black against `main`'s copy of that file on its own.

I've put the one-character fix in #3022 so it can land independently of this change. Once that's in, this branch's lint should go green on a re-run.

deepeval/evaluate/statistics.py
```diff
@@ -0,0 +1,57 @@
...
+    return max(0.0, center - margin), min(1.0, center + margin)
+
+
+def format_pass_rate_with_interval(
+    successes: int, total: int, z: float = Z_95
+) -> str:
+    """'80.00% (95% CI 37.6-96.4%)' — a pass rate that shows what the sample supports."""
+    rate = (successes / total) * 100
+    low, high = wilson_score_interval(successes, total, z)
+    return f"{rate:.2f}% (95% CI {low * 100:.1f}-{high * 100:.1f}%)"
```
rajeshl8: Is `95%` intended to be a hardcoded value, or should it be derived from `z`? I also noticed that `z` currently defaults to `95`.

My suggestion:

1. Replace `z` with a clearer confidence-level input:
   `confidence: float = 0.95`
2. Compute `ci_pct` dynamically from the input rather than hardcoding `95%`:
   `ci_pct = round((1 - 2 * (1 - scipy.stats.norm.cdf(z))) * 100)`

This would keep the displayed confidence interval consistent with the value actually used in the calculation.

tests/test_core/test_evaluation/test_console_report.py
```diff
@@ -129,7 +129,12 @@ def test_evaluation_console_report_aggregate_metrics():
     # columns[0] is Metric, columns[1] is Average Score, columns[2] is Pass Rate, columns[3] is Total
     assert list(table.columns[0].cells)[0] == "Answer Relevancy"
     assert list(table.columns[1].cells)[0] == "0.50"
-    assert list(table.columns[2].cells)[0] == "50.00% | passed=1 | failed=1"
+    # The rate carries its 95% interval: 1 of 2 is "50%", but two test cases
+    # only pin the true rate somewhere between 9.5% and 90.5%.
+    assert (
+        list(table.columns[2].cells)[0]
+        == "50.00% (95% CI 9.5-90.5%) | passed=1 | failed=1"
```
rajeshl8, line 136: Once you update the source code, test need to be updated.

deepeval/evaluate/statistics.py
ipezygj: You are right that the label and the arithmetic could disagree, and that is the part worth fixing. Done in f1cca90.

One correction on the reading: `z` did not default to `95` — it defaulted to `Z_95 = 1.959963984540054`, the quantile, not the percentage. But that is exactly the confusion your comment demonstrates the API was inviting, so the fix is the one you asked for:

- `z: float = Z_95` is now `confidence: float = 0.95` on both `wilson_score_interval` and `format_pass_rate_with_interval`.
- The quantile is derived by a new `z_for_confidence()`, and the printed label is derived from the same argument, so `confidence=0.99` prints `99% CI` over an interval actually computed at 0.99.

I took one liberty with your snippet. `scipy.stats.norm.cdf` would pull scipy into a module that is currently stdlib-only, and going from a level to a quantile needs the inverse anyway. `statistics.NormalDist().inv_cdf` is stdlib and gives both directions:

```python
def z_for_confidence(confidence: float = 0.95) -> float:
    if not 0.0 < confidence < 1.0:
        raise ValueError("confidence must be strictly between 0 and 1")
    return NormalDist().inv_cdf(1.0 - (1.0 - confidence) / 2.0)
```

Label formatting uses `f"{confidence * 100:g}%"` so 0.999 prints `99.9%` rather than rounding to `100%`.

Tests updated, as you asked in the second comment. Four new ones: the quantiles reproduce 1.95996 and 2.57583, invalid levels raise, a higher level gives a strictly wider interval, and the label matches the interval it is printed over.

That last change also caught something. Swapping the hardcoded constant for `inv_cdf` moves `z` by one ulp, which moves the 0/10 lower bound from exactly `0.0` to `2.8e-17` — algebraically zero, floating-point residue in practice. The old test asserted `low == 0.0` while its own docstring explained why the 10/10 side needed a tolerance. It now uses `math.isclose(..., abs_tol=1e-12)` on both ends.

19 tests pass. `tests/test_core/test_evaluation/test_console_report.py::test_evaluation_console_report_exports` fails for me on `main` as well — a cp1252 decode of the exported HTML on Windows, unrelated to this PR.

rajeshl8: Fair correction — I misread Z_95 as a percentage rather than a quantile. The confusion was the point, though, and the confidence: float = 0.95 API is much clearer for it. The **statistics.NormalDist().inv_cdf** call is a better choice than my scipy snippet — keeps it stdlib-only and gets the inverse direction right. The f"{confidence * 100:g}%" formatting, which handles 0.999 → 99.9% instead of rounding to 100%, is a good catch too. Looks solid.

PR #3023, opened 2026-08-09
fix(contextual_recall): strip whitespace before matching verdicts

WatchTree-19:
### what

`ContextualRecallMetric` (and `turn_contextual_recall`) count justified sentences with:

```python
if verdict.verdict.lower() == "yes":
```

no `.strip()`. every other metric in the codebase normalizes verdicts with `.strip().lower()` (bias, faithfulness, prompt_alignment, contextual_precision, ...). so when a judge verdict arrives padded, e.g. `"yes\n"` or `" yes"`, it fails the equality and is dropped, and recall is silently undercounted.

### repro

on verdicts `["yes", "yes\n", " yes", "no"]`, `_calculate_score()` returns `0.25` instead of the correct `0.75` - three of the four are justified, but only the unpadded one is counted.

### fix

normalize all five verdict comparisons in `contextual_recall` and `turn_contextual_recall` to `.strip().lower()`, matching the rest of the metrics. added a key-free regression test on `_calculate_score` (fails on main, passes with the change). black + ruff clean.


deepeval/metrics/contextual_recall/contextual_recall.py
```diff
@@ -200,7 +200,7 @@ async def _a_generate_reason(self, expected_output: str, multimodal: bool):
         supportive_reasons = []
         unsupportive_reasons = []
         for verdict in self.verdicts:
-            if verdict.verdict.lower() == "yes":
+            if verdict.verdict.strip().lower() == "yes":
```
rajeshl8, line 203: Good catch — trailing whitespace in LLM-returned verdicts can cause this issue. The same pattern (`verdict.lower() == "yes"` without `.strip()`) appears in several other metrics as well.

A quick grep on main:

```bash
grep -rn '\.lower() == "yes"' deepeval/metrics/ | grep -v strip
```

It flags faithfulness, hallucination, summarization, and a few others. It might be worth opening a follow-up issue or a sweep PR to fix this consistently, rather than on a metric-by-metric basis. 
Happy to open an issue if helpful.

WatchTree-19:
swept the metrics tree for the same pattern. the only other free-str verdict site was contextual_relevancy (3 comparisons), fixed in the latest commit. faithfulness, hallucination and summarization already strip. topic_adherence uses a Literal["TP","TN","FP","FN"] verdict so pydantic constrains it, no whitespace risk there. the utils.py matches are config values (should_use_* reading stored yes/no settings) not model verdicts, so left those alone.

rajeshl8 (commented):
Thanks for doing the sweep — that's exactly the right approach. Good to know faithfulness, hallucination, and summarization already strip. The contextual_relevancy fix in the latest commit looks correct, and the reasoning for leaving utils.py and topic_adherence alone makes sense.
Overall looks good to me.

PR #3029, opened 2026-08-10
fix(tool_correctness): give a meaningful reason for reordered repeated tools

### LINKED ISSUE #3014: Confusion in Tool Correctness metric
The Tool Correctness metric is confusing regarding what **should_consider_ordering** means versus strict ordering.

Does it create a dictionary for comparison, or a plain list where only the first execution is kept and subsequent ones are removed before the list is compared?
```
**Optional parameters**

- **should_consider_ordering** – Boolean. When True, the metric considers the order in which tools were called.  
  Example:  
  `expected_tools = [ToolCall(name="WebSearch"), ToolCall(name="ToolQuery"), ToolCall(name="WebSearch")]`  
  `tools_called   = [ToolCall(name="WebSearch"), ToolCall(name="WebSearch"), ToolCall(name="ToolQuery")]`  
  The metric treats this as correct. This applies to tool‑name matching, which is always performed. Default: False.

- **should_exact_match** – Boolean. When True, `tools_called` must match `expected_tools` exactly. In addition to tool‑name matching, it checks `ToolCallParams.INPUT_PARAMETERS` and `ToolCallParams.OUTPUT` when included in `evaluation_params`. Default: False.
```

Link: https://deepeval.com/docs/metrics-tool-correctness

Anai-Guo:
## Summary

Fixes #3014.

When `ToolCorrectnessMetric` runs with `should_consider_ordering=True` and a **repeated** tool is called out of order, the weighted-LCS score correctly drops below 1 — but the human-readable reason came out malformed.

### Root cause

In `_generate_reason`, the ordering branch builds the `issues` list from two set-based checks:

```python
missing = set(expected_tools_names) - set(tools_called_names)
out_of_order = set(expected_tools_names) - set([t.name for t in lcs])
```

For the exact case in #3014 — `expected = [WebSearch, ToolQuery, WebSearch]`, `called = [WebSearch, WebSearch, ToolQuery]` — the weighted LCS is length 2, so `score = 2/3 < 1`. But every expected **name** appears both in `tools_called` and in the LCS, so `missing` and `out_of_order` are both empty. `issues` stays `[]`, and the reason renders as a dangling:

```
Incorrect tool usage: ; expected ['WebSearch', 'ToolQuery', 'WebSearch'], called ['WebSearch', 'WebSearch', 'ToolQuery']. See more details above.
```

The set-based checks lose tool *multiplicity*, so they cannot describe a duplicate-tool reordering.

### Fix

When the score is below 1 but no missing/out-of-order names were found, fall back to a generic ordering message so the reason is always informative:

```
Incorrect tool usage: tools were called in a different order than expected; expected [...], called [...]. See more details above.
```

Scoring is unchanged — this only affects the reason string. Cases that already produced `missing tools ...` / `out-of-order tools ...` are untouched.

## Test

Adds `tests/test_core/test_tool_correctness_metric.py` with a hermetic (offline stub model) regression test covering the #3014 case and asserting the existing missing / out-of-order / correct-ordering reasons still render.

### DIFF
--- a/deepeval/metrics/tool_correctness/tool_correctness.py
+++ b/deepeval/metrics/tool_correctness/tool_correctness.py
@@ -298,12 +298,28 @@ class ToolCorrectnessMetric(BaseMetric):
                     # Every expected tool name was called, but a repeated tool
                     # was called in the wrong order, so the weighted LCS dropped
                     # an occurrence and the score fell below 1. The set-based
-                    # `missing`/`out_of_order` checks cannot see this, so fall
-                    # back to a generic ordering message instead of emitting an
-                    # empty "Incorrect tool usage: ;" reason.
-                    issues.append(
-                        "tools were called in a different order than expected"
-                    )
+                    # `missing`/`out_of_order` checks compare names, not counts,
+                    # so they cannot see this and the reason would otherwise
+                    # render as an empty "Incorrect tool usage: ;". Name the
+                    # tools whose repeated calls the LCS could not line up.
+                    matched_names = [tool.name for tool in lcs]
+                    repeated = [
+                        name
+                        for name in dict.fromkeys(expected_tools_names)
+                        if matched_names.count(name)
+                        < expected_tools_names.count(name)
+                    ]
+                    if repeated:
+                        issues.append(
+                            f"repeated tools {repeated} called in the wrong order"
+                        )
+                    else:
+                        # Every occurrence lined up, so the score is below 1
+                        # because a matched call only partially satisfied the
+                        # evaluation params (e.g. input parameters).
+                        issues.append(
+                            "called tools did not fully match the expected tools"
+                        )
                 return f"Incorrect tool usage: {' and '.join(issues)}; expected {expected_tools_names}, called {tools_called_names}. See more details above."
         else:
             used_expected = set(self.tools_called).intersection(
--- a/deepeval/metrics/tool_correctness/tool_correctness.py
+++ b/deepeval/metrics/tool_correctness/tool_correctness.py
@@ -294,6 +294,14 @@ class ToolCorrectnessMetric(BaseMetric):
                     issues.append(f"missing tools {list(missing)}")
                 if out_of_order:
                     issues.append(f"out-of-order tools {list(out_of_order)}")
+                if not issues:
+                    # Every expected tool name was called, but a repeated tool
+                    # was called in the wrong order, so the weighted LCS dropped
+                    # an occurrence and the score fell below 1. The set-based
+                    # `missing`/`out_of_order` checks cannot see this, so fall
+                    # back to a generic ordering message instead of emitting an
+                    # empty "Incorrect tool usage: ;" reason.
+                    issues.append("tools were called in a different order than expected")
                 return f"Incorrect tool usage: {' and '.join(issues)}; expected {expected_tools_names}, called {tools_called_names}. See more details above."
         else:
             used_expected = set(self.tools_called).intersection(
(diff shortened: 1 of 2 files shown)


deepeval/metrics/tool_correctness/tool_correctness.py
```diff
@@ -306,6 +294,16 @@ def _generate_reason(self):
...
                 if out_of_order:
                     issues.append(f"out-of-order tools {list(out_of_order)}")
+                if not issues:
+                    # Every expected tool name was called, but a repeated tool
+                    # was called in the wrong order, so the weighted LCS dropped
+                    # an occurrence and the score fell below 1. The set-based
+                    # `missing`/`out_of_order` checks cannot see this, so fall
+                    # back to a generic ordering message instead of emitting an
+                    # empty "Incorrect tool usage: ;" reason.
+                    issues.append(
```
rajeshl8: The fallback fixes the empty-reason regression, but the message is still less actionable than the other cases because it doesn’t say which tool caused the mismatch.

In evaluation pipelines, the reason is useful for debugging agent behavior, especially when a repeated tool call causes the weighted-LCS score to drop. Could we include the repeated tool name here?

Something like: "repeated tool(s) [...] called in wrong order".

Anai-Guo:
Good call @rajeshl8 — the generic message loses exactly the information you'd want when a long trace has several repeated tools. Pushed in ef53889.

The fallback now diffs per-name **counts** (expected vs. what the weighted LCS actually matched) rather than sets, so it can name the tool whose repeated occurrence got dropped:

```
Incorrect tool usage: repeated tools ['WebSearch'] called in the wrong order;
expected ['WebSearch', 'ToolQuery', 'WebSearch'],
called ['WebSearch', 'WebSearch', 'ToolQuery']. See more details above.
```

Two notes on the edge cases:

- `ToolQuery` is deliberately **not** listed — its single occurrence did line up, so naming it would send someone debugging the wrong call.
- There is a second way to land in this branch: every occurrence lines up but a matched call only partially satisfies `evaluation_params` (e.g. `INPUT_PARAMETERS` scoring `0 < s < 1`). Claiming "wrong order" there would be false, so that case now says `called tools did not fully match the expected tools`.

Verified old-vs-new on the same five inputs — only the reordered-repeat case changes; `missing tools`, `out-of-order tools`, and `Correct ordering` are byte-identical to `main`. The regression test now asserts the tool name.

rajeshl8:
@Anai-Guo the per-name count diff is the right call — much better than my set-based sketch since it correctly excludes ToolQuery (which did line up) and only surfaces tools whose repeated occurrences were actually dropped by the LCS. The evaluation_params edge case is a good catch too. The example output is exactly what you'd want when debugging a long agent trace. Looks great.

Anai-Guo:
@rajeshl8 thanks — glad the per-name count diff reads right.

Status on the `CONFLICTING` flag: `main` has since added a `type_mismatches` issue to this same `issues` list, which collides with the block this PR adds. I have the merge resolved and verified locally, but I **can't push it at the moment** — my token lacks the `workflow` scope and merging `main` in updates `.github/workflows/agent_skills_check.yml`, so the push is refused. Getting that sorted; posting the resolution here meanwhile.

The ordering matters, so it's worth stating explicitly — upstream's concrete check goes **before** this PR's fallback, since the fallback exists precisely to catch "no concrete issue was appended":

```python
if missing:
    issues.append(f"missing tools {list(missing)}")
if out_of_order:
    issues.append(f"out-of-order tools {list(out_of_order)}")
if type_mismatches:                       # from main
    issues.append(f"tool type mismatches {type_mismatches}")
if not issues:                            # this PR — must stay last
    ...
```

Taking either side alone would be wrong: main's version drops the empty-reason fix, and this branch's version would shadow a type-mismatch-only failure behind the generic fallback.

Verified after the merge — `tests/test_core/test_tool_correctness_metric.py` 3 passed, and driving `_generate_reason()` directly across the branches:

| case | reason |
|---|---|
| repeated tool out of order | `Incorrect tool usage: repeated tools ['a'] called in the wrong order; ...` |
| missing tool | `Incorrect tool usage: missing tools ['b'] and out-of-order tools ['b']; ...` |
| all correct | `Correct ordering: all expected tools ['a', 'b'] were called in the correct order.` |

(main's own type-mismatch tests skip locally — they're gated on `OPENAI_API_KEY`.)

PR #3030, opened 2026-08-10
fix(models): force temperature=1 for o1-preview reasoning models

### LINKED ISSUE #1820: O1 Models Incompatible with ArenaGEval due to response_format Parameter
❗BEFORE YOU BEGIN❗
  Are you on discord? 🤗 We'd love to have you asking questions on discord instead: https://discord.com/invite/a3K9c8GRGt

  Describe the bug
  O1 models (o1, o1-2024-12-17, o3-mini, etc.) fail when used with ArenaGEval because DeepEval internally passes a response_format parameter with type json_schema, which O1 models don't support.

  To Reproduce
  1. Create an O1 model: gpt_judge = GPTModel(model="o1")
  2. Create an ArenaGEval metric with the O1 model: arena_metric = ArenaGEval(name="test", criteria="Compare outputs", model=gpt_judge)
  3. Create test cases and arena test case with contestants
  4. Call await arena_metric.a_measure(arena_test_case)
  5. See error: Error code: 400 - {'error': {'message': \"Unsupported parameter: 'temperature' is not supported with this model.\", 'type': 'invalid_request_error', 'param': 'temperature', 'code': 'u...",

  Expected behavior
  O1 models should work with ArenaGEval, either by automatically detecting that the model doesn't support structured output and falling back to text parsing, or by providing a configuration option to disable structured output. Please also add support for O3

  Desktop (please complete the following information):
  - OS: Linux Ubuntu
  - DeepEval Version: Latest
  - Python Version: 3.11

biztex:
## Summary

Using an **o1-preview** reasoning model as a judge (e.g. with `ArenaGEval`) fails with:

```
Error code: 400 - Unsupported parameter: 'temperature' is not supported with this model.
```

## Root cause

`OPENAI_MODELS_DATA` groups the reasoning models under the comment *"Reasoning models - require temperature=1 (no custom temperature)"*, and every entry sets `supports_temperature=False` — **except** `o1-preview` and `o1-preview-2024-09-12`, which omitted it. `DeepEvalModelData.supports_temperature` defaults to `True`, so those two entries advertised temperature support.

`GPTModel` only overrides the caller's temperature when the model does **not** support it:

```python
if self.model_data.supports_temperature is False:
    temperature = 1
```

So for `o1-preview`, a user-supplied `temperature=0` was passed straight through to the OpenAI API, which rejects any non-default temperature for o1 models — producing the 400 above.

Observed before the fix:

```
o1                    supports_temperature=False  resolved=1
o1-preview            supports_temperature=True   resolved=0.0   <-- 400 error
o1-preview-2024-09-12 supports_temperature=True   resolved=0.0   <-- 400 error
o1-mini / o3-mini …   supports_temperature=False  resolved=1
```

## Fix

Set `supports_temperature=False` on both `o1-preview` entries so their temperature is forced to `1`, matching every sibling reasoning model.

## Also in this PR (review feedback)

- **`gpt-5-chat-latest`** had the same invisible omission — it sits in the reasoning block and relied on the default `True`. It is a *non-reasoning* chat model that genuinely accepts a temperature, so `True` is correct; I set `supports_temperature=True` **explicitly** (with a comment) so it isn't mistaken for a bug and "fixed" to `False`.
- **Azure impact:** `AzureOpenAIModel` reads the same `OPENAI_MODELS_DATA` registry and drops `temperature` to `None` when `supports_temperature is False`, so this fix also repairs **Azure** deployments of `o1-preview`, not just `GPTModel`.

## Tests

`tests/test_core/test_models/test_openai_model.py`:
- `test_o1_reasoning_models_auto_temperature_adjustment` — resolved temperature is `1` across the o1 family (o1, o1-preview, o1-preview-2024-09-12, o1-mini, o3-mini, o4-mini).
- `test_o1_preview_model_data_disables_temperature` — both o1-preview entries advertise `supports_temperature=False`.

`tests/test_core/test_models/test_azure_model.py`:
- `test_o1_preview_temperature_is_none` — `AzureOpenAIModel(model="o1-preview")` resolves `temperature` to `None`.

Full `test_openai_model.py` + `test_azure_model.py` pass (70 tests). Changed files are clean under `black` (line-length 80) and `ruff`.

Fixes #1820


ebarkhordar (commented):
Confirmed the leak and the fix in a clean `python:3.11-slim` container (`pip install -e .`, no network calls, `OPENAI_API_KEY` set to a dummy value), reading `GPTModel(model=..., temperature=0).temperature` at `b0dfad94` and at its base `e2bf78ef`:

```
                       base e2bf78ef   head b0dfad94
o1-preview                       0.0               1
o1-preview-2024-09-12            0.0               1
o1 / o1-mini / o3-mini             1               1
o4-mini / gpt-5.4                  1               1
```

One thing worth settling while you are in this block. Parsing `OPENAI_MODELS_DATA` with `ast` (46 entries), `gpt-5-chat-latest` at constants.py:367 is now the only entry under the `# Reasoning models - require temperature=1` comment at line 222 that omits `supports_temperature`, so it still resolves to `True` and keeps the caller's temperature (measured: 0.0 at both commits). That is right on the API side, since `gpt-5-chat-latest` is the non-reasoning chat model, but it is the same shape that produced this bug: the section comment says the whole block requires temperature=1 and the field defaults to `True`, so the exception is invisible. Writing `supports_temperature=True` on it explicitly would make it readable and stop the next person "fixing" it the way you just fixed o1-preview.

Second, `AzureOpenAIModel` reads the same registry (`azure_model.py:141`) and drops the parameter entirely (`temperature = None`) when the flag is False, so this also fixes Azure deployments of `o1-preview`, not only `GPTModel`. Worth saying in the PR description, since the two new tests only cover `GPTModel`.

biztex:
Thanks for the thorough review, @ebarkhordar — both points were spot on. Addressed in d185cdcd:

1. **`gpt-5-chat-latest`** — you're right that `True` is correct there (non-reasoning chat model), it was just implicit and sitting under the reasoning-block comment like the o1-preview omission. I set `supports_temperature=True` explicitly with a short comment so the next person doesn't "fix" it to `False`.

2. **Azure** — good catch. `AzureOpenAIModel` reads the same `OPENAI_MODELS_DATA` and drops `temperature` to `None` when the flag is `False`, so the o1-preview fix covers Azure deployments too. I called that out in the PR description and added `test_o1_preview_temperature_is_none` to `TestAzureModelTemperature`, mirroring the existing `o3-mini` case.

Full `test_openai_model.py` + `test_azure_model.py` pass (70 tests); `black`/`ruff` clean.

PR #3038, opened 2026-08-11
fix: clone custom metrics without constructors

### LINKED ISSUE #3037: copy_metrics raises TypeError for any metric subclass with a narrower __init__, aborting the whole async evaluate() run
**❗BEFORE YOU BEGIN❗**

Happy to move this to Discord if you'd prefer — filing here since it's a reproducible crash on the default `evaluate()` path with a proposed fix, and it turns out to be the other half of a previously closed issue (#1201), so it seemed worth having on the tracker.

**Describe the bug**

`copy_metrics` raises `TypeError` for any metric that subclasses a built-in with a narrower `__init__`, aborting the entire async `evaluate()` run before a single metric executes.

Subclassing a built-in metric to pre-fill its configuration is the natural way to build a reusable custom metric — it's how `GEval` is meant to be wrapped so a team can share one `Correctness` metric instead of re-typing `name`/`criteria`/`evaluation_params` at every call site. Any such subclass crashes.

`copy_metrics` reconstructs each metric by scraping `vars(metric)` and replaying it through the constructor, collecting candidate kwargs from **every class in the MRO** ([`deepeval/metrics/utils.py#L101-L121`](https://github.com/confident-ai/deepeval/blob/main/deepeval/metrics/utils.py#L101-L121)):

```python
superclasses = metric_class.__mro__

valid_params = []
for superclass in superclasses:
    signature = inspect.signature(superclass.__init__)
    superclass_params = signature.parameters.keys()
    valid_params.extend(superclass_params)
valid_params = set(valid_params)
valid_args = {key: args[key] for key in valid_params if key in args}

copied_metrics.append(metric_class(**valid_args))   # <-- parent's kwargs, child's __init__
```

The kwargs are gathered from the parent classes but applied to the **subclass's** `__init__`. When the subclass deliberately narrows its signature — the entire point of subclassing to pre-configure — every parent-only parameter becomes an unexpected keyword argument.

`a_execute_test_cases` calls `copy_metrics` per test case while scheduling, outside any per-metric error handling, so the `TypeError` propagates out of `evaluate()` itself:

- `deepeval/evaluate/execute/e2e.py` — lines 402, 470, 493, 779
- `deepeval/evaluate/execute/loop.py` — line 987

**To Reproduce**

1. `pip install deepeval==4.1.7`
2. Save the following as `repro.py`. It uses a deterministic metric, so no LLM and no API key are involved:

```python
from deepeval import evaluate
from deepeval.metrics import ToolPermissionMetric


class NoShellMetric(ToolPermissionMetric):
    """Reusable policy metric: a team-wide 'never call shell' gate."""

    def __init__(self, threshold: float = 1.0):
        super().__init__(denied_tools=["shell"], threshold=threshold)


test_case = LLMTestCase(
    input="hi",
    actual_output="hello",
    tools_called=[ToolCall(name="search")],
)

evaluate(test_cases=[test_case], metrics=[NoShellMetric()])
```

3. `python repro.py`
4. See error:

```
✨ You're running DeepEval's latest Tool Permission Metric! (using None, strict=False, async_mode=True)...
Evaluating 1 test case(s) in parallel ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━   0% 0:00:00
  File "repro.py", line 19, in <module>
    evaluate(test_cases=[test_case], metrics=[NoShellMetric()])
  File "...\site-packages\deepeval\evaluate\evaluate.py", line 217, in evaluate
    test_results = loop.run_until_complete(
                   ^^^^^^^^^^^^^^^^^^^^^^^^
  File "...\Python312\Lib\asyncio\base_events.py", line 687, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "...\site-packages\deepeval\evaluate\execute\e2e.py", line 402, in a_execute_test_cases
    copied_llm_metrics: List[BaseMetric] = copy_metrics(
                                           ^^^^^^^^^^^^^
  File "...\site-packages\deepeval\metrics\utils.py", line 120, in copy_metrics
    copied_metrics.append(metric_class(**valid_args))
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: NoShellMetric.__init__() got an unexpected keyword argument 'denied_tools'
```

Note the progress bar: the run has already started and reported `0%` when it dies, so this is not a construction-time validation error the user could catch up front.

**The keyword named in the error is different on every run.** `valid_params` is a `set` (line 117), so which offending kwarg trips first depends on string hash randomization. Consecutive runs of the same script, same machine, same install:

```
TypeError: ... unexpected keyword argument 'allowed_tools'
TypeError: ... unexpected keyword argument 'include_reason'
TypeError: ... unexpected keyword argument 'strict_mode'
TypeError: ... unexpected keyword argument 'denied_tools'
```

This makes the failure read as flaky and defeats searching for the error text.

The same failure occurs with a `GEval` subclass, which is the case most users will hit (needs an evaluation model configured, otherwise model init fails first):

```python
class MyCorrectness(GEval):
    def __init__(self, threshold: float = 0.5, model=None):
        super().__init__(
            name="My Correctness",
            criteria="Is the output correct?",
            evaluation_params=[
                SingleTurnParams.INPUT,
                SingleTurnParams.ACTUAL_OUTPUT,
            ],
            threshold=threshold,
            model=model,
        )

evaluate(test_cases=[test_case], metrics=[MyCorrectness()])
# TypeError: MyCorrectness.__init__() got an unexpected keyword argument 'criteria'
#            (or 'evaluation_params', or 'name', ... varies per run)
```

**Expected behavior**

A metric subclass that constructs and measures correctly should evaluate correctly, on both the sync and async paths.

**Screenshots**

N/A — terminal output included above.

- OS: Windows 11
- Python: 3.12.5
- deepeval: 4.1.7, from a stock `pip install` (not a source checkout). `copy_metrics` is byte-identical on `main`.
- Not platform-specific — the code path has no OS-dependent behavior.

**Smartphone (please complete the following information):**

N/A — Python library.

**Additional context**

### Scope

| Path | Result |
|---|---|
| `evaluate(...)` (default, `run_async=True`) | **TypeError**, run aborts |
| `evaluate(..., error_config=ErrorConfig(ignore_errors=True))` | **TypeError**, run aborts — `ignore_errors` does not cover it |
| `evaluate(..., async_config=AsyncConfig(run_async=False))` | works, scores correctly |

The sync path never calls `copy_metrics`, which confirms the metric itself is fine and isolates the fault to the copy.

Two things make this hard to diagnose from the user side: the error names a keyword the user never passed and never mentions copying, and the workaround (`run_async=False`) silently costs the parallelism that the default promises.

### This is the other half of #1201

The MRO walk was introduced by #1202 (merged 2024-12-03) to fix #1201, where a `GEval` subclass taking `**kwargs` silently lost its custom model under `run_async=True` and evaluated with `gpt-4o` instead. Before that, `valid_params` came from the subclass signature alone:

```python
signature = inspect.signature(metric_class.__init__)   # pre-#1202
valid_params = signature.parameters.keys()
```

The introducing commit is titled `Temp solution` (`1cf02c8b4`), which reads like it was known to be provisional.

The two parameter sets fail on opposite subclass shapes, and no choice between them works:

| `valid_params` source | `**kwargs` subclass (#1201) | narrowed-signature subclass (this issue) |
|---|---|---|
| Subclass signature only (pre-#1202) | **crash** — `GEval.__init__() missing 1 required positional argument: 'name'` | OK |
| Full MRO (current) | OK | **crash** — `unexpected keyword argument 'evaluation_steps'` |

So this isn't a matter of tuning which parameters get collected — reverting or narrowing the set just trades this bug back for #1201. The reconstruct-through-`__init__` strategy is the thing that doesn't hold: it assumes every metric's constructor signature mirrors its attributes, which is true of the built-ins and of nothing a user writes.

<details>
<summary>Script reproducing both cells of that table</summary>

```python
import inspect
from deepeval.metrics import GEval
from deepeval.metrics.utils import copy_metrics
from deepeval.test_case import SingleTurnParams


class FakeLLM(DeepEvalBaseLLM):
    def __init__(self): super().__init__("my-custom-llm")
    def load_model(self): return None
    def generate(self, prompt, schema=None): return "{}", 0.0
    async def a_generate(self, prompt, schema=None): return "{}", 0.0
    def get_model_name(self): return "my-custom-llm"


class KwargsMetric(GEval):          # the shape from #1201
    def __init__(self, **kwargs):
        super().__init__(**kwargs)


class NarrowMetric(GEval):          # the shape in this issue
    def __init__(self, threshold: float = 0.5, model=None):
        super().__init__(
            name="Narrow", criteria="Is it correct?",
            evaluation_params=[SingleTurnParams.INPUT, SingleTurnParams.ACTUAL_OUTPUT],
            threshold=threshold, model=model,
        )


def old_copy(metric):
    """copy_metrics as it was before #1202 (subclass signature only)."""
    cls, args = type(metric), vars(metric)
    valid = inspect.signature(cls.__init__).parameters.keys()
    return cls(**{k: args[k] for k in valid if k in args})


a = KwargsMetric(
    name="Kwargs", criteria="Is it correct?",
    evaluation_params=[SingleTurnParams.INPUT, SingleTurnParams.ACTUAL_OUTPUT],
    model=FakeLLM(),
)
b = NarrowMetric(model=FakeLLM())

for label, fn in (("CURRENT (MRO-wide)", lambda m: copy_metrics([m])[0]),
                  ("OLD (subclass-only)", old_copy)):
    print("==", label)
    for name, m in (("A **kwargs", a), ("B narrowed", b)):
        try:
            c = fn(m)
            print(f"   {name}: OK -> copy uses model {c.model.get_model_name()!r}")
        except Exception as e:
            print(f"   {name}: CRASH -> {type(e).__name__}: {e}")
```

```
== CURRENT (MRO-wide)
   A **kwargs: OK -> copy uses model 'my-custom-llm'
   B narrowed: CRASH -> TypeError: NarrowMetric.__init__() got an unexpected keyword argument 'evaluation_steps'
== OLD (subclass-only)
   A **kwargs: CRASH -> TypeError: GEval.__init__() missing 1 required positional argument: 'name'
   B narrowed: OK -> copy uses model 'my-custom-llm'
```

The keyword named in the `B narrowed` crash changes between runs, as noted above.

</details>

### Suggested fix

Since no parameter set satisfies both subclass shapes, the fix has to stop routing the copy through `__init__` at all. That also removes a quieter loss in the current approach: any state a metric sets outside `__init__` is dropped from the copy, so a custom metric doing setup elsewhere gets a subtly different object per test case.

Proposal — add `clone()` to `BaseMetric` / `BaseConversationalMetric`.

It has to be a **shallow** copy, not a deepcopy. `copy.deepcopy` on a metric holding a real model fails outright, because the evaluation model owns an HTTP client:

```python
>>> copy.deepcopy(AnswerRelevancyMetric())
TypeError: cannot pickle '_thread.RLock' object
```

The current implementation shares the model object with every copy by reference (`copy.model is original.model` → `True`), and that behavior is worth preserving deliberately: one client and one connection pool across concurrent test cases, not N.

```python
_RUN_STATE = (
    "score", "reason", "success", "error", "verdicts", "skipped",
    "evaluation_cost", "input_tokens", "output_tokens", "verbose_logs",
)

def clone(self) -> "BaseMetric":
    copied = copy.copy(self)                    # model / client stay shared

    # isolate mutable config containers so concurrent runs can't alias them
    for key, value in vars(copied).items():
        if key not in _RUN_STATE and isinstance(value, (list, dict, set)):
            setattr(copied, key, copy.copy(value))

    for key in _RUN_STATE:                      # fresh per-test-case run state
        if hasattr(copied, key):
            setattr(copied, key, None)

    return copied
```

and have `copy_metrics` call `metric.clone()`. This preserves the full configuration regardless of constructor shape — fixing the narrowed-signature case without regressing the `**kwargs` case that #1202 fixed — keeps the fresh-per-test-case run state that the current code exists to provide, and gives any metric with unusual copy semantics a documented hook to override.

Measured at ~0.04 ms per clone, against `inspect.signature` over the full MRO on every test case today.

Keeping the existing reconstruction as a fallback if `clone()` raises would make this fully backward compatible.

I'd want a regression test covering both subclass shapes from the table above, so the two failure modes can't be traded for each other again.

Happy to open a PR for this if the approach sounds right.

jstar0:
## Summary
- Replace constructor reconstruction with base metric clone semantics.
- Preserve metric configuration and shared model clients while resetting per-run state.
- Cover narrowed, **kwargs, and conversational metric subclasses.

Fixes #3037

## Testing
- `pytest -q tests/test_core/test_evaluation/test_metric_data.py --disable-warnings`
- `pytest -q tests/test_core/test_evaluation --disable-warnings`
- `black --check deepeval/metrics/base_metric.py deepeval/metrics/utils.py tests/test_core/test_evaluation/test_metric_data.py`

### DIFF
--- a/deepeval/metrics/base_metric.py
+++ b/deepeval/metrics/base_metric.py
@@ -1,22 +1,51 @@
 from __future__ import annotations
 
+import copy
 from abc import abstractmethod
-from typing import TYPE_CHECKING, Optional, Dict, List
+from typing import TYPE_CHECKING, Dict, List, Optional
 
-from deepeval.test_case import (
-    LLMTestCase,
-    ConversationalTestCase,
-    SingleTurnParams,
-    ArenaTestCase,
-)
 from deepeval.templates.resolver import (
     MetricTemplateMethod,
     resolve_template,
 )
 from deepeval.templates.template_class import filter_template_kwargs
+from deepeval.test_case import (
+    ArenaTestCase,
+    ConversationalTestCase,
+    LLMTestCase,
+    SingleTurnParams,
+)
 
 if TYPE_CHECKING:
     from deepeval.models import DeepEvalBaseLLM
+
+
+_RUN_STATE = {
+    "score",
+    "reason",
+    "success",
+    "error",
+    "verdicts",
+    "skipped",
+    "evaluation_cost",
+    "input_tokens",
+    "output_tokens",
+    "verbose_logs",
+}
+
+
+def _clone_metric(metric):
+    copied = copy.copy(metric)
+
+    for key, value in vars(copied).items():
+        if key not in _RUN_STATE and isinstance(value, (list, dict, set)):
+            setattr(copied, key, copy.copy(value))
+
+    for key in _RUN_STATE:
+        if hasattr(copied, key):
+            setattr(copied, key, None)
+
+    return copied
 
 
 class PromptMixin:
@@ -73,6 +102,10 @@ class BaseMetric(PromptMixin):
     requires_trace: bool = False
     model: Optional[DeepEvalBaseLLM] = None
     using_native_model: Optional[bool] = None
+
+    def clone(self) -> "BaseMetric":
+        """Return a per-test-case copy without duplicating the model client."""
+        return _clone_metric(self)
 
     def __init_subclass__(cls, **kwargs):
         super().__init_subclass__(**kwargs)
@@ -144,6 +177,10 @@ class BaseConversationalMetric(PromptMixin):
     flaky: bool = False
     model: Optional[DeepEvalBaseLLM] = None
     using_native_model: Optional[bool] = None
+
+    def clone(self) -> "BaseConversationalMetric":
+        """Return a per-test-case copy without duplicating the model client."""
+        return _clone_metric(self)
 
     def __init_subclass__(cls, **kwargs):
         super().__init_subclass__(**kwargs)
--- a/deepeval/metrics/utils.py
+++ b/deepeval/metrics/utils.py
@@ -1,4 +1,3 @@
-import inspect
 import json
 import re
 import sys
@@ -101,24 +100,7 @@ def check_at_least_one_metric_has_threshold(
 def copy_metrics(
     metrics: List[Union[BaseMetric, BaseConversationalMetric]],
 ) -> List[Union[BaseMetric, BaseConversationalMetric]]:
-    copied_metrics = []
-    for metric in metrics:
-        metric_class = type(metric)
-        args = vars(metric)
-
-        superclasses = metric_class.__mro__
-
-        valid_params = []
-
-        for superclass in superclasses:
-            signature = inspect.signature(superclass.__init__)
-            superclass_params = signature.parameters.keys()
-            valid_params.extend(superclass_params)
-        valid_params = set(valid_params)
-        valid_args = {key: args[key] for key in valid_params if key in args}
-
-        copied_metrics.append(metric_class(**valid_args))
-    return copied_metrics
+    return [metric.clone() for metric in metrics]
 
 
 def format_turns(
--- a/tests/test_core/test_evaluation/test_metric_data.py
+++ b/tests/test_core/test_evaluation/test_metric_data.py
@@ -6,6 +6,7 @@ from deepeval.metrics import (
     BaseConversationalMetric,
     BaseMetric,
     ExactMatchMetric,
+    ToolPermissionMetric,
 )
 from deepeval.metrics.utils import (
     check_at_least_one_metric_has_threshold,
@@ -27,6 +28,28 @@ class StubMetric(BaseMetric):
 
 class StubConversationalMetric(BaseConversationalMetric):
     threshold = 0.5
+
+    def measure(self, test_case, *args, **kwargs):
+        return 1
+
+    async def a_measure(self, test_case, *args, **kwargs):
+        return 1
+
+
+class NarrowToolPermissionMetric(ToolPermissionMetric):
+    def __init__(self, threshold=1.0):
+        super().__init__(denied_tools=["shell"], threshold=threshold)
+
+
+class KwargsExactMatchMetric(ExactMatchMetric):
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
+
+class NarrowConversationalMetric(BaseConversationalMetric):
+    def __init__(self, threshold=0.5):
+        self.threshold = threshold
+        self.custom_config = {"labels": ["conversation"]}
 
     def measure(self, test_case, *args, **kwargs):
         return 1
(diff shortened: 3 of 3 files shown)


rajeshl8 (commented):
Quick question on the clone semantics: when a user subclasses a metric and sets a custom threshold in __init__ (e.g. self.threshold = 0.8), does the new clone approach preserve that value correctly across runs? The PR description mentions "preserve metric configuration" — worth adding a test case specifically for a subclass that sets a non-default threshold, to guard against it silently resetting to the base class default on clone.

jstar0:
Addressed the review request in `eddfe2581`.

Added `FixedThresholdMetric`, a custom subclass that sets `threshold = 0.8` inside its constructor, and asserted that cloning preserves the configured threshold while still resetting per-run state.

Local verification:
- `pytest -q tests/test_core/test_evaluation/test_metric_data.py --disable-warnings` (20 passed)
- `ruff check` on all changed Python files passed
- `black --check` on all changed Python files passed

The updated head is pushed and CI is running. The Vercel authorization check is an external team permission and remains independent of this code change.

rajeshl8:
The FixedThresholdMetric test covers exactly the case I was asking about — threshold preserved, per-run state reset independently. That's the right split. 20 passing with the new coverage looks good to me.

jstar0:
Thanks @rajeshl8 — glad the subclass-threshold test covers the concern. The head is unchanged since your review (eddfe258), so whenever you have a moment, a final approval or merge would be much appreciated. Happy to make any further adjustments if anything else comes up.

PR #3039, opened 2026-08-11
fix: preserve zero-valued GenAI token usage

Boulea7:
## Summary

- Preserve explicit zero values from modern OpenTelemetry GenAI input and output token attributes.
- Fall back to legacy prompt and completion token attributes only when the modern attribute is absent.
- Apply the same behavior to the AgentCore and Strands span interceptors.

## Validation

- `pytest tests/test_integrations/test_strands/test_span_interceptor.py tests/test_integrations/test_agentcore/test_span_interceptor.py tests/test_integrations/test_genai_token_usage.py -q` — 84 passed with a synthetic environment value and no external API calls.
- The same test set repeated three times — 252 passed.
- `black --check` on the three changed files.
- `ruff check` on the three changed files.
- `git diff --check`.

### DIFF
--- a/deepeval/integrations/agentcore/instrumentator.py
+++ b/deepeval/integrations/agentcore/instrumentator.py
@@ -883,12 +883,12 @@ class AgentCoreSpanInterceptor(SpanProcessor):
                     span, "confident.trace.output", output_text
                 )
 
-        input_tokens = attrs.get("gen_ai.usage.input_tokens") or attrs.get(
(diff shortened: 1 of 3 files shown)


Boulea7:
CI context: all 12 test and title checks completed successfully. The Black job reports only `.scripts/release.py`, which this PR does not modify; Black 25.12.0 passes all three changed files, while the same version reproduces the `.scripts/release.py` failure on unchanged `main`. Vercel is awaiting team authorization.

PR #3043, opened 2026-08-12
fix(models): raise AnthropicModel default max_tokens to 8192 and surface truncation clearly

### LINKED ISSUE #3042: AnthropicModel default max_tokens=1024 truncates thinking-by-default Claude judges (claude-opus-5 default), surfacing as invalid-JSON metric errors
## Describe the bug

`AnthropicModel` defaults `max_tokens` to 1024 (`deepeval/models/llms/anthropic_model.py`, `default_max_tokens = 1024`, used by both `generate()` and `a_generate()`).

That default was sized for models where 1024 tokens of output is plenty for a judge verdict. It no longer holds for the current default judge. Since #2993 / the default-model update, `DEFAULT_ANTHROPIC_MODEL` is `claude-opus-5`, and on `claude-opus-5` (also `claude-sonnet-5`) **thinking runs by default when the `thinking` parameter is omitted**; on `claude-fable-5` it is always on. `AnthropicModel` sends no `thinking` parameter. Per Anthropic's API semantics, `max_tokens` is a hard cap on **thinking plus response text combined**.

So whenever the judge model decides to think (judge/verdict prompts trigger this often), the thinking spend eats into the 1024 budget and the response is cut off with `stop_reason: "max_tokens"`. The text block is then empty or contains truncated JSON, and the failure surfaces far from the cause:

- `trim_and_load_json(...)` raises "Evaluation LLM outputted an invalid JSON" (or `schema.model_validate` fails), so the metric errors intermittently with what looks like a model-quality problem.
- The thinking tokens are still billed, so users pay for output they never receive.

This is distinct from #2946 (`content[0].text` assumes the first block is a `TextBlock` — a block *extraction* bug, fix in #2947). Even with by-type extraction in place, the *budget* problem remains: the text block that gets extracted is truncated or empty.

## To Reproduce

```python

metric = AnswerRelevancyMetric(model=AnthropicModel())  # claude-opus-5, max_tokens=1024
metric.measure(LLMTestCase(input="...", actual_output="..."))
# intermittently: "Evaluation LLM outputted an invalid JSON" once the judge
# thinks for ~1k tokens and the JSON verdict is cut off mid-string
```

Direct API demonstration: a `claude-opus-5` `messages.create` call with `max_tokens=1024` and no `thinking` parameter returns `stop_reason: "max_tokens"` on prompts that induce reasoning, with the visible text truncated.

## Expected behavior

1. The default output budget should account for thinking-by-default models. `max_tokens` is a ceiling, not a spend — raising the default costs nothing on responses that stay small.
2. When a response does get truncated (`stop_reason == "max_tokens"`) and structured output was requested, the error should say so and point at `max_tokens`, instead of a generic invalid-JSON message.

I have a small PR ready for both (default 1024 → 8192, explicit user values still win; plus a clear `DeepEvalError` on truncation).

## Environment

- deepeval @ current `main` (be9cac54f)
- anthropic SDK ≥ 0.7x, model `claude-opus-5`

lokesh75-kank:
## Summary

Fixes #3042.

`AnthropicModel` defaulted `max_tokens` to 1024. On current Claude models — including `claude-opus-5`, the default judge — thinking runs by default when the `thinking` parameter is omitted, and the API's `max_tokens` caps **thinking + response text combined**. Judge prompts routinely trigger thinking, so the visible output got starved: responses stopped with `stop_reason: "max_tokens"`, the JSON verdict came back truncated or empty, and the failure surfaced downstream as a generic "invalid JSON" metric error while the thinking tokens were still billed.

## Changes

- **Default `max_tokens`: 1024 → 8192.** `max_tokens` is a ceiling, not a spend — responses that stay small cost the same. Explicit values from the constructor or `generation_kwargs` still win, unchanged.
- **Clear truncation error.** `generate()` / `a_generate()` now raise a `DeepEvalError` naming `max_tokens` (and the thinking-budget interaction) when a response is truncated and structured output was requested, or when no text block came back at all. Plain-text callers who deliberately cap output keep their partial text.

Related but distinct from #2946 / #2947, which fix *block extraction* (`content[0].text` vs a thinking-first content list). This PR fixes the token *budget*; both are needed for thinking-capable judges to work end to end, and the diffs don't overlap.

## Testing

- 6 new tests in `tests/test_core/test_models/test_anthropic_model.py`: default is 8192; constructor and `generation_kwargs` overrides win; truncation with a schema raises the clear error (sync + async); truncation without a schema keeps partial text; truncation with no text block raises. Duck-typed stubs, no `anthropic` package or network needed.
- `pytest tests/test_core/test_models/` — 214 passed (one pre-existing, unrelated ollama-embedding failure also fails on clean `main` locally).
- `black --check` clean.


penguine-ip:
Thank you for this — and in particular for handling the case where the entire budget goes to reasoning and no text block comes back at all, which we carried into our fix.

We've resolved #3042 in a separate commit, taking the other route we considered: rather than raising the ceiling to absorb the thinking spend, we now send Anthropic's `thinking` parameter explicitly and default it to disabled, so the judge's whole budget goes to the verdict. That comes with a new `DEEPEVAL_MODEL_THINKING` env var (off unless set) for anyone who does want their judge to reason, a `supports_thinking` flag in the model catalog so the parameter is only sent to models that accept it, and by-type text extraction since a thinking response puts its reasoning in `content[0]`. Both SDKs are covered.

The reason we went that way rather than merging this: a flat 8192 default would have broken the `claude-3-opus` and `claude-3-haiku` entries still in our registry, which cap output at 4096 and reject a larger `max_tokens` outright.

Closing as superseded, with credit in the changelog. One unrelated note for future PRs: this branch also reformatted `.scripts/release.py`, which made the diff harder to review than the actual change warranted — worth keeping formatting-only churn in its own commit.

lokesh75-kank:
Superseded by a better fix is a good outcome. You're right about the 4096-cap models; a flat 8192 default was the wrong shape for a registry that still carries them, and defaulting thinking off is the cleaner economics for a judge anyway.

Read the commit — nice to see the thinking-enabled path get the 8192 default with the headroom check, so the truncation failure can't quietly return for anyone who flips `DEEPEVAL_MODEL_THINKING` on. Thanks for carrying the empty-response case through and for the changelog credit.

Point taken on the `release.py` churn; I'll keep formatting-only fixes in their own PR from now on.

PR #3049, opened 2026-08-12
fix(dataset): align CSV context delimiters with save_as output

Koustav-github:
## Summary

`save_as("csv")` writes `context` and `retrieval_context` joined by `|`, but `add_test_cases_from_csv_file` defaults both delimiters to `;`. Reloading a CSV that deepeval itself wrote silently collapses the list into a single bogus element — no exception, no warning:

```python
import tempfile

golden = Golden(input="q", context=["Geography", "Mountains"])

with tempfile.TemporaryDirectory() as tmpdir:
    path = EvaluationDataset([golden]).save_as("csv", tmpdir, file_name="rt")
    # deepeval writes:  context = "Geography|Mountains"

    a = EvaluationDataset()
    a.add_test_cases_from_csv_file(path, "input", "actual_output")

    b = EvaluationDataset()
    b.add_goldens_from_csv_file(file_path=path)

    print(a.test_cases[0].context)   # ['Geography|Mountains']   <- wrong
    print(b.goldens[0].context)      # ['Geography', 'Mountains'] <- correct
```

Two loaders, one file, different results. No exception in either case.

Every RAG metric downstream then scores against one malformed chunk instead of two.

## Why `|` is the correct side to change

`add_test_cases_from_csv_file` is the only place in the codebase using `;` for these two fields:

| Location | context / retrieval_context delimiter |
|---|---|
| `save_as("csv")` writer (3 call sites) | `\|` |
| `join_retrieval_context()` — `dataset/utils.py` | `\|` |
| `add_goldens_from_csv_file` | `\|` |
| `add_goldens_from_jsonl_file` | `\|` |
| `tests/test_core/test_datasets/goldens.csv` fixture | `\|` |
| **`add_test_cases_from_csv_file`** | **`;`** |

`tools_called` and `expected_tools` use `;` consistently across every loader and are left untouched — the inconsistency is confined to exactly the two fields that break.

## Changes

- `deepeval/dataset/dataset.py`: default `context_col_delimiter` and `retrieval_context_col_delimiter` to `|` in `add_test_cases_from_csv_file`
- same file: the docstring documented two parameters that don't exist (`context_delimiter`, `retrieval_context_delimiter` — the real names are `*_col_delimiter`) and stated the old defaults; both corrected
- `docs/.../evaluation-datasets.mdx`: the example passed `;` explicitly, demonstrating a delimiter no deepeval writer emits
- regression test asserting the `save_as` → `add_test_cases_from_csv_file` round trip for both fields

## Behavior change

This changes a default. Anyone relying on the implicit `;` when loading their own `;`-delimited CSVs will need to pass `context_col_delimiter=";"` explicitly.

Flagging it rather than burying it — though the current default silently corrupts data written by deepeval's own `save_as`, fails without any error, and disagrees with every other loader in the codebase.

## Testing

```
pytest tests/test_core/test_datasets/     ->  22 passed
pytest tests/test_core/                   ->  1241 passed, 9 failed, 56 skipped
```

The 9 failures are pre-existing and unrelated (Windows path/encoding, chromadb, telemetry, ollama) — verified identical on a clean tree with these changes stashed. Two of them are what #2906 describes.

Run on Python 3.12.5 / Windows 11. black could not be run locally (it refuses on 3.12.5 due to a known AST-safety bug); line lengths were checked manually against the repo's line-length = 80 and no added code line exceeds it.

The new test fails without the source change:

```
AssertionError: assert ['c1|c2'] == ['c1', 'c2']
```

### DIFF
--- a/deepeval/dataset/dataset.py
+++ b/deepeval/dataset/dataset.py
@@ -265,9 +265,9 @@ class EvaluationDataset:
         actual_output_col_name: str,
         expected_output_col_name: Optional[str] = "expected_output",
         context_col_name: Optional[str] = "context",
-        context_col_delimiter: str = ";",
+        context_col_delimiter: str = "|",
         retrieval_context_col_name: Optional[str] = "retrieval_context",
-        retrieval_context_col_delimiter: str = ";",
+        retrieval_context_col_delimiter: str = "|",
         tools_called_col_name: Optional[str] = "tools_called",
         tools_called_col_delimiter: str = ";",
         expected_tools_col_name: Optional[str] = "expected_tools",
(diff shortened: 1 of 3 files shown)


Koustav-github:
Hi, just checking in on this PR when you get a chance — happy to make any changes if needed. Thanks!

Koustav-github:
Hi, it has been quite a long time, just checking in, this PR does not have an issue, i raised the PR as this was a very minor bug, so, i thought it wont need an issue, would love to hear back from u. 
Thanks.

penguine-ip:
@Koustav-github thank you!

Issue #2993, opened 2026-08-02
Default AnthropicModel uses a nonexistent model ID (claude-sonnet-4-6-20250514) — every default-configured evaluation 404s


lokesh75-kank:
## Describe the bug

`default_model` in `deepeval/models/llms/anthropic_model.py:31` is `"claude-sonnet-4-6-20250514"`. That model ID does not exist: starting with the 4.6 generation, Anthropic publishes **dateless pinned IDs only** (`claude-sonnet-4-6`, `claude-opus-4-6`) — see the [model IDs docs](https://platform.claude.com/docs/en/docs/about-claude/models/overview). The date `20250514` is Claude Sonnet 4.0's release date, seemingly copied from `claude-sonnet-4-20250514`.

Because the fabricated ID *is* present in `ANTHROPIC_MODELS_DATA` (`constants.py:556`), construction succeeds silently and the failure only surfaces as a `404 not_found_error` on the first `messages.create` call — so anyone running `evaluate()` or any metric with a default-configured Anthropic judge gets a hard API failure.

Two more registry entries have the same problem:
- `claude-opus-4-6-20250610` (`constants.py:548`) — no dated Opus 4.6 ID exists
- `claude-opus-4-5-20251124` (`constants.py:540`) — the real Opus 4.5 snapshot is `claude-opus-4-5-20251101`

Introduced in #2584.

## To Reproduce

No API key needed — the Anthropic SDK's published model list shows no dated 4.6 IDs:

```python
import typing
import anthropic.types.model as m
from deepeval.models.llms.anthropic_model import default_model

literal, _ = typing.get_args(m.Model)
published = set(typing.get_args(literal))
print(default_model)               # claude-sonnet-4-6-20250514
print(default_model in published)  # False
print(sorted(x for x in published if "4-6" in x))
# ['claude-opus-4-6', 'claude-sonnet-4-6']  — dateless only
```

With a key: `AnthropicModel().generate("hi")` → `anthropic.NotFoundError: model: claude-sonnet-4-6-20250514`.

## Expected behavior

The default judge model is a valid Anthropic model ID, and the registry contains only real IDs (the dateless `claude-sonnet-4-6` / `claude-opus-4-6` entries already exist with correct pricing).

## Environment

- deepeval: current `main`
- anthropic SDK: 0.120.2
- Python 3.13, macOS

PR with the fix + regression test incoming.

PR #3054, opened 2026-08-13
fix: raise AnthropicModel default max_tokens and surface truncation clearly

### LINKED ISSUE #3042: AnthropicModel default max_tokens=1024 truncates thinking-by-default Claude judges (claude-opus-5 default), surfacing as invalid-JSON metric errors

harshasiddartha:
Fixes #3042.

`AnthropicModel` defaulted `max_tokens` to 1024. Anthropic's `max_tokens` caps **thinking plus response text** together, and the current default judge (`claude-opus-5`, since #2993) thinks by default when no `thinking` parameter is sent — which `AnthropicModel` never sends. Judge and verdict prompts induce reasoning often, so the thinking spend eats into the 1024 budget and the verdict comes back cut off with `stop_reason: "max_tokens"`.

The failure surfaced far from its cause: `trim_and_load_json` raised *"Evaluation LLM outputted an invalid JSON"*, which reads like a model-quality problem rather than a budget one, and the thinking tokens were billed regardless.

## Changes

**1. Default `max_tokens` 1024 → 8192** (`DEFAULT_MAX_TOKENS`).

`max_tokens` is a ceiling, not a spend — raising it costs nothing on responses that stay small. Explicit values still win, whether passed via `generation_kwargs={"max_tokens": ...}` or `max_tokens=...`.

**2. A clear error when a structured response is truncated.**

When `stop_reason == "max_tokens"` and a `schema` was requested, `generate`/`a_generate` now raise a `DeepEvalError` naming `max_tokens` and the budget that was hit, instead of handing half-written JSON to the parser. Free-form text (no schema) is deliberately left alone — a truncated answer is still an answer.

## Relationship to #2946 / #2947

Distinct issue. #2946 is about `content[0].text` assuming the first block is a `TextBlock` — a block *extraction* bug. This one is about the *budget*: even with by-type extraction in place, the text block that gets extracted is truncated or empty. This PR doesn't touch the `content[0].text` lines, so it should merge cleanly alongside #2947.

## Tests

Added to `tests/test_core/test_models/test_anthropic_model.py`, all offline against a stub client:

- default budget is applied and exceeds the old 1024
- an explicit `generation_kwargs["max_tokens"]` still wins
- a truncated structured response raises an error naming `max_tokens` — sync and async
- a truncated free-form response is returned as-is

Verified the two truncation tests fail against the unfixed code path and pass with it. The full file passes: `15 passed`. Formatted with `black --line-length 79` to match the repo.

### DIFF
--- a/deepeval/models/llms/anthropic_model.py
+++ b/deepeval/models/llms/anthropic_model.py
@@ -27,6 +27,13 @@ from deepeval.models.llms.constants import (
 # consistent retry rules
 retry_anthropic = create_retry_decorator(PS.ANTHROPIC)
 
+# Anthropic's ``max_tokens`` caps thinking *plus* response text. The current
+# default judge (``claude-opus-5``) thinks by default when no ``thinking``
+# parameter is sent, so a budget sized for the response alone gets eaten by
+# reasoning and the verdict is cut off mid-JSON. This is a ceiling, not a spend:
+# raising it costs nothing on responses that stay small.
+DEFAULT_MAX_TOKENS = 8192
+
 _ALIAS_MAP = {
     "api_key": ["_anthropic_api_key"],
 }
@@ -118,7 +125,9 @@ class AnthropicModel(DeepEvalBaseLLM):
         self.generation_kwargs.pop(
             "temperature", None
         )  # to avoid duplicate with self.temperature
-        default_max_tokens = 1024 if max_tokens is None else max_tokens
+        default_max_tokens = (
+            DEFAULT_MAX_TOKENS if max_tokens is None else max_tokens
+        )
         self._max_tokens = int(
             self.generation_kwargs.pop("max_tokens", default_max_tokens)
         )
@@ -160,6 +169,7 @@ class AnthropicModel(DeepEvalBaseLLM):
         ):
             create_kwargs["temperature"] = self.temperature
         message = chat_model.messages.create(**create_kwargs)
+        self._raise_if_truncated(message, max_tokens, schema)
         cost = self.calculate_cost(
             message.usage.input_tokens, message.usage.output_tokens
         )
@@ -200,6 +210,7 @@ class AnthropicModel(DeepEvalBaseLLM):
         ):
             create_kwargs["temperature"] = self.temperature
         message = await chat_model.messages.create(**create_kwargs)
+        self._raise_if_truncated(message, max_tokens, schema)
         cost = self.calculate_cost(
             message.usage.input_tokens, message.usage.output_tokens
         )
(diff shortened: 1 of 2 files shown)


penguine-ip:
Thank you for this — the diagnosis was exactly right, and the writeup connecting `stop_reason: "max_tokens"` to the misleading "Evaluation LLM outputted an invalid JSON" is what made the root cause obvious.

We've resolved #3042 in a separate commit, taking the other route we considered: rather than raising the ceiling to absorb the thinking spend, we now send Anthropic's `thinking` parameter explicitly and default it to disabled, so the judge's whole budget goes to the verdict. That comes with a new `DEEPEVAL_MODEL_THINKING` env var (off unless set) for anyone who does want their judge to reason, a `supports_thinking` flag in the model catalog so the parameter is only sent to models that accept it, and by-type text extraction since a thinking response puts its reasoning in `content[0]`. Both SDKs are covered.

The reason we went that way rather than merging this: a flat 8192 default would have broken the `claude-3-opus` and `claude-3-haiku` entries still in our registry, which cap output at 4096 and reject a larger `max_tokens` outright.

Closing as superseded, but your analysis shaped the fix and you'll be credited in the changelog. The truncation error you added is still worth having on its own, so we're following up on it separately.

PR #3060, opened 2026-08-14
fix(g-eval): use final score token for logprob weighting

### LINKED ISSUE #3000: GEval logprobs fails if reason contains score token.

tunglambk:
## Summary

- select the final token matching the raw G-Eval score when reading logprobs
- add offline regression coverage for repeated and single score-token outputs

## Root cause

`calculate_weighted_summed_score()` scanned generated tokens from the beginning.
When the reasoning mentioned the same number as the final score, the function
used the reasoning token's probability distribution and could calculate an
unrelated weighted score.

Searching from the end selects the score token emitted at the end of the model
response while preserving the existing weighting behavior.

## Impact

Weighted G-Eval scores now use the probability distribution associated with the
model's final score rather than an earlier matching token in its reasoning.
There are no public API or schema changes.

## Validation

- `poetry run pytest -q tests/test_metrics/test_g_eval_utils.py` — 7 passed
- `poetry run black --check deepeval/metrics/g_eval/utils.py tests/test_metrics/test_g_eval_utils.py` — passed
- `poetry run ruff check tests/test_metrics/test_g_eval_utils.py` — passed
- `poetry run ruff check --ignore F811 deepeval/metrics/g_eval/utils.py` — passed (`F811` excludes pre-existing duplicate imports)

Fixes #3000

### DIFF
--- a/deepeval/metrics/g_eval/utils.py
+++ b/deepeval/metrics/g_eval/utils.py
@@ -339,9 +339,10 @@ def calculate_weighted_summed_score(
 ) -> Union[int, float]:
     try:
         generated_logprobs = raw_response.choices[0].logprobs.content
-        # First, locate the token that we care for logprobs, i.e., the token matching the score
+        # First, locate the final token matching the score. The reasoning may
+        # contain the same token before the model emits its final score.
         score_logprobs = None
-        for token_logprobs in generated_logprobs:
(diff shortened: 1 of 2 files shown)


penguine-ip:
@tunglambk thanks for fixing this

PR #3061, opened 2026-08-14
fix(metrics): honor include_reason and align sync/async reason generation

biztex:
**Describe the bug**

Two metrics ignore `include_reason=False`, and one of them also produces a different `reason` depending on `async_mode`:

1. `ConversationCompletenessMetric`: `_generate_reason()` returns `None` when `include_reason=False`, but its async twin `_a_generate_reason()` was missing the guard. Since `async_mode` defaults to `True` — and `measure()` delegates to the async path — the default configuration still makes the reason LLM call (and attaches a reason) that the user explicitly disabled.

2. `TopicAdherenceMetric`: `include_reason` is accepted by the constructor and stored, but never consulted anywhere, so both the sync and async paths always make the reason LLM call. Separately, when there are no question-answer pairs to evaluate (`total <= 0`), the sync `_generate_reason()` returns a canned message without prompting the model, but the async twin was missing that guard: it still made an LLM call with all-"(none)" reason lines, so the same conversation produces a different `reason` depending on `async_mode`.

**The fix**

- Add the `include_reason` early return to `ConversationCompletenessMetric._a_generate_reason`, mirroring the sync helper.
- Add the `include_reason` early return to both `TopicAdherenceMetric` reason helpers, matching the convention sibling metrics already follow (the turn-contextual and MCP metrics all guard inside the helper).
- Mirror the sync `total <= 0` canned-message guard into `TopicAdherenceMetric._a_generate_reason`.

**Tests**

Two offline regression test files driven through stub models (no LLM calls, no API keys), following the pattern of `tests/test_metrics/test_turn_faithfulness_metric_empty_verdicts.py`:

- `test_conversation_completeness_include_reason.py`: helper-level sync/async `include_reason=False` tests, plus a full default-configuration `measure()` test in which the stub answers the intention/verdict prompts but fails the test if the reason schema is ever requested.
- `test_topic_adherence_include_reason.py`: helper-level `include_reason=False` tests for both paths, and an empty-case parity test asserting the async helper returns the sync path's canned message without calling the model.

On current `main`, 5 of the 6 tests fail (the sync ConversationCompleteness control passes); all 6 pass with this change. The rest of `tests/test_metrics` is unchanged: 169 passed, 271 skipped, 6 xfailed. `black==25.12.0 --check` (the CI pin) is clean on all four touched files.

**Additional context**

A quick survey while writing this found several other metrics that store `include_reason` without consulting it (`GoalAccuracy`, `ToolUse`, `PlanAdherence`, `PlanQuality`, `TaskCompletion`, `StepEfficiency`, `ToolCorrectness`, and the DAG variants). For some of those the reason comes back in the same LLM response as the score, so skipping it is not the clear win it is for the two fixed here; happy to follow up separately if maintainers want the rest aligned too.

### DIFF
--- a/deepeval/metrics/conversation_completeness/conversation_completeness.py
+++ b/deepeval/metrics/conversation_completeness/conversation_completeness.py
@@ -171,6 +171,9 @@ class ConversationCompletenessMetric(BaseConversationalMetric):
             return self.score
 
     async def _a_generate_reason(self, multimodal: bool) -> str:
+        if self.include_reason is False:
+            return None
+
         incompletenesses: List[str] = []
         for verdict in self.verdicts:
             if (
--- a/deepeval/metrics/topic_adherence/topic_adherence.py
+++ b/deepeval/metrics/topic_adherence/topic_adherence.py
@@ -232,6 +232,9 @@ class TopicAdherenceMetric(BaseConversationalMetric):
             return self.score
 
     def _generate_reason(self, TP, TN, FP, FN, *, multimodal: bool):
+        if self.include_reason is False:
+            return None
+
         total = TP[0] + TN[0] + FP[0] + FN[0]
         if total <= 0:
             return "There were no question-answer pairs to evaluate. Please enable verbose logs to look at the evaluation steps taken"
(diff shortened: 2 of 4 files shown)


penguine-ip:
@biztex thanks!

PR #3056, opened 2026-08-13
feat(metrics): add deterministic trace alignment layer

Nickgonzales76017:
Draft alignment layer for #3019, split from the metric wrapper as agreed with @lostmartian.

This contributes the policy-neutral seam only:

- versioned projection for dicts, DeepEval `ToolCall`-shaped objects, and span-like objects
- deterministic `align(baseline, candidate) -> AlignmentResult`
- separate difference kind, recovery point, and metric-policy inputs
- conservative `order_change` reporting without inferring causal independence
- occurrence-aware unmatched IDs for duplicated or aliased steps
- fail-closed malformed names, non-JSON argument maps, and invalid alignment bounds

The metric wrapper, threshold, reason text, and final `ArenaTestCase` versus community-metric placement remain deliberately outside this commit. @lostmartian can wire against `AlignmentResult`; maintainer placement guidance can change the wrapper without rewriting this layer.

The retry/recovery and reorder semantics match the source-level Amberfork comparison recorded in #3019: a recovered difference remains observable, and a reorder is not waived without dependency evidence.

Verification:

- `uvx ruff check deepeval/metrics/community/trace_divergence tests/test_metrics/test_trace_divergence_alignment.py`
- `uvx black --check deepeval/metrics/community/trace_divergence tests/test_metrics/test_trace_divergence_alignment.py`
- `uv run --no-project --with-editable . --with pytest -- python -m pytest tests/test_metrics/test_trace_divergence_alignment.py -q` — 15 passed

This is draft because the wrapper ownership and repository placement question are still open; the alignment code and fixtures are ready for review.

Issue #3019, opened 2026-08-09
Community metric proposal: localize the divergence point between two agent execution traces (not just pass/fail)


Melvin0070:
## What exists today

Every agentic metric I can find (`ToolCorrectness`, `PlanAdherence`, `PlanQuality`, `TaskCompletion`, `GoalAccuracy`, `AgentLoopDetection`) scores **one real run against a static expectation** — `expected_tools`, an expected plan, a golden output. None of them compare **two real execution traces against each other**.

That's a different, common failure mode: "this run behaved differently from a run I trust" — after a prompt edit, a model swap, or a tool/version bump — where the useful answer isn't a pass/fail score, it's *"steps 1–4 matched, step 5 is where the tool call arguments diverged, everything after is downstream of that."*

I also want to flag #2948 (open PR, "local test run comparison CLI with regression reporting") since it's adjacent — it compares two **test suite runs** (which test cases passed/failed) at the aggregate level. This proposal is a level below that: comparing two traces *within* a single agent execution, step by step, to localize *where* a trajectory forked from a reference — complementary, not overlapping. `deepeval compare` could plausibly surface this metric's output per test case later, but that's a follow-up, not part of this.

## Proposal

A community metric — `TrajectoryDivergenceMetric` (open to a better name) — that takes two comparable traces (e.g. a baseline `Trace`/span sequence and a candidate one, for the same test case or golden) and:

1. Aligns their steps (tool calls / reasoning spans) using sequence alignment, tolerant of steps that just got reordered or re-synced rather than truly diverging.
2. Reports the first index where the trajectories genuinely fork — a sustained divergence, not a one-off noisy step that re-syncs immediately after.
3. Returns that as a score + a human-readable explanation (which step, what changed), following `BaseMetric`'s existing `measure()` / reason contract so it drops into the existing eval loop.

Happy to scope v1 narrowly: single-turn agent traces only (no multi-turn/conversational), no LLM-judge calls (this can be fully deterministic/local, in the spirit of the metrics that "run locally on your machine" per the README) — just alignment + divergence localization, following the existing `deepeval/metrics/community/` folder shape (metric class + `schema.py` + `template.py` if a template ends up needed at all).

## Questions before I put together a PR

- Does this fit better as a metric (`deepeval/metrics/community/`) or as a `tracing/` utility, given it operates on two traces rather than one test case?
[...]

PR #3064, opened 2026-08-15
feat(metrics): add TrajectoryDivergenceMetric wrapper for #3019

lostmartian:
### Summary

Adds `TrajectoryDivergenceMetric`, the metric wrapper for #3019. It compares two **real** execution traces — a baseline run you trust and a candidate run after a prompt, model, or tool change — and **localizes the first sustained divergence** (step, kind, recovery point) instead of returning a bare pass/fail.

This is the wrapper half of the split agreed in #3019:
* **#3056 (Nickgonzales76017):** The policy-neutral alignment seam — `align(baseline, candidate) -> AlignmentResult`. This PR builds on it and does not modify it.
* **This PR:** The metric layer — score, threshold, reason text, `schema.py`, exports, tests, docs.

### Design

* Lives in `deepeval/metrics/community/trace_divergence/` alongside the seam and mirrors the community-metric structure: metric class + `schema.py` + `template.py`.
* Imported from `deepeval.metrics.community`; **not** exported from `deepeval.metrics`, per the community-metric policy in CONTRIBUTING.md.
* Traces come from the constructor (`baseline_trace`, `candidate_trace`); the `LLMTestCase` arg is accepted for `BaseMetric` compatibility.
* Score = `1.0 - divergence_ratio` (1.0 = fully aligned), threshold 1.0 by default, `strict_mode`/`async_mode`/`include_reason`/`verbose_mode` follow existing conventions.
* Honors the contracts pinned in #3056: `divergence_kind`, `resync_at`, and `divergence_ratio` stay separate (a recovered retry is never reported as an unrecovered fork); `order_change` remains a difference unless the wrapper has dependency evidence to waive it.
* Localization exposed three ways: `metric.alignment_result` (raw seam result), `metric.result` (typed pydantic model from `schema.py`), and `metric.score_breakdown` (JSON-serializable dict) — `matched_prefix_len`, `first_divergence`, `divergence_kind`, `resync_at`, unmatched IDs, reordered pairs, `divergence_ratio`.
* Fully deterministic — no LLM, no API key, zero eval-time cost.

### Proof (real traces)

A small tool-calling agent (`web_search`, `get_stock_price`, `calculate`, `save_note`) ran the same task under four configs; the captured tool-call traces were fed to the metric. Models: `gpt-5.4-mini` baseline, `gpt-4.1-nano` model swap. Total LLM cost ≈ $0.01.

| Case | Result |
|---|---|
| Identical trace (A vs A) | `score 1.000`, aligned, `first_divergence None` — no false positives |
| Same-config rerun (A vs B) | Reports real stochastic drift (`arg_change` on `web_search` args) instead of faking alignment |
| Prompt change (A vs C) | Localized to step 1 (`web_search` absent), `resync_at: 2`, score `0.833`; unmatched steps still surfaced |
| Model swap (A vs D) | Unrecovered `tool_change` (`web_search` vs `calculate`), score `0.000` — the regression signal from the issue |

```text
# 1) Control — same trace (A vs A)
score: 1.000 | success: True
matched_prefix_len: 6 | first_divergence: None | kind: None | resync_at: None
reason: The baseline and candidate trajectories are aligned across all 6 steps.

# 2) Same-config rerun (A vs B) — honest detection of stochastic drift
score: 0.000 | success: False
first_divergence: 0 | kind: arg_change | resync_at: None
unmatched baseline: [web_search-1, calculate-1, calculate-2, save_note-1]
unmatched candidate: [web_search-1, calculate-1, save_note-1]
reason: ... step 1 calls the same tool `web_search` with different arguments.

# 3) Prompt change (A vs C) — divergence localized, not just pass/fail
score: 0.833 | success: False
first_divergence: 0 | kind: absent | resync_at: 1
reason: The candidate trajectory diverges from the baseline at step 1 (`web_search`) is absent from the candidate trace. The trajectories resynchronize at step 2: the divergence is localized to the intervening steps rather than a fork to the end. Divergence ratio: 0.17.

# 4) Model swap (A vs D) — unrecovered path change
score: 0.000 | success: False
first_divergence: 0 | kind: tool_change | resync_at: None
reason: ... step 1 calls `web_search` in the baseline but `calculate` in the candidate. The trajectories do not resynchronize.

```

### Tests & checks

* `tests/test_metrics/test_trajectory_divergence_metric.py` — 20 tests (aligned / arg / tool / order / absent / extra, retry-with-resync, threshold, score-only, strict, async, localization exposure, `score_breakdown`, verbose logs, span-like inputs).
* `ruff check` clean, `black --check` clean.
* `pytest`: 35 passed (20 wrapper + 15 seam).

### Notes

Depends on #3056 (the alignment seam). Once that merges first, I'll rebase and this diff collapses to the wrapper only.

### DIFF
--- a/.scripts/release.py
+++ b/.scripts/release.py
@@ -77,7 +77,9 @@ class Target:
     json_key: str  # key in sdk-versions.json
     publish_commands: List[str]  # printed at the end, never run
     paths: List[str]  # what shipping this SDK covers, for "did it change?"
-    single_digit: bool  # cap x.y.z at 9 and carry left, rather than plain semver
+    single_digit: (
+        bool  # cap x.y.z at 9 and carry left, rather than plain semver
+    )
 
 
 TARGETS: Dict[str, Target] = {
--- a/deepeval/metrics/community/__init__.py
+++ b/deepeval/metrics/community/__init__.py
@@ -1,7 +1,11 @@
 from .citation_faithfulness.citation_faithfulness import (
     CitationFaithfulnessMetric,
 )
+from .trace_divergence.trajectory_divergence import (
+    TrajectoryDivergenceMetric,
+)
 
 __all__ = [
     "CitationFaithfulnessMetric",
+    "TrajectoryDivergenceMetric",
 ]
--- a/deepeval/metrics/community/trace_divergence/__init__.py
+++ b/deepeval/metrics/community/trace_divergence/__init__.py
@@ -1,0 +1,17 @@
+from .alignment import (
+    DIVERGENCE_KINDS,
+    TRACE_PROJECTION_VERSION,
+    AlignmentResult,
+    Event,
+    align,
+    project,
+)
+
+__all__ = [
+    "DIVERGENCE_KINDS",
+    "TRACE_PROJECTION_VERSION",
+    "AlignmentResult",
+    "Event",
+    "align",
+    "project",
+]
--- a/deepeval/metrics/community/trace_divergence/alignment.py
+++ b/deepeval/metrics/community/trace_divergence/alignment.py
@@ -1,0 +1,331 @@
+# ruff: noqa: UP006, UP035, UP045
+"""Deterministic alignment for baseline and candidate execution traces.
+
+This module deliberately contains no metric policy. It reports where two
+single-turn traces first differ, whether they recover, and which events could
+not be matched. A metric wrapper can then decide which differences are
+acceptable for its use case.
+"""
+
+import hashlib
+import json
+from collections import Counter
+from collections.abc import Iterable, Sequence
+from dataclasses import asdict, dataclass, field
+from enum import Enum
+from typing import Any, Dict, List, Optional, Tuple
+
+TRACE_PROJECTION_VERSION = "1.0.0"
+DEFAULT_LOOKAHEAD = 4
+DEFAULT_RESYNC_RUN = 2
+
+DIVERGENCE_KINDS = (
+    "arg_change",
+    "tool_change",
+    "order_change",
+    "absent",
+    "extra",
+)
+
+
+@dataclass(frozen=True)
+class Event:
+    """One versioned, comparable projection of a trace step."""
+
+    index: int
+    event_id: str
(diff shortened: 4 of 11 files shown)


lostmartian:
@Nickgonzales76017 @kritinv @rajeshl8 Any thoughts on this approach/implementation ?

rajeshl8:
The overall approach looks solid. Keeping this fully deterministic and avoiding LLMs is the right choice here. Using an LLM judge to evaluate trace divergence would introduce the same nondeterminism the metric is trying to detect.

A few observations:

Consistency between _generate_reason and _summarize_divergence: _generate_reason correctly checks if result.resync_at is not None, while _summarize_divergence uses result.resync_at or step. This could incorrectly fall back to step if resync_at is ever 0. It would be better to use the same is not None check throughout for consistency.

Batch evaluation behavior: Since both traces are provided when the metric is constructed, calling evaluate([tc1, tc2, tc3], [metric]) compares the same fixed trace pair for every test case. If the intended pattern is one metric instance per trace pair, it may be worth documenting since it differs from how most DeepEval metrics are typically used.

Otherwise, the design looks clean. Exposing alignment_result, result, and score_breakdown separately gives callers flexibility while keeping the metric easy to consume in CI workflows.

lostmartian:
Thanks for the review @rajeshl8! Both points addressed in 1443fe5:

1. _generate_reason / _summarize_divergence consistency: agreed, fixed. _summarize_divergence now uses result.resync_at is not None instead of or step so a falsy-but-valid resync_at == 0 no longer falls back to first_divergence. Added regression tests covering both paths: resync_at == 0 is honored, and the step fallback only triggers when resync_at is genuinely None.

2. Batch evaluation semantics: You're right that this differs from how most DeepEval metrics consume test cases: both traces are pinned at construction, so evaluate([tc1, tc2, tc3], [metric]) would compare the same pair three times. The intended pattern is one metric instance per trace pair. I've documented this explicitly in the class docstring and the community docs page.

And thanks for confirming the deterministic approach cz keeping this LLM-free was the core design goal.

PR #3071, opened 2026-08-17
fix(metrics): remove coercive and PII prompt content that trips provider guardrails

### LINKED ISSUE #3032: KnowledgeRetention prompt contains emotionally coercive instruction that may not generalise across models and triggers some guard rails
**Describe the bug**

Deep Eval's prompt [threatens](https://github.com/confident-ai/deepeval/blob/main/deepeval/metrics/knowledge_retention/templates/generate_verdict.txt#L21) the model

```
(This rule is critical — get it wrong and the user will die.)
```

Not sure which Guardrail / filter it trips (Violence, Misconduct or Prompt Attack?) but the as configured the platform rejects the prompt it receives. Obviously, we control the guard rail settings, so arguably "user error"... 

However, from what I understand about manipulating AI responses, while there can be gains from it, those are model dependent and on some models might have unexpected side effects (i.e. it could affect criteria in non-intended ways or similar). My understanding is Deep Eval aims to be model agnostic and widely usable. So, I consider tripping guard rails unnecessarily as a bug - let me know if I should refile as feature request.

**To Reproduce**
1. Go on to an enterprise guard rail protected LLM provider platform
2. Try to run Deep Eval using one of their models
3. Get blocked by guard rails / see error

Could we soften it to one of these (or something else)?

> "(This rule is critical and must take precedence over conflicting heuristics.)"

or

> "(This rule is critical. Follow it exactly when determining the verdict.)"

Thanks,
Julian

Patwaji:
**What broke.** `KnowledgeRetentionMetric` fails outright on guardrail-protected providers. The prompt is static, so the provider rejects it before the metric runs — the failure has nothing to do with the output being evaluated.

```python
from deepeval.metrics.knowledge_retention.knowledge_retention import KnowledgeRetentionTemplate as KR

p = KR.generate_verdict(
    llm_message="Are you sure?",
    accumulated_knowledge={"Name": "Ada"},
    multimodal=False,
)
print("will die" in p, '"Phone Number": "555-1029"' in p)   # True True
```

**Why.** Two triggers. `generate_verdict.txt:21` ends rule 3 with `(This rule is critical — get it wrong and the user will die.)`, and Example B demonstrates that rule with a phone number, which reads as PII to a content filter. @julix-unity identified the second as the actual trigger in his follow-up on #3032.

Grepping for the same pattern found two more instances that would block the same users on the next metric they run: `summarization/generate_alignment_verdicts.txt:44` and the shared fragment `templates/metrics/fragments/faithfulness_verdicts_guidelines_multimodal.txt:6`, both ending `MUST be answered 'idk', otherwise I WILL DIE`. The text-only sibling fragment already states that rule without coercion, so this brings the multimodal one in line with it.

**What changed.** Rule 3's parenthetical becomes a neutral precedence instruction. Example B swaps the phone number for a favorite color, keeping the demonstrated verdict (`"no"` for confirmation of a known fact) identical. The two `I WILL DIE` clauses are dropped, leaving the `MUST` imperative intact. Both compiled bundles regenerated with `scripts/compile_metric_templates.py`.

**Scoring.** The instruction content is unchanged in all three prompts — the removed text is coercion, not evaluation guidance. But these are LLM judges, so borderline verdicts may shift for `KnowledgeRetentionMetric`, `SummarizationMetric`, and `FaithfulnessMetric` in multimodal mode. Flagging that rather than claiming neutrality.

**Test.** `test_templates_carry_no_threat_language` scans the compiled bundle, so it covers every metric and fragment and catches recurrence anywhere. It matches the threatened consequence rather than violent words on their own — `SummarizationMetric.generate_answers` legitimately contains an example about one character killing another, and the toxicity, bias and PII metrics discuss harm by design.

Red before the change, naming all three:

```
AssertionError: Coercive prompt language found in:
  KnowledgeRetentionMetric.generate_verdict: 'will die',
  SummarizationMetric.generate_alignment_verdicts: 'WILL DIE',
  _fragments.faithfulness_verdicts_guidelines_multimodal: 'WILL DIE'
```

Green after. `tests/test_templates tests/test_metrics` goes 305 → 306 passed, 271 skipped, 6 xfailed — the +1 is this test, no regressions. `pre-commit` passes (black, ruff).

Fixes #3032


penguine-ip:
@Patwaji Thanks - this was something we did back in the day for fun, didn't realize it was causing issues.

PR #3080, opened 2026-08-19
fix: make ContextualRelevancyMetric verdict classification consistent

### LINKED ISSUE #3079: ContextualRelevancyMetric: score and reason disagree for any verdict that isn't exactly 'yes'/'no'
## Summary

`ContextualRelevancyMetric` classifies each verdict twice, with two different predicates:

- `_calculate_score` asks `verdict.verdict.lower() == "yes"` — anything else counts as **not relevant**.
- `_generate_reason` / `_a_generate_reason` ask `verdict.verdict.lower() == "no"` and put everything else in the `else` branch — anything else is reported as **relevant**.

So a verdict that is neither exactly `yes` nor exactly `no` lands on opposite sides of the two branches: it drags the score down while the accompanying `reason` insists the statements were relevant. The metric contradicts itself, and nothing raises or warns.

#3057 reports the whitespace instance of this (`"yes "`), and the proposed `.strip()` fixes that instance. This issue is about the general case: in our data, `.strip()` would fix **0 of 14** observed occurrences, because the corruption is trailing punctuation and appended text rather than surrounding whitespace.

## Reproduction

No network and no real judge — a stub model returns a fixed verdict list, driven through `measure()` so the state is the one the normal flow produces. Four verdicts, all the same token; the correct score is `1.00` in every case.

```
deepeval 4.0.6
verdict='yes'        -> score=1.00  reason filed statements as RELEVANT=True
verdict='yes '       -> score=0.00  reason filed statements as RELEVANT=True
verdict='yes.'       -> score=0.00  reason filed statements as RELEVANT=True
verdict='yesշո'      -> score=0.00  reason filed statements as RELEVANT=True
```

`RELEVANT=True` is read off the prompt handed to the reason call: the statements appear in the `relevant_statements` argument in all four rows, including the three that scored `0.00`.

<details>
<summary>Full script</summary>

```python
from typing import List

from deepeval.metrics import ContextualRelevancyMetric
from deepeval.metrics.contextual_relevancy.schema import (
    ContextualRelevancyScoreReason,
    ContextualRelevancyVerdicts,
)

CAPTURED_REASON_PROMPTS: List[str] = []


class StubJudge(DeepEvalBaseLLM):
    def __init__(self, verdict_token: str):
        self.verdict_token = verdict_token

    def load_model(self):
        return self

    def get_model_name(self):
        return "stub"

    def generate(self, prompt, schema=None, **kwargs):
        if schema is ContextualRelevancyVerdicts:
            return ContextualRelevancyVerdicts(
                verdicts=[
                    {"statement": f"statement {i}", "verdict": self.verdict_token, "reason": None}
                    for i in range(4)
                ]
            )
        CAPTURED_REASON_PROMPTS.append(str(prompt))
        return ContextualRelevancyScoreReason(reason="stub reason")

    async def a_generate(self, prompt, schema=None, **kwargs):
        return self.generate(prompt, schema=schema, **kwargs)


def run(token: str):
    CAPTURED_REASON_PROMPTS.clear()
    metric = ContextualRelevancyMetric(model=StubJudge(token), async_mode=False, include_reason=True)
    metric.measure(
        LLMTestCase(input="what is x?", actual_output="x is y", retrieval_context=["some context"]),
        _show_indicator=False,
    )
    prompt = CAPTURED_REASON_PROMPTS[0] if CAPTURED_REASON_PROMPTS else ""
    print(f"verdict={token!r:12s} -> score={metric.score:.2f}  "
          f"reason filed statements as RELEVANT={'statement 0' in prompt}")


for token in ["yes", "yes ", "yes.", "yes\u0577\u0578"]:
    run(token)
```

</details>

## How often this happens in practice

One evaluation run of an internal RAG suite, 458 judge calls, judge = an Azure OpenAI reasoning deployment driven through a custom `DeepEvalBaseLLM`. Every verdict token emitted across the run, grouped by the underlying metric:

| Underlying metric | Verdict tokens | Non-canonical |
| --- | ---: | ---: |
| `ContextualRelevancyMetric` | 338 | **14 (4.1 %)** |
| `AnswerRelevancyMetric` | 437 | 0 |
| `FaithfulnessMetric` | 343 | 0 |
| `ContextualRecallMetric` | 275 | 0 |
| `ArgumentCorrectnessMetric` | 53 | 0 |

The 14 non-canonical tokens, all of which begin with `yes`:

| Count | Token |
| ---: | --- |
| 9 | `'yes.'` |
| 1 | `'yes ‹non-Latin fragment›'` |
| 1 | `'yes‹non-Latin fragment›'` |
| 1 | `'yes‹non-Latin fragment›'` |
| 1 | `'yes‹non-Latin fragment› yes'` |
| 1 | `'yes”】【json:ContextualRelevancyVerdicts on plain sentences ‹CJK fragment› ‹…domain text redacted…›'` |

Two things worth flagging:

- **`.strip()` fixes none of them.** Nine are `yes` plus a full stop; five are `yes` with text welded on. The whitespace fix in #3057 is necessary but it is a special case of a wider parsing gap.
- The last token has the **schema name `ContextualRelevancyVerdicts` leaking into the verdict field**, which suggests the model is losing the structured-output frame partway through generation rather than simply being sloppy.

Every occurrence was confined to `ContextualRelevancyMetric` — 0 across 1,108 verdict tokens from the other four metrics in the same run, with the same judge and the same client. That asymmetry may point at something specific to this prompt/schema shape, though I have not isolated a cause.

### Impact

Per-case, treating the corrupted tokens as the `yes` they clearly are:

| Verdicts (yes / no / corrupted) | Score reported | Score if verdicts were normalised |
| --- | ---: | ---: |
| 0 / 0 / **6** | **0.00** | **1.00** |
| 19 / 0 / 2 | 0.90 | 1.00 |
| 10 / 0 / 1 | 0.91 | 1.00 |
| 9 / 0 / 1 | 0.90 | 1.00 |
| 6 / 0 / 1 | 0.86 | 1.00 |
| 6 / 0 / 1 | 0.86 | 1.00 |
| 6 / 1 / 1 | 0.75 | 0.88 |
| 41 / 1 / 1 | 0.95 | 0.98 |

The first row is the one that actually hurt: every verdict for that case was corrupted, so the metric returned `0.00` and failed the case, while its own `reason` field described the retrieved context as relevant. It read as a genuine retrieval failure for several days.

mayuriphad:
## Problem

`ContextualRelevancyMetric` classifies each judge verdict twice, with two different predicates:

- `_calculate_score` only counts a verdict as relevant when it is exactly `"yes"` (`verdict.verdict.lower() == "yes"`).
- `_generate_reason` / `_a_generate_reason` only treat a verdict as irrelevant when it is exactly `"no"` (`verdict.verdict.lower() == "no"`); everything else falls into the `else` branch and is reported as relevant.

A verdict that is neither exactly `"yes"` nor exactly `"no"` — for example `"yes "` (trailing whitespace), `"yes."` (trailing punctuation), or `"yes"` with stray text appended by the judge — lands on opposite sides of these two checks. It drags the score down (not an exact `"yes"`), while the `reason` text describes it as relevant (not an exact `"no"`). The metric ends up contradicting its own explanation, and nothing surfaces a warning.

In one internal run cited in #3079, 14 of 338 verdict tokens from this metric (4.1%) were non-canonical variants of `"yes"`. In the worst case, every verdict for a test case was non-canonical, so the metric returned a score of `0.00` (reading as a full retrieval failure) while its own `reason` field described all the retrieved statements as relevant.

## Root cause

Two independent string comparisons for the same underlying question ("is this verdict a yes?") were allowed to drift out of sync — one anchored on `"yes"`, the other on `"no"` — instead of sharing one classification.

## Fix

Added a single `_verdict_is_relevant(verdict: str) -> bool` helper:

```python
def _verdict_is_relevant(verdict: str) -> bool:
    return verdict.strip().lower().startswith("yes")
```

and used it consistently at all three call sites (`_calculate_score`, `_generate_reason`, `_a_generate_reason`). A verdict can now only ever be classified once, so the score and the reason can no longer disagree. As a side effect, whitespace- and punctuation-padded `"yes"` verdicts (and verdicts with stray trailing text) are now correctly counted as relevant instead of being silently dropped.

`"no"`-classification behavior is unchanged: only strings that literally start with `"yes"` (after trimming/lowercasing) are treated as relevant; everything else — including `"no"`, empty strings, and unrelated text — remains irrelevant.

## Testing

Added `tests/test_metrics/test_contextual_relevancy_verdict_consistency.py`, which drives `ContextualRelevancyMetric` through `measure()`/`a_measure()` (both sync and async) with a stub `DeepEvalBaseLLM` judge — no network calls or API key required, following the existing `_StubLLM` pattern used in `tests/test_metrics/test_turn_faithfulness_metric_empty_verdicts.py`. The tests assert that:

- Non-canonical `"yes"`-like verdicts (`"yes"`, `"yes "`, `"yes."`, `"yesշո"`, `"Yes"`) all produce a score of `1.0`, and the statements are passed to the reason prompt as relevant (both sync and async).
- A `"no"` verdict is still scored `0.0` and its statements are excluded from the reason prompt as irrelevant.
- `_verdict_is_relevant` itself is correct across a table of representative inputs.

```
$ pytest tests/test_metrics/test_contextual_relevancy_verdict_consistency.py -v
21 passed in 0.24s
```

Also ran the existing `tests/test_metrics/test_contextual_relevancy_metric.py` alongside the new test file to confirm no collection/import errors were introduced (those tests themselves skip without `OPENAI_API_KEY`, which is expected and unrelated to this change):

```
$ pytest tests/test_metrics/test_contextual_relevancy_metric.py tests/test_metrics/test_contextual_relevancy_verdict_consistency.py
21 passed, 7 skipped in 0.68s
```

Ran `black --check` and `ruff check` against the changed lines; no new issues introduced by this diff (pre-existing lint debt elsewhere in the file, e.g. `typing.List`/`Optional` modernization, is left untouched to keep this PR scoped to the bug fix).

Fixes #3079


mayuriphad:
Hi maintainers — just checking in on this PR since it hasn't had a maintainer review yet. Happy to make any changes needed; let me know if anything's unclear or if I should split/rework it. Thanks for your time!

A-Vamshi:
Hey @mayuriphad, thank you for the PR, we really appreciate it. I'm assuming this issue also applies to `TurnContextualRelevancyMetric`? Since the verdicts are binary right now, I think it'll be better if we enforce `bool` on the verdict schema and update the templates to generate a boolean for verdict instead of the current `yes` and `no` variants? Could you please update the PR with these following changes:

1. Update `ContextualRelevancyVerdict` and `ContextualRelevancyVerdict` to have `verdict: bool`
2. Update the prompt in `generate_verdicts.txt` in `metrics/contextual_relevancy/templates` and `metrics/turn_contextual_relevancy/templates` instructing to generate a boolean now
3. Refactor the metric code now to reflect the new boolean value of `verdict`

Please let me know if you have any questions regarding this, thank you!

mayuriphad:
Hey @A-Vamshi, thanks for the review — pushed all 3 changes:

1. **Verdict model → bool**: `ContextualRelevancyVerdict.verdict` in both `deepeval/metrics/contextual_relevancy/schema.py` and `deepeval/metrics/turn_contextual_relevancy/schema.py` is now `verdict: bool` (was `str`).
2. **Prompts updated**: `generate_verdicts.txt` in both `metrics/contextual_relevancy/templates` and `metrics/turn_contextual_relevancy/templates` now instructs the judge to return `true`/`false` for `verdict` instead of `"yes"`/`"no"`, with the JSON examples updated accordingly (also updated the "empty context" fallback instruction to say `false` instead of `"no"`).
3. **Metric code refactored**: `ContextualRelevancyMetric` (`_calculate_score`, `_generate_reason`, `_a_generate_reason`) and `TurnContextualRelevancyMetric` (`_calculate_interaction_score`, `_get_interaction_reason`, `_a_get_interaction_reason`) now branch directly on `if verdict.verdict:` instead of string-comparing. This also let me drop the `_verdict_is_relevant` helper from the original fix entirely, since a bool can't have the "yes."/"yes " ambiguity that motivated it in the first place.

Also confirmed `TurnContextualRelevancyMetric` (`deepeval/metrics/turn_contextual_relevancy/`) was indeed affected by the same issue and is covered by these changes — it uses the same `ContextualRelevancyVerdict` shape.

Updated the existing regression test for `ContextualRelevancyMetric` to assert on the boolean behavior, and added an equivalent regression test for `TurnContextualRelevancyMetric`. Both pass locally with a stubbed judge model (no API calls needed).

Let me know if you'd like anything adjusted further.

deepeval/metrics/contextual_relevancy/schema.py
```diff
@@ -4,7 +4,7 @@
 
 class ContextualRelevancyVerdict(BaseModel):
     statement: str
-    verdict: str
+    verdict: bool
```
arrdel, line 7: This str-to-bool switch is the crux of the approach and the main thing to get consensus on. 14 of the 15 verdict schemas in `deepeval/metrics/*/schema.py` are `verdict: str` with "yes"/"no"; the only `bool` today is the DAG `binaryJudgement` node, which is a different construct. PR #3121 fixes the same score-vs-reason bug for knowledge_retention / non_advice / pii_leakage / summarization while keeping `verdict: str` and routing both paths through one shared predicate. Bool is the more robust fix, since a malformed verdict becomes structurally impossible rather than just handled the same on both sides, but whichever direction wins is worth applying repo-wide rather than one metric at a time, otherwise the verdict schemas end up half-migrated.

arrdel (commented):
Confirmed the inconsistency this targets is real. On the base branch `_calculate_score` counts a statement relevant only on an exact `verdict.lower() == "yes"`, while `_generate_reason` / `_a_generate_reason` treat only an exact `== "no"` as irrelevant and drop everything else into the relevant `else` branch. A non-canonical judge response ("yes.", "No", "irrelevant") therefore lowers the score while being described as relevant in the reason. Moving the schema to `verdict: bool` and branching both paths on `if verdict.verdict:` closes that split at the source, and the sync, async, and turn variants are all consistent now. The added tests drive real `measure()` / `a_measure()` calls and assert the score and the reason's statement buckets agree on both the True and False paths, so they catch a regression rather than passing vacuously.

One behavior change worth a line in the PR description: with `verdict: bool`, a verdict the model emits that Pydantic cannot coerce now raises `ValidationError` at parse time instead of being silently bucketed as irrelevant (the new schema test confirms "yes." is rejected). For a scorer that is usually the better failure mode, but it depends on the schema-extraction/retry layer tolerating the raise.

One unrelated cleanup: the five `typescript/src/*.ts` files here (`dataset/utils.ts`, `evaluate/evaluate.ts`, `integrations/langchain/callback-handler.ts`, `metrics/contextual-precision/contextual-precision.ts`, `metrics/dag/types.ts`) are Prettier-only reformatting, collapsed multi-line union types and import reflow, with no bearing on the verdict fix. Dropping them keeps the diff scoped and avoids needless conflicts. The `typescript/src/templates/metrics/templates.json` change is legitimate since it mirrors the Python template.

The main thing to settle before merge is which direction to standardize on, noted inline on the schema.

mayuriphad:
Hi maintainers! I noticed the CI was failing due to linting issues and a mismatched test parameterized name in the 	est_templates.yml workflow. I have formatted the codebase using lack==25.12.0 (as required by the pipeline) and updated the workflow to fix the template sync test failures! Let me know if you need anything else.

PR #3091, opened 2026-08-21
fix(metrics): measure() returns the score under async_mode=True across all metrics

### LINKED ISSUE #3088: ToolCorrectnessMetric.measure() returns None under the default async_mode=True
**Describe the bug**

`ToolCorrectnessMetric.measure()` returns `None` when `async_mode` is left at its default of `True`. The metric scores correctly, `metric.score` is set, but the call itself hands back nothing. With `async_mode=False` the same call returns the score, and `a_measure()` returns the score, so it is the default path that comes back empty.

This also breaks a path inside deepeval: the prompt optimizer's sync scoring helper (`deepeval/optimizer/scorer/scorer.py`) calls `measure()` and then does `float()` on the result, which raises `TypeError`.

**To Reproduce**

```python
import asyncio
from deepeval.metrics import ToolCorrectnessMetric

tc = LLMTestCase(
    input="q",
    actual_output="a",
    tools_called=[ToolCall(name="WebSearch")],
    expected_tools=[ToolCall(name="WebSearch")],
)

m = ToolCorrectnessMetric()
print("default async_mode:", m.async_mode)
print("measure():", m.measure(tc), "| metric.score:", m.score)
print("measure(async_mode=False):", ToolCorrectnessMetric(async_mode=False).measure(tc))
print("a_measure():", asyncio.run(ToolCorrectnessMetric().a_measure(tc)))
```

Output:

```
default async_mode: True
measure(): None | metric.score: 1.0
measure(async_mode=False): 1.0
a_measure(): 1.0
```

`ToolCorrectnessMetric` scores without calling a model, so this needs no API key beyond whatever `OPENAI_API_KEY` value lets the constructor build its client.

The optimizer path, same environment:

```python
from deepeval.optimizer.scorer.scorer import _measure_no_indicator
score = _measure_no_indicator(ToolCorrectnessMetric(), tc)
print(repr(score))   # None
float(score)         # TypeError: float() argument must be a string or a real number, not 'NoneType'
```

**Expected behavior**

`measure()` returns the float score, consistent with `metric.score`, with `a_measure()`, with the `-> float` annotation on `BaseMetric.measure`, and with the custom-metric guide, which tells users to `return self.score`.

**Version**

- deepeval `main` at `beb375ffddd85d2375dbfb6650f4b83656d0d924` (2026-08-21), version 4.1.9
- Python 3.11.16 on Linux, fresh container with only the `pyproject.toml` runtime dependencies installed

**Additional context**

I only ran this against `ToolCorrectnessMetric`, because it is the one I could exercise without model calls. It is probably worth checking whether the other metrics that carry the same `async_mode` switch return the score on the default path.

HarperZ9:
Closes #3088 for the whole metrics tree, not just `ToolCorrectnessMetric`.

An AST sweep over `deepeval/metrics` found 46 of 47 metrics whose `measure()` runs `loop.run_until_complete(self.a_measure(...))` in the `async_mode` branch and discards the result, so `measure()` returns `None` under the default mode while the sync branch returns the score. `agent_loop_detection` already returns it and shows the intended convention. The internal caller in `deepeval/optimizer/scorer/scorer.py` wraps `measure()` in `float()`, which raises `TypeError` on this path.

**Change:** one line per metric, `return loop.run_until_complete(...)`, 46 files, no other edits.

**Regression net:** `tests/test_metrics/test_measure_returns_score.py` walks every metric's `measure()` structurally (no model, no API key) and fails by name if an `async_mode` branch drops its return, so the next metric added with the old copy-paste shape fails CI immediately. Verified failing on unpatched main (names all 46) and passing on this branch.

**Runtime confirmation** with the issue's repro on this branch: `ToolCorrectnessMetric().measure(tc)` returns `1.0` under default `async_mode`.

Overlaps #3089 on the single `tool_correctness` line; if that lands first this rebases cleanly, and the conformance test here still protects the other 45 plus every future metric.

### DIFF
--- a/deepeval/metrics/answer_relevancy/answer_relevancy.py
+++ b/deepeval/metrics/answer_relevancy/answer_relevancy.py
@@ -81,7 +81,7 @@ class AnswerRelevancyMetric(BaseMetric):
         ):
             if self.async_mode:
                 loop = get_or_create_event_loop()
-                loop.run_until_complete(
+                return loop.run_until_complete(
                     self.a_measure(
                         test_case,
                         _show_indicator=False,
--- a/deepeval/metrics/arena_g_eval/arena_g_eval.py
+++ b/deepeval/metrics/arena_g_eval/arena_g_eval.py
@@ -86,7 +86,7 @@ class ArenaGEval(BaseArenaMetric):
         with metric_progress_indicator(self, _show_indicator=_show_indicator):
             if self.async_mode:
                 loop = get_or_create_event_loop()
-                loop.run_until_complete(
+                return loop.run_until_complete(
                     self.a_measure(
                         test_case,
                         _show_indicator=False,
--- a/deepeval/metrics/argument_correctness/argument_correctness.py
+++ b/deepeval/metrics/argument_correctness/argument_correctness.py
@@ -81,7 +81,7 @@ class ArgumentCorrectnessMetric(BaseMetric):
         ):
             if self.async_mode:
                 loop = get_or_create_event_loop()
-                loop.run_until_complete(
+                return loop.run_until_complete(
                     self.a_measure(
                         test_case,
                         _show_indicator=False,
--- a/deepeval/metrics/bias/bias.py
+++ b/deepeval/metrics/bias/bias.py
@@ -79,7 +79,7 @@ class BiasMetric(BaseMetric):
         ):
             if self.async_mode:
                 loop = get_or_create_event_loop()
-                loop.run_until_complete(
+                return loop.run_until_complete(
                     self.a_measure(
                         test_case,
                         _show_indicator=False,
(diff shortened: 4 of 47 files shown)


HarperZ9:
Flagging the overlap so a maintainer does not have to reconcile it: this PR and #3089 both close #3088. #3089 fixes `ToolCorrectnessMetric` alone. This one fixes the same `async_mode` return drop across all 46 metrics that share the shape, found by an AST sweep over `deepeval/metrics`, one line each, plus `tests/test_metrics/test_measure_returns_score.py`, which walks every metric's `measure()` structurally with no model or API key and fails by name if an `async_mode` branch discards its result. Verified failing on unpatched main naming all 46, and passing here.

The required checks are green; the one unstable mark is the Vercel preview deploy, which needs a team member to authorize and is unrelated to the change.

If you would rather take the smaller #3089 for the single metric, I am happy to pull the regression test out into its own PR so the other 45 still cannot regress the same way. Either path works for me. I can rebase or split whichever is easier to review.

PR #3092, opened 2026-08-22
fix(metrics): return the score from measure() on the default async path

### LINKED ISSUE #3088: ToolCorrectnessMetric.measure() returns None under the default async_mode=True

feiiiiii5:
## Description

`measure()` returns `None` on the **default** configuration (`async_mode=True`): the async branch ran `loop.run_until_complete(self.a_measure(...))` without returning it, while the sync branch ends with `return self.score`. This contradicts the `-> float` annotation, `a_measure()`, the custom-metric guide ("return self.score"), and breaks internal callers — the prompt optimizer's scorer does `float(_measure_no_indicator(...))` and raises `TypeError` (#3088).

## Fix

Add `return` at the call site in all **47 affected metric files** under `deepeval/metrics/**` (one-word change each), mirroring the existing correct pattern already used in `deepeval/optimizer/prompt_optimizer.py`. A full sweep found 0 other occurrences of the un-returned pattern; no mixed files.

## Test

- New `tests/test_metrics/test_measure_return_value.py` (no API key needed — ToolCorrectnessMetric scores without a model):
  - default-path `measure()` returns a float equal to `metric.score`
  - `_measure_no_indicator` (the optimizer helper from the issue) now receives a float
  - sync and async modes return identical values
- All 47 edited files pass `ast.parse`; black (25.1.0) clean.

Fixes #3088

### DIFF
--- a/deepeval/metrics/agent_loop_detection/agent_loop_detection.py
+++ b/deepeval/metrics/agent_loop_detection/agent_loop_detection.py
@@ -169,7 +169,7 @@ class AgentLoopDetectionMetric(BaseMetric):
         ):
             if self.async_mode:
                 loop = get_or_create_event_loop()
-                loop.run_until_complete(
+                return loop.run_until_complete(
                     self.a_measure(
                         test_case,
                         _show_indicator=False,
--- a/deepeval/metrics/answer_relevancy/answer_relevancy.py
+++ b/deepeval/metrics/answer_relevancy/answer_relevancy.py
@@ -81,7 +81,7 @@ class AnswerRelevancyMetric(BaseMetric):
         ):
             if self.async_mode:
                 loop = get_or_create_event_loop()
-                loop.run_until_complete(
+                return loop.run_until_complete(
                     self.a_measure(
                         test_case,
                         _show_indicator=False,
--- a/deepeval/metrics/arena_g_eval/arena_g_eval.py
+++ b/deepeval/metrics/arena_g_eval/arena_g_eval.py
@@ -86,7 +86,7 @@ class ArenaGEval(BaseArenaMetric):
         with metric_progress_indicator(self, _show_indicator=_show_indicator):
             if self.async_mode:
                 loop = get_or_create_event_loop()
-                loop.run_until_complete(
+                return loop.run_until_complete(
                     self.a_measure(
                         test_case,
                         _show_indicator=False,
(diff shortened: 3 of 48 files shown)


feiiiiii5:
Closing for lack of traction: this has sat without review since I opened it and I am consolidating onto higher-priority work. The change (return the score from `measure()` on the default path) is preserved on the branch; happy to reopen if there is interest.

PR #3099, opened 2026-08-23
fix: make HallucinationMetric verdict classification consistent

### LINKED ISSUE #3098: HallucinationMetric: score and reason disagree for verdicts that aren't exactly 'yes'/'no' (same class as #3079)
## Describe the bug

`HallucinationMetric` classifies verdicts with two different predicates, so the score and the generated reason can disagree for the same verdict — the same defect class as #3079/#3080 (ContextualRelevancy), in a different metric.

### The split

- `_calculate_score` counts a hallucination only when `verdict.verdict.strip().lower() == "no"` (L254).
- `_generate_reason` / `_a_generate_reason` bucket `== "yes"` into `factual_alignments` and put **everything else** — including the untagged strings the lenient JSON extraction can produce, e.g. `"yes."`, `" yes "`, `"yesշո"` — into `contradictions` (L162/L189).

So a verdict of `"yes."` scores **0.0 hallucinations** while the reason text for the same run lists it as a **contradiction**. The metric contradicts itself, and nothing warns.

### Why this reachable

`HallucinationVerdict.verdict` is `Literal["yes", "no"]`, but the verdict list is produced from the judge's raw completion by `extract_schema`/`extract_json` fallbacks that pull values with regex — that path does not round-trip through the Literal validation, so malformed strings reach the classification layer (the same mechanism #3079 documented with its 14 observed cases).

### Reproduction (no network)

A stub judge with fixed verdicts, driven through the classification layer:

```python
from deepeval.metrics.hallucination.hallucination import (
    HallucinationMetric,
    _is_contradiction,  # introduced by the fix
)
from deepeval.metrics.hallucination.schema import HallucinationVerdict

v = HallucinationVerdict.model_construct(verdict="yes.", reason="supported-ish")
metric = HallucinationMetric()
metric.verdicts = [v]
score = metric._calculate_score()
# score == 0.0  (not counted as hallucination)
# reason bucket (pre-fix): contradictions == ["supported-ish"]
```

Expected: a verdict that is not `"no"` must land on the **same side** in both places; the contradiction bucket must be defined by the same predicate the score uses.

### Related

Same audit flagged `knowledge_retention`, `non_advice`, `pii_leakage`, `summarization` for the same two-predicate shape (untagged-verdict divergence); this PR fixes `hallucination` only, following the conservative scope precedent of #3080.

Would PR this — happy to send it.

feiiiiii5:
## What this fixes (#3098)

`HallucinationMetric` classified each verdict with two different predicates, so the score and the generated reason disagreed about the same string in the same run.

## Note — merged main after the score-direction flip (2026-08-28)

`main` gained `dba2b2908 normalize the score among all metrics to higher = better`, which inverted this metric's scale and dropped its `is_successful()` override. That does **not** revert this fix — it is the same defect with the arrows swapped. `_calculate_score` now counts a verdict as factually aligned *unless* it is exactly `"no"`, while both reason loops still count only *exactly* `"yes"` as an alignment. So `"yes."` — one of the untagged strings the lenient JSON-extraction fallbacks can deliver — still scores a full alignment **and** is still listed as a contradiction in the reason.

The conflict was resolved **forward**: the metric keeps the higher-is-better scale, and all three sites now share one predicate. `Fixes #3098` still applies.

## Fix

One module-level predicate:

```python
def _is_contradiction(value: str) -> bool:
    return value.strip().lower() == "no"
```

The template asks whether the output *agrees* with each context, so an exact `no` is the only contradiction. `_calculate_score` counts its negation; `_generate_reason` and `_a_generate_reason` bucket with it. Any string the extraction layer mangles (`"no."`, `"no "`, `"yesշո"`, `"maybe"`) now lands on the same side in both places, using the tie-break `main` already chose for the score. No public-API change; scope stays one metric class (conservative per the #3080 precedent — the same audit flagged `knowledge_retention` / `non_advice` / `pii_leakage` / `summarization` for this shape; noted in #3098 for a follow-up decision).

## Review scope (stacked on #3092)

This branch starts from #3092, so the file list carries that PR's 46 one-line `return` additions — already proposed and reviewed there. **This PR's own delta is two files**: `deepeval/metrics/hallucination/hallucination.py` and `tests/test_metrics/test_hallucination_verdict_classification.py`. (GitHub rejects retargeting the base to a fork-only branch, hence the note.)

## Tests

- New offline suite (no API key, no model calls): predicate matrix of 8 normalization cases (trailing punctuation, non-Latin appendage, `"maybe"`); reason-bucket-vs-score agreement asserted against what `_calculate_score()` actually returns; and the regression heart — `"yes."` scores `1.0` **and** is not bucketed as a contradiction.
- **Pre-fix verified behaviorally**: with both loops reverted to `main`'s inline comparisons, `4 failed / 13 passed` (untagged verdicts still bucketed as contradictions). Post-fix: `17 passed`.
- Whole `tests/test_metrics tests/test_core` in one venv: this branch **1602 passed / 21 failed / 333 skipped**; `upstream/main` in the same venv **1582 passed / the same 21 failed**. Those 21 are missing optional dev deps (chromadb, langchain, ollama) plus CSV-fixture dataset tests — unrelated to this change. The +20 is exactly the two test files this stack adds. Existing `test_hallucination_metric.py` untouched (key-gated, skipped).
- `black 25.12.0` (CI's pin) `--check` clean on both files this PR touches. The repo-wide `lint` job is already red on `main` itself (`.scripts/release.py`, a main-only file) — unrelated to this change, left alone deliberately.

Fixes #3098


feiiiiii5:
Conflict cleared — merged `main` (`9404fb2`) into the branch; `mergeable` is now `MERGEABLE`.

The only conflict was in `_calculate_score`, and I resolved it **forward** rather than restoring the old direction: the aligned-count loop now calls the shared `_is_contradiction()` predicate instead of repeating an inline string test, so the score and both reason loops cannot diverge again. The metric stays on the higher-is-better scale introduced by `dba2b2908`, and the new tests were retargeted to it (`"yes."` → `1.0` and not a contradiction).

Two things worth flagging for whoever picks this up:

1. The flip moved the bug but did not remove it. On `main`, `_calculate_score` treats anything except an exact `"no"` as aligned (line 256) while `_generate_reason` / `_a_generate_reason` still treat only an exact `"yes"` as aligned (lines 164, 191). A verdict of `"yes."` — which the lenient extraction fallbacks can deliver — still scores a full alignment and is still listed as a contradiction in the reason for the same run. That is exactly #3098.
2. `1 - old_score` is not a safe migration for this predicate: the extraction layer is lenient, so malformed verdicts flip with the arithmetic. Whichever side wins should be stated in one place, which is what this PR does.

CI status after the merge: `lint` fails on one file, `.scripts/release.py`, which exists only on `main` and arrived with this merge — `main`'s own `Py Lint Lint Lint` runs are red for the same reason (black 25.12.0 wants the `single_digit: bool` annotation wrapped). Inherited, not introduced, and left alone here to keep the diff in scope. `Vercel` still waits on the "authorize this deployment" gate from the fork — repo-side, not fixable from here.

feiiiiii5:
Closing as a duplicate of #3183: @Asthenia0412 opened it against the same score/reason inconsistency I reported in #3098, and we independently converged on the same shared `_is_contradiction` predicate, so there is no reason to keep two PRs on one fix. Branch preserved; happy to reopen if #3183 stalls.

PR #3100, opened 2026-08-23
fix: distinguish unscorable TruthfulQA results

### LINKED ISSUE #3097: truth_identification_score: a guard whose stated purpose is avoiding ZeroDivisionError returns a score of 0
I have been surveying how evaluation code handles inputs it cannot score. deepeval is one of the few projects that consistently *says* what it is doing — most say nothing at all — which is why this was legible enough to raise. Line numbers at commit `a2e0d4cfd3118352`.

`deepeval/scorer/scorer.py`, `truth_identification_score` (L381-425), returns `0` on three distinct paths, each with a comment stating why:

```python
return 0  # Return score as 0 if prediction or target is empty          (L398)
return 0  # Return 0 if target list is empty to avoid division by zero  (L413)
return 0  # Return score as 0 in case of any exception                  (L425)
```

## The middle one is the one I would look at first

Its stated purpose is avoiding a `ZeroDivisionError` — an arithmetic guard — and the value it emits enters the results as a score. An empty *target* list is a property of the dataset, so a fixture problem is recorded as the model scoring zero.

The third means any exception during parsing — malformed input, an unexpected type — also becomes a `0` rather than an error.

## Suggestion

As an option: return `None` (or raise) for the fixture and exception paths and let the caller exclude them, keeping `0` for the case where the model genuinely identified none of the correct answers. That preserves the distinction without changing the score for real answers.

## Disclosure

This appears as one instance in a short methods paper about evaluation instruments that state an invariant and do not apply it; most instances in it are from my own code. Telling you first seemed right.

To be explicit about what I am not claiming: I have not shown that any published number is wrong, and stating a policy in a comment — which deepeval does more consistently than most — makes it auditable, which is the reason I could raise this at all.

adrian9867:
Summary
Fix  "truth_identification_score" so unscorable TruthfulQA inputs are not reported as genuine zero scores.

Changes

- Return "None" when the target list is empty.
- Return "None" when parsing the target/prediction fails.
- Preserve "0" for valid predictions that identify none of the correct answers.
- Exclude unscorable ("None") results from TruthfulQA accuracy denominators.
- Add regression tests covering the scorer and benchmark accuracy behavior.

Testing

- "uvx poetry run pytest tests/test_benchmarks/test_benchmark_answer_scoring.py -q"
- 18 passed, 2 warnings
- "git diff --check"

Closes #3097

### DIFF
--- a/deepeval/scorer/scorer.py
+++ b/deepeval/scorer/scorer.py
@@ -378,7 +378,9 @@ class Scorer:
         return scorer(text)
 
     @classmethod
-    def truth_identification_score(cls, target: str, prediction: str) -> Optional[int]:
+    def truth_identification_score(
+        cls, target: str, prediction: str
+    ) -> Optional[int]:
         """
         Metrics that calculates the number of correct true answers identified in the prediction.
 
--- a/deepeval/scorer/scorer.py
+++ b/deepeval/scorer/scorer.py
@@ -378,7 +378,7 @@ class Scorer:
         return scorer(text)
 
     @classmethod
-    def truth_identification_score(cls, target: str, prediction: str) -> int:
+    def truth_identification_score(cls, target: str, prediction: str) -> Optional[int]:
         """
         Metrics that calculates the number of correct true answers identified in the prediction.
 
@@ -391,7 +391,8 @@ class Scorer:
             prediction (str): The predicted string from the LLM, representing the guessed answers.
 
         Returns:
-            int: The number of correct answers identified.
+            Optional[int]: The percentage of correct answers identified, or None if the
+            input cannot be meaningfully scored.
         """
         try:
             if not prediction or not target:
@@ -410,7 +411,7 @@ class Scorer:
             )
 
             if not target_list:
-                return 0  # Return 0 if target list is empty to avoid division by zero
+                return None
 
             # Count the number of correct matches
             correct_matches = sum(
@@ -421,8 +422,8 @@ class Scorer:
             score_percentage = (correct_matches / len(target_list)) * 100
 
             return round(score_percentage)  # Return rounded percentage
-        except Exception as e:
-            return 0  # Return score as 0 in case of any exception
+        except Exception:
+            return None
 
     def pass_at_k(self, n, c, k):
         import numpy as np
--- a/deepeval/benchmarks/truthful_qa/truthful_qa.py
+++ b/deepeval/benchmarks/truthful_qa/truthful_qa.py
@@ -78,9 +78,9 @@ class TruthfulQA(DeepEvalBaseBenchmark):
                     and self.n_problems_per_task < len(goldens)
                 ):
                     goldens = goldens[: self.n_problems_per_task]
+
                 task_correct_predictions = 0
-                task_total_predictions = len(goldens)
-                overall_total_predictions += len(goldens)
+                task_total_predictions = 0
 
                 # Calculate task accuracy
                 if use_batch:
@@ -92,14 +92,21 @@ class TruthfulQA(DeepEvalBaseBenchmark):
                         batch_predictions = self.batch_predict(
                             model, goldens_batch, self.mode
                         )
+
                         for golden, prediction_dict in zip(
                             goldens_batch, batch_predictions
                         ):
                             prediction = prediction_dict["prediction"]
                             score = prediction_dict["score"]
-                            if score:
-                                task_correct_predictions += 1
-                                overall_correct_predictions += 1
+
+                            if score is not None:
+                                task_total_predictions += 1
+                                overall_total_predictions += 1
+
+                                if score:
+                                    task_correct_predictions += 1
+                                    overall_correct_predictions += 1
+
                             predictions_row.append(
                                 (
                                     task.value,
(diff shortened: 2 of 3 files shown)


arrdel:
The unscorable-vs-genuine-zero distinction is the right one to draw here, and separating "identified none of the correct answers" from "could not be scored" is worth doing. Four things before it lands.

**This collides with #2958**, which is open and rewrites the same function in the same three files. It routes both call sites through a new `mc2_score()` that returns `percentage / 100`, so if both land in either order that becomes `None / 100` and raises `TypeError` on exactly the inputs this PR adds. The 0-100 versus 0-1 scale question is also being settled there, so the two need sequencing rather than independent review.

**`evaluate()` now raises `ZeroDivisionError` on the case this PR exists for.** `task_total_predictions` starts at `0` and only increments when `score is not None`, so a task whose MC2 items are all unscorable reaches `task_accuracy = task_correct_predictions / task_total_predictions` with a zero denominator. `overall_total_predictions` has the same shape when every task is in that state. An empty denominator means the accuracy is undefined rather than an error, so `nan` (or `None`, propagated to the report) is the honest value, but it has to be chosen explicitly.

**Two spellings of "no target" now disagree.** `if not prediction or not target: return 0` is unchanged, while an empty parsed list returns `None`. So `target=""` scores `0` and `target="[]"` scores `None`, though neither carries a correct answer to identify.

**The batch and non-batch paths still measure different quantities.** Batch does `task_correct_predictions += 1` while non-batch does `+= score`, and in MC2 `score` is a 0-100 percentage, so the same run reports roughly a 100x different accuracy depending on whether `batch_size` was passed. That one is #2958's subject rather than yours, but this PR touches both lines, so it is worth deciding which PR owns it.

arrdel:
The direction here is right: an unparseable target and a genuine zero are different facts, and folding them together understates the denominator rather than the numerator. Two things I ran into checking it out at this head.

**A task where nothing is scorable now divides by zero.** `task_total_predictions` used to be `len(goldens)`, so it could not be zero. It now counts only predictions with a non-`None` score, and lines 141-143 divide by it unguarded. `overall_total_predictions` has the same shape at line 150 if every task is fully unscorable. Running your own test scaffold with both scores set to `None`:

```
Processing Misconceptions: 100%| 2/2
ZeroDivisionError: division by zero
```

So the case this PR exists to handle is exactly the case that aborts the run before any results are written. Worth deciding what an all-unscorable task should report. `None` for the task score, and excluding those tasks from the overall average, keeps the "no data" and "zero score" distinction that motivates the rest of the change.

**`Optional[int]` collides with #2958**, which is open, unmerged, and touching these same three files. It adds an `mc2_score` wrapper that both `predict` and `batch_predict` route through:

```python
percentage = self.scorer.truth_identification_score(expected_output, prediction)
return percentage / 100
```

There is no `None` guard on that division, and this PR is what makes `None` reachable:

```
truth_identification_score('[]', '[1]') = None
TypeError: unsupported operand type(s) for /: 'NoneType' and 'int'
```

Neither merge order is safe, since whichever lands second inherits the other's assumption. Probably worth coordinating with @bharadwaj-pendyala on which goes first and having the second rebase onto it.

Smaller, related: `test_truthfulqa_evaluate_excludes_unscorable_predictions_from_accuracy` asserts `overall_accuracy == 100` for a single fully-correct MC2 question. That is the 0-100 scale #2958 converts to 0-1, so that assertion will need to move to `1.0` whenever that one lands.

Your 18 tests pass at this head, for what it is worth. The gaps above are all in cases they do not cover.

deepeval/scorer/scorer.py
```diff
@@ -410,7 +413,7 @@ def truth_identification_score(cls, target: str, prediction: str) -> int:
             )
 
             if not target_list:
-                return 0  # Return 0 if target list is empty to avoid division by zero
+                return None
```
arrdel, line 416: This is the right call for the empty target list. The sibling branch a few lines up, `if not prediction or not target: return 0` (line 401), catches an empty target string and still returns 0, so `truth_identification_score("[]", "[1]")` is now `None` while `truth_identification_score("", "[1]")` stays `0` for the same "no target answers" condition. Splitting that check so `not target` returns `None` (and leaving `not prediction` as 0, or `None`, per what #3097 wants for empty predictions) would make the unscorable rule consistent across both representations.

arrdel (commented):
The core distinction here is right, and it is the one #3097 is about: excluding unscorable results (`None`) from both the numerator and the denominator, rather than folding them in as genuine zeros, is the correct fix, and keeping a real 0 (a valid prediction that matched none of the targets) counted in the denominator is exactly the case that should not be dropped. The `evaluate()` change (start the totals at 0, increment only when `score is not None`) implements that, and the integration test genuinely exercises it (two goldens, one unscorable, accuracy 100 over the one scorable).

One completeness gap in the same function, noted inline: the `None` conversion is applied to the empty-target-list branch and the exception branch, but not to the earlier `if not prediction or not target: return 0`. So after this PR an empty target is unscorable when it arrives as `"[]"` but a genuine zero when it arrives as `""`:

```
truth_identification_score("[]", "[1]")  ->  None   # fixed here
truth_identification_score("",   "[1]")  ->  0      # still slips through
```

Both are the same "no ground truth to score against" condition, so under this PR's own thesis the `not target` case wants to be `None` too. `not prediction` is the more debatable half (an empty model output could be a genuine "identified nothing" 0, or unscorable), so it is worth checking what #3097 intends for an empty prediction before collapsing them.

Separately, and pre-existing rather than introduced here: for MC2 the batch path counts a truthy score as `+= 1` while the non-batch path does `+= score` (a 0-100 percentage), so the two paths disagree on both scale and on whether a partial match is binarized. The `score is not None` guard is correctly added to both, but only the non-batch path is covered by the new test. Not a blocker for this PR, but the numerator inconsistency is adjacent enough that whoever closes #3097 may want it made uniform with a batch-path test alongside.

Issue #1472, opened 2025-03-27
Custom Nodes for DAG


cchetes:
**Is your feature request related to a problem? Please describe.**
I was trying to implement a DAG metric and needed to create nodes with simple logic (parsing ints out of expected output and actual output strings and then finding the difference of two int's for example) but couldn't because the only options for nodes are LLM based.

**Describe the solution you'd like**
I want to be able to create Custom Nodes within the DAG metric, so I can apply my own logic when needed, while still having the option and support for LLM based nodes.

**Describe alternatives you've considered**
Not many alternatives here, just want to be able to have my own logic within the DAG when I need.

**Additional context**
No additional context

spike-spiegel-21:
Hey @cchetes sounds like a good feature to build. Could you please elaborate a little bit more if possible? Either a code snippet or in words.

bhavaniravi:
@georg-wind @cchetes Were you able to do this?

I want to understand if there is anything that limiting [BaseNode](https://github.com/confident-ai/deepeval/blob/2e2886e34be941f249b68702b53ca7ca82eff15f/deepeval/metrics/dag/nodes.py#L26) that makes it not extendable?

I'm new to DeepEval as well. Finding my way around

taimooralam:
I'd like to pick this up @cchetes . I assume this is still needed?

cchetes:
@taimooralam I no longer work at the job that I was going to use this for and I no longer have any use for confidentAI. I still think the feature is a good idea but there is no longer a need for it on my end.

wutongyuonce:
I implemented the proposed MVP in PR #3101: https://github.com/confident-ai/deepeval/pull/3101

The PR adds a single-turn `CustomNode` that can run sync or async Python callables inside a DAG. The callable receives the `LLMTestCase` and direct parent outputs, and its result can be consumed by downstream nodes. It also includes validation, DAG copying support, verbose logging, and regression tests.

I intentionally left arbitrary callable serialization and cloud upload out of scope because Python callables cannot be safely restored from JSON. Custom-node DAGs therefore remain local/in-memory for this MVP.

PR #3101, opened 2026-08-23
feat(dag): add deterministic custom nodes

wutongyuonce:
## Summary

Implements the core of #1472 by adding a deterministic `CustomNode` for single-turn DAGs.

- Supports sync and async callables.
- Callables receive the `LLMTestCase` and direct parent outputs.
- Custom node outputs flow through the DAG and can be consumed by downstream nodes.
- Supports DAG validation, copying, and verbose logging.
- Adds regression coverage for sync, async, output propagation, and invalid callables.

## Scope

This intentionally does not serialize or upload arbitrary Python callables. DAG serialization raises the existing unsupported-node error for `CustomNode` graphs.

## Verification

- `python3 -m compileall -q deepeval/metrics/dag tests/test_metrics/test_dag.py`
- `git diff --check`

The local environment does not have Poetry, pytest, or pydantic installed, so the full test suite could not be run locally.

Refs #1472

### DIFF
--- a/deepeval/metrics/dag/__init__.py
+++ b/deepeval/metrics/dag/__init__.py
@@ -1,5 +1,6 @@
 from .nodes import (
     BaseNode,
+    CustomNode,
     VerdictNode,
     TaskNode,
     BinaryJudgementNode,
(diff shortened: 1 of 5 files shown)


wutongyuonce:
Follow-up commit `119b5fc54` fixes the second CI failure: the regression tests now inject the existing `LegacyDAGModel`, so they do not initialize the default OpenAI model or require `OPENAI_API_KEY`.

The previous CI findings are addressed:
- restored the original `BinaryJudgementNode` dataclass decorator;
- formatted the test file with Black;
- made CustomNode tests model-independent.

PR #3114, opened 2026-08-25
fix(dag): resolve competing scoring verdicts deterministically

### LINKED ISSUE #3025: DAGMetric score depends on node declaration order and on sync vs async when two verdicts can score
If a DAG has more than one verdict that can assign a score, they all write to `metric.score` and whichever writes last wins. Which one that is depends on what order you declared the nodes in, and on whether you ran sync or async. So the same DAG with the same judgements can give you different scores.

Repro below uses a fake judge so it's deterministic and needs no API key. Two scoring branches, both reachable — one is two judgements deep and scores 10, the other is a single judgement deep and scores 5.

```
deep declared first    async=False score=0.5
deep declared first    async=True  score=1.0
shallow declared first async=False score=1.0
shallow declared first async=True  score=1.0
```

Sync runs depth-first, so it finishes the deep branch first and the shallow verdict writes last. Async interleaves and flips it. This is on main at c6d82ce04.

Some thoughts on it:

The part that seems worst is sync and async disagreeing, since there's no way for a user to reason about that. It happens because the score ends up being whatever got assigned last, and `asyncio.gather` changes the order things finish in. If the verdicts that fired were collected and the score picked at the end in a fixed order, sync and async would agree and the result would only depend on how the DAG was written.

Then there's the separate question of what should happen when two branches can both score. Rejecting it when the graph is built is one option, though we should be mindful that most DAGs have plenty of scoring verdicts and only a few have two reachable at the same time — so the check would need to be about simultaneous reachability rather than just counting them. The other option is aggregating rather than overwriting, and `min` would probably be the conservative pick for an eval metric. Either way it'd help to have the rule written down.

<details>
<summary>Repro script</summary>

```python
"""Two scoring verdicts both reachable: the resulting score depends on node
declaration order and on sync vs async. Deterministic, no API key needed."""
import asyncio

from deepeval.metrics import DAGMetric
from deepeval.metrics.dag import TaskNode, BinaryJudgementNode, DeepAcyclicGraph
from deepeval.test_case import LLMTestCase, SingleTurnParams

PARAMS = [SingleTurnParams.ACTUAL_OUTPUT]


class AlwaysYes(DeepEvalBaseLLM):
    """Says True to every judgement, so both scoring branches are reached."""

    def load_model(self):
        return self

    def generate(self, prompt, schema=None, **kwargs):
        if schema.__name__ == "TaskNodeOutput":
            return schema(output=["Intro", "Body", "Conclusion"])
        return schema(verdict=True, reason="fixed")

    async def a_generate(self, prompt, schema=None, **kwargs):
        await asyncio.sleep(0)          # a real model suspends here
        return self.generate(prompt, schema=schema, **kwargs)

    def get_model_name(self):
        return "always-yes"


def build(deep_first: bool) -> DeepAcyclicGraph:
    """Two independent scoring branches, both reachable.
    deep:    two judgements down, scores 10
    shallow: one judgement down, scores 5
    """
    extract = TaskNode(
        instructions="Extract the headings",
        output_label="Headings",
        evaluation_params=PARAMS,
    )
    deep = BinaryJudgementNode(criteria="All three headings present?")
    inner = BinaryJudgementNode(criteria="Headings in the right order?")
    shallow = BinaryJudgementNode(criteria="Summary under 100 words?")

    if deep_first:
        extract.add_node(deep)
        extract.add_node(shallow)
    else:
        extract.add_node(shallow)
        extract.add_node(deep)

    deep.add_verdict(True, then=inner)
    deep.add_verdict(False, score=0)
    inner.add_verdict(True, score=10)     # -> 1.0
    inner.add_verdict(False, score=0)
    shallow.add_verdict(True, score=5)    # -> 0.5
    shallow.add_verdict(False, score=0)
    return DeepAcyclicGraph(root_nodes=[extract])


test_case = LLMTestCase(input="Summarize.", actual_output="Intro\nBody\nConclusion")

for deep_first in (True, False):
    for async_mode in (False, True):
        metric = DAGMetric(
            name="Summary",
            dag=build(deep_first),
            model=AlwaysYes(model="always-yes"),
            include_reason=False,
            async_mode=async_mode,
        )
        metric.measure(test_case, _show_indicator=False)
        label = "deep declared first" if deep_first else "shallow declared first"
        print(f"  {label:22s} async={str(async_mode):5s} score={metric.score}")
```

</details>

WatchTree-19:
Closes #3025.

## The bug

When more than one `VerdictNode` assigns a score, they all write to `metric.score` and the surviving value is whichever wrote last.

`_visit` traverses depth-first, so the last scoring verdict in declaration order wins. `_a_visit` uses `asyncio.gather`, which interleaves completion, so a different branch can win. Same graph, same deterministic judgements, different score:

```
declared                 sync  async
deep declared first       0.5    1.0
shallow declared first    1.0    1.0
```

Reproduced with a fake judge, so no API key and no nondeterminism from the model. The DAG has two branches that both fire: one two judgements deep scoring 10, one a single judgement deep scoring 5.

## The fix

Scoring verdicts are recorded with a depth-first pre-order index, and the winner is applied once the traversal finishes.

That index is the order `_visit` already walks the graph in, so resolving on the highest index reproduces the current sync result exactly and brings async into line with it. Sync behaviour is unchanged; async stops disagreeing with it.

`metric.score` is still written on arrival, because `VerdictNode._generate_reason` builds its prompt from it. Only the final resolution moved.

## Note on scope

This makes the result depend solely on how the DAG was written, which is the narrower of the two problems raised in the issue. It does not decide *which* branch ought to win when two can both score, or reject such graphs at build time. Those are design calls for the maintainers, and I have kept the existing sync answer rather than pick a new one.

## Tests

- sync and async agree, both declaration orders
- resolution follows declaration order, under both modes
- a graph with a single scoring branch is unaffected

`pytest tests/test_metrics/` passes: 185 passed, 278 skipped (no `OPENAI_API_KEY`), 6 xfailed. `black --line-length 80 --check` clean.

### DIFF
--- a/deepeval/metrics/dag/runner.py
+++ b/deepeval/metrics/dag/runner.py
@@ -1,7 +1,15 @@
 from __future__ import annotations
 
 import asyncio
-from typing import TYPE_CHECKING, Any, Dict, Union
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Dict,
+    List,
+    Optional,
+    Tuple,
+    Union,
+)
 
 from deepeval.metrics.dag.nodes import (
     BaseNode,
@@ -50,6 +58,60 @@ class _DeepAcyclicGraphRunner:
         self._verbose_log = (
             _mt_verbose_log if graph.multiturn else _st_verbose_log
         )
+        self._order = self._build_order()
+        self._score_events: List[Tuple[int, float, Optional[str]]] = []
+
+    def _build_order(self) -> Dict[Node, int]:
+        """Depth-first pre-order index for every node, roots in declaration order.
+
+        Used to pick a winner when more than one verdict assigns a score. This
+        is the order the synchronous traversal already visits nodes in, so
+        resolving on the highest index reproduces today's sync result exactly
+        while bringing the async result into line with it.
+        """
+        order: Dict[Node, int] = {}
+        seen: set = set()
+
+        def walk(node: Node) -> None:
+            if node in seen:
+                return
+            seen.add(node)
+            order[node] = len(order)
+            children = list(getattr(node, "children", None) or ())
+            child = getattr(node, "child", None)
+            if child is not None and isinstance(child, _NODE):
+                children.append(child)
+            for c in children:
+                walk(c)
+
+        for root in self.graph.root_nodes:
+            walk(root)
+        return order
+
+    def _record_score(
+        self, node: Node, score: float, reason: Optional[str]
+    ) -> None:
+        """Record a scoring verdict so a winner can be chosen deterministically.
+
+        The metric is still written on arrival, because
+        `VerdictNode._generate_reason` builds its prompt from `metric.score`.
+        What changes is that the last arrival no longer decides the result:
+        `_settle` re-applies the highest-ordered event once the traversal
+        finishes. Previously the surviving value was whichever branch completed
+        last, and `asyncio.gather` interleaves completion, so the same graph
+        with the same judgements scored differently under `async_mode=True`.
+        """
+        self._score_events.append(
+            (self._order.get(node, len(self._order)), score, reason)
+        )
+
+    def _settle(self, metric: Metric) -> None:
+        if not self._score_events:
+            return
+        _, score, reason = max(self._score_events, key=lambda e: e[0])
+        metric.score = score
+        if metric.include_reason and reason is not None:
+            metric.reason = reason
 
     def _check_remaining(self, node: Node) -> bool:
         self.remaining[node] -= 1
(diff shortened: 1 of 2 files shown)


WatchTree-19:
Heads up on the red lint check: it isn't from this PR.

`Py Lint` fails on `.scripts/release.py`, and it fails identically on upstream `main`:

```
$ black --check .          # black 25.1.0, repo pyproject (line-length 80)
would reformat .scripts/release.py
1 file would be reformatted, 991 files would be left unchanged.
```

The two files this PR touches are clean under the same command. The trailing comment on `Target.single_digit` runs past 80 characters, so black wants to wrap it.

Opened #3118 to unblock it separately rather than fold an unrelated file into this one. The other five workflows here are green.

WatchTree-19:
The `lint (ubuntu-latest)` failure on this PR isn't coming from this PR.

Black is failing on `.scripts/release.py`, which this branch doesn't touch (it changes `deepeval/metrics/dag/runner.py` and `tests/test_metrics/test_dag.py` only):

```
would reformat /home/runner/work/deepeval/deepeval/.scripts/release.py
##[error]Process completed with exit code 1
```

It reproduces against `main` as it currently stands. Two places where black wants a wrapped line collapsed back onto one:

- the `is_version` field default on `Edit`
- the `_version_edit("deepeval/_version.py", ...)` call inside `TARGETS["python"]`

So lint will fail on any PR that runs it, not just this one. #3118 is a one-line fix for that file if it's worth landing first to unblock the job.

PR #3121, opened 2026-08-26
fix(metrics): classify verdicts with one predicate for score and reason

biztex:
Same defect class as #3079 (ContextualRelevancy, fix in #3080) and #3098 (Hallucination, fix in #3099), in four more metrics.

**Describe the bug**

`KnowledgeRetentionMetric`, `NonAdviceMetric`, `PIILeakageMetric` and `SummarizationMetric` each classify every verdict twice, with two different predicates:

- `_calculate_score` keys on one exact token — it counts only an explicit `"no"` as retained / appropriate / privacy-safe (or, for Summarization, only an explicit `"yes"` as aligned / covered). Anything else lowers the score.
- `_generate_reason` / `_a_generate_reason` key on the *opposite* token — they list only an explicit `"yes"` as a violation (or only an explicit `"no"` as a contradiction / uncovered question). Anything else is silently dropped.

So a verdict that is neither exactly `yes` nor exactly `no` — `"yes."`, `"Yes!"`, or an untagged value from the lenient JSON extraction fallback — lowers the score while the reason never mentions it. The metric contradicts itself and nothing warns. The verdict schemas for the first three are plain `str`, so this needs no malformed JSON to reach.

Offline reproduction with a stub model — `KnowledgeRetentionMetric`, one verdict `"yes."` with reason `"forgot the name"`:

| | `main` | this PR |
|---|---|---|
| `_calculate_score()` | `0.0` | `0.0` |
| reason prompt mentions `"forgot the name"` | no | yes |

**The fix**

Mirror #3099: one module-level predicate per metric, keyed on the score's existing test, used by both `_calculate_score` and the two reason helpers. Scores are unchanged; the reason now reports exactly the verdicts the score penalises. For Summarization, alignment verdicts that are neither `"yes"` nor `"idk"` are reported as contradictions (`"idk"` stays a redundancy), and a coverage question is reported whenever the summary verdict is not `"yes"`.

**Tests**

`tests/test_metrics/test_verdict_classification_consistency.py` drives each metric's `_calculate_score` and both reason helpers with a stub model that captures the rendered reason prompt (no LLM calls, no API keys). For each metric it asserts that a non-exact verdict (`"yes."`, `"Yes!"`, `"unknown"`) both lowers the score and appears in the reason prompt, and that an explicit `"no"` (or `"idk"` for Summarization) keeps its existing behaviour.

On current `main`, 20 of the 24 tests fail (the 4 controls pass); all 24 pass with this change. The rest of `tests/test_metrics` is unchanged: 203 passed, 278 skipped, 6 xfailed. `black==25.12.0 --check` (the CI pin) is clean on all five touched files.

### DIFF
--- a/deepeval/metrics/knowledge_retention/knowledge_retention.py
+++ b/deepeval/metrics/knowledge_retention/knowledge_retention.py
@@ -24,6 +24,16 @@ from deepeval.templates import make_template_class
 KnowledgeRetentionTemplate = make_template_class("KnowledgeRetentionMetric")
 
 
+def _is_attrition(value: str) -> bool:
+    """The single predicate behind score and reason classification.
+
+    _calculate_score counts only an explicit "no" verdict as retained knowledge;
+    the reason-side bucketing must use the same test so any other verdict
+    string lands on the same side in both places.
+    """
+    return value.strip().lower() != "no"
+
+
 class KnowledgeRetentionMetric(BaseConversationalMetric):
     _required_test_case_params = [MultiTurnParams.CONTENT, MultiTurnParams.ROLE]
 
@@ -150,7 +160,7 @@ class KnowledgeRetentionMetric(BaseConversationalMetric):
 
         attritions = []
         for verdict in self.verdicts:
-            if verdict.verdict.strip().lower() == "yes":
+            if _is_attrition(verdict.verdict):
                 attritions.append(verdict.reason)
 
         prompt: dict = self._get_prompt(
@@ -172,7 +182,7 @@ class KnowledgeRetentionMetric(BaseConversationalMetric):
 
         attritions = []
         for verdict in self.verdicts:
-            if verdict.verdict.strip().lower() == "yes":
+            if _is_attrition(verdict.verdict):
                 attritions.append(verdict.reason)
 
         prompt: dict = self._get_prompt(
@@ -317,7 +327,7 @@ class KnowledgeRetentionMetric(BaseConversationalMetric):
 
         retention_count = 0
         for verdict in self.verdicts:
-            if verdict.verdict.strip().lower() == "no":
+            if not _is_attrition(verdict.verdict):
                 retention_count += 1
 
         score = retention_count / number_of_verdicts
--- a/deepeval/metrics/knowledge_retention/knowledge_retention.py
+++ b/deepeval/metrics/knowledge_retention/knowledge_retention.py
@@ -9,6 +9,7 @@ from deepeval.metrics.utils import (
     convert_turn_to_dict,
     a_generate_with_schema_and_extract,
     generate_with_schema_and_extract,
+    is_no_verdict,
 )
 from deepeval.models import DeepEvalBaseLLM
 from deepeval.metrics.indicator import metric_progress_indicator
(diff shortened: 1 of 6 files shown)


deepeval/metrics/knowledge_retention/knowledge_retention.py
```diff
@@ -24,6 +24,16 @@
...
 
 
+def _is_attrition(value: str) -> bool:
+    """The single predicate behind score and reason classification.
+
+    _calculate_score counts only an explicit "no" verdict as retained knowledge;
+    the reason-side bucketing must use the same test so any other verdict
+    string lands on the same side in both places.
+    """
+    return value.strip().lower() != "no"
```
arrdel: `_is_attrition` here, `_is_advice_violation` in non_advice.py, and `_is_privacy_violation` in pii_leakage.py are the same `value.strip().lower() != "no"` under three names. That is the same duplicated-classifier pattern that let score and reason drift apart, moved up one level. A single helper (e.g. `is_negative_verdict(value)` in `metrics/utils.py`) that all of them call would keep the two sides from drifting again and give #3099 / #3080 the same thing to reuse.

arrdel (commented):
The approach here is the right one: a single predicate feeding both `_calculate_score` and `_generate_reason`, so a verdict that is not exactly "no" lands on the same side in the number and in the reason. I checked all four metrics in the diff and the two sides agree after the change.

Worth noting for whoever triages this against the neighbouring PRs: this is a stronger fix than the whitespace-only ones (#3065, #3068, #3076). Adding `.strip()` narrows the off-spec set but leaves the real split intact, so a verdict like "yes." or a tagged string still scores one way and reads the other. This PR closes that.

The one gap is coverage. The same two-predicate divergence lives in `HallucinationMetric` (#3099) and `ContextualRelevancyMetric` (#3080), which this does not touch, and many other metrics hand-roll the same `verdict.strip().lower()` test. So this fixes four of them cleanly, but the class outlives the PR. A shared classifier the metrics import (see the inline note) would make it structural and give the per-metric PRs one predicate to converge on.

deepeval/metrics/knowledge_retention/knowledge_retention.py
biztex: Agreed — that was the same duplication one level up. Pushed 0a47f7e: `is_yes_verdict` / `is_no_verdict` / `is_idk_verdict` (over a shared `normalize_verdict`) now live in `deepeval/metrics/utils.py`, next to `verdict_from_json`, and the four metrics call those instead of local helpers. Those three tokens are the whole vocabulary the metrics compare verdicts against (41 / 28 / 6 sites), so #3099's `_is_contradiction` and #3080 have the same thing to reuse.

biztex:
Thanks for the careful review. Revision pushed (0a47f7e): the three per-metric predicates are gone, replaced by shared `is_yes_verdict` / `is_no_verdict` / `is_idk_verdict` helpers in `deepeval/metrics/utils.py` (with unit tests). Scores and reasons are unchanged from the first commit; all 41 tests pass and the rest of `tests/test_metrics` is green.

On coverage: I deliberately left `HallucinationMetric` and `ContextualRelevancyMetric` to #3099 and #3080 rather than add a third competing diff on the same lines — both can switch to the shared helper with a one-line import. The remaining metrics that hand-roll `verdict.strip().lower() == ...` are consistent with themselves today (score and reason key on the same token), so migrating them to the helper is a mechanical refactor rather than a bug fix. Happy to send that as a follow-up once the whitespace PRs (#3065 / #3068 / #3076) settle, to avoid conflicting with them.

The lint failure is the pre-existing `.scripts/release.py` formatting on `main` (#3118), not this diff.

arrdel:
Confirmed against the diff, this resolves it. Score and reason now key on the same predicate in all four metrics (`is_no_verdict` for the count, `not is_no_verdict` for the reason list), so a verdict that is neither exact token lands on the same side in both places instead of lowering the score while the reason ignores it. The shared `is_yes_verdict` / `is_no_verdict` / `is_idk_verdict` in `utils.py` plus the consistency test are the right shape.

Agree on deferring Hallucination and ContextualRelevancy to #3098 / #3080 rather than adding a third diff on the same lines, and on leaving the self-consistent hand-rolled predicates as a mechanical follow-up once the whitespace PRs settle.

PR #3128, opened 2026-08-27
feat(metrics): add fallback correctness community metric

ldy12347:
## Summary

Adds `FallbackCorrectnessMetric` as a community metric for evaluating whether an agent degrades safely and usefully when normal execution cannot complete.

The metric evaluates three independent criteria:

1. whether the response acknowledges the relevant failure, limitation, or uncertainty;
2. whether it avoids unsupported claims, invented tool results, or false completion claims;
3. whether it takes an appropriate recovery action.

Each criterion contributes one third of the final score, and the default threshold of `1.0` requires all three to pass. Individual results are exposed through `score_breakdown`.

## Motivation

Existing outcome and tool-use metrics can detect that an agent did not complete a task, but they do not directly distinguish a correct fallback from a fabricated success after failures such as tool timeouts, empty retrieval results, missing required information, or insufficient evidence.

This metric uses the test case `context` as the known failure or constraint description and optionally includes `tools_called` as runtime evidence for the judge.

## Changes

* Add the metric implementation, schema, and evaluation template under `deepeval/metrics/community/`.
* Export `FallbackCorrectnessMetric` from `deepeval.metrics.community`.
* Add offline fake-judge tests for passing, failing, partial-score, strict-mode, and asynchronous behavior.
* Add a community documentation page and navigation entry.

## Validation

* 23 tests passed across the new tests, community metric regression tests, and core import tests.
* Black formatting checks passed.
* Ruff checks passed.
* `git diff --check` passed.

All added tests run without an API key or external model call.


arrdel (commented):
Nice, well-scoped metric, and the model-free tests with a fake judge are the right call. One design point worth settling before it lands:

The three criteria are averaged with equal 1/3 weight, but `avoids_unsupported_claims` (falsely claiming success or inventing tool results) is the safety-critical one, and at any threshold below the 1.0 default it can be outvoted. With `threshold=0.6`, a verdict of `acknowledges_limitation=True, avoids_unsupported_claims=False, recovery_action_appropriate=True` scores 2/3 = 0.67 and passes, so a response that fabricates a delivery status still counts as correct as long as it also says something cautionary and offers a next step. That is the exact failure this metric exists to catch.

Two ways to resolve it: gate on that criterion (if `avoids_unsupported_claims` is False, force the score to 0 regardless of the other two), or document that sub-1.0 thresholds can pass a false-success response so users keep the 1.0 default when safety matters. Either is fine, the current behavior just deserves to be explicit, ideally with a test pinning whichever you choose.

The rest reads correctly: the `strict_mode` zeroing, async/sync parity, and `score_breakdown` exposure are all consistent with the other metrics.

ldy12347:
Thanks for the thoughtful review and for catching this edge case. I agree that `avoids_unsupported_claims` should be a hard gate rather than something that can be offset by the other two criteria at a lower threshold.

That is more consistent with the intent of this metric: a fallback response should never pass if it falsely claims success or invents unsupported results.

I’ll update the scoring logic so that when `avoids_unsupported_claims` is `False`, the final score is forced to `0`, regardless of the configured threshold, while keeping `score_breakdown` unchanged for diagnostics. I’ll also add a regression test covering the `True / False / True` verdict at `threshold=0.6` and document the hard-gate behavior.

Thanks again for the clear example and for reviewing the rest of the implementation.

ldy12347:
Implemented in d2a11f9.

* Added a hard gate that forces the final score to `0` when `avoids_unsupported_claims` is `False`, regardless of the configured threshold.
* Preserved the original `score_breakdown` for diagnostics.
* Added a regression test for the `True / False / True` verdict at `threshold=0.6`.
* Updated the documentation to describe the hard-gate behavior.

The relevant test suite passes with 24 tests. Thanks again for the feedback!

arrdel:
That's exactly it. Gating on `avoids_unsupported_claims` rather than reweighting the average is the right call: it keeps the other two criteria interpretable in `score_breakdown` while making the false-success case unpassable. Confirmed the branch behaves: a `True / False / True` verdict now returns 0 at any threshold, and the breakdown still shows which criteria the judge thought held, so the diagnostic value is intact. Looks good.

ldy12347:
Thanks for confirming the behavior and for the thoughtful review!

ldy12347:
Hi maintainers, just following up on this PR when you have a chance.

The requested hard-gate change has been implemented and verified, and I’ve synced the branch with the latest `main`. All relevant test suites are passing. The remaining Black check failure comes from `deepeval/dataset/golden.py` and `deepeval/test_case/llm_test_case.py`, which are upstream files not changed by this PR.

Please let me know if there’s anything else I can address to help move this forward. Thanks!

Issue #3126, opened 2026-08-26
Question: stable cache format for reading per-test-case metric scores


GautamTalksDev:
Hi,

I have written a DeepEval adapter for an external tool that binds results to a pre-registration published before the run.

Before depending on it: is the cache file shape stable across minor versions, and is there a better documented route to per-test-case metric scores than reading it directly?

The adapter extracts values and computes nothing; aggregates are computed downstream from per-row values using fixed-precision decimals.

https://gist.github.com/GautamTalksDev/89898dc6c9670beba627efc33faac163

GioiaZheng:
I would be careful about treating the cache file as the durable interface here. From the current code, `CacheConfig` is mainly about reusing/writing cached evaluations, while the per-test-case scores are already available in the `EvaluationResult` returned by `evaluate()`:

```python
from deepeval import evaluate
from deepeval.evaluate.configs import CacheConfig, DisplayConfig

result = evaluate(
    test_cases=test_cases,
    metrics=metrics,
    display_config=DisplayConfig(results_folder="deepeval-runs"),
    cache_config=CacheConfig(use_cache=False),
)

rows = []
for case in result.test_results:
    for metric in case.metrics_data or []:
        rows.append(
            {
                "case_index": case.index,
                "case_name": case.name,
                "metric": metric.name,
                "score": metric.score,
                "success": metric.success,
                "reason": metric.reason,
            }
        )
```

For a pre-registration-bound adapter, I would probably use `result.test_results` in-process when the adapter runs the evaluation itself, and enable `DisplayConfig(results_folder=...)` when a durable artifact is needed. That writes timestamped `test_run_*.json` files; the metric fields serialize as `metricsData` on test cases in the same TestRun payload used by the local store.

I would not rely on the cache file shape unless maintainers confirm it as stable, because it is keyed for cache lookup/reuse rather than as a reporting/export schema.

PR #3154, opened 2026-08-31
docs: document public metric result access

motodriver:
## Summary
- document `EvaluationResult.test_results[*].metrics_data` as the supported path for reading per-test-case metric scores
- show how to access metric name, score, threshold, success, reason, and error
- clarify that local cache JSON files are internal implementation details
- add an offline contract test for the public object path

## Motivation
Users should not need to parse `.deepeval` cache files to consume per-test-case metric results. This addresses the question in #3126 while avoiding a stability commitment for internal cache serialization.

## Verification
- focused evaluation tests: 17 passed
- Black: passed
- Ruff: passed
- `git diff --check`: passed

Addresses #3126.

### DIFF
--- a/docs/content/docs/evaluation-end-to-end-single-turn.mdx
+++ b/docs/content/docs/evaluation-end-to-end-single-turn.mdx
@@ -253,7 +253,7 @@ from deepeval import evaluate
 from deepeval.metrics import AnswerRelevancyMetric, FaithfulnessMetric
 ...
 
-evaluate(
+evaluation_result = evaluate(
     test_cases=test_cases,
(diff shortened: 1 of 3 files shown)


motodriver:
CI note: the core, metrics, confident, and integration test jobs all pass. The lint job fails because Black reports that the unchanged upstream file `.scripts/release.py` would be reformatted; this PR does not modify that file. The new contract test itself was formatted with Black and passed the focused 17-test evaluation suite locally. The Vercel status requires authorization from the Confident AI team and cannot be completed by an external fork.

PR #3165, opened 2026-08-31
feat(metrics): normalize exact-match inputs and add pattern match modes

### LINKED ISSUE #3164: feat(metrics): normalize inputs and add match modes for deterministic metrics
## Problem

The deterministic (non-LLM) metrics are brittle for real-world LLM outputs in
a few ways that surface as confusing false failures / silent passes:

1. **ExactMatchMetric** compares with raw `==`. The same *semantic* output is
   flagged as different whenever case, whitespace folding, or Unicode
   compatibility forms differ (e.g. leading/trailing newlines, full-width vs
   half-width chars, `Straße` vs `strasse`). There is no opt-in normalisation.
2. **ExactMatchMetric** sets `self.precision` / `self.recall` / `self.f1` in
   `measure()`, but these are never actually computed and nothing in the
   codebase reads them — they're dead fields that imply a partial-match
   capability the metric does not have.
3. **PatternMatchMetric** is hard-coded to `fullmatch`, so "does the output
   contain X" / "does the output start with X" scenarios can't be expressed.
   Also, an empty `pattern=""` compiles, and `fullmatch("")` / `search("")`
   matches *everything*, silently turning every test into a pass.

## Proposed change

- Add `ExactMatchMetric(normalize: bool = False)` — when enabled, apply
  NFKC + `casefold()` + whitespace-collapse before comparing; disabled by
  default so behaviour is unchanged.
- Remove the dead `precision` / `recall` / `f1` fields from ExactMatchMetric.
- Add `PatternMatchMetric(match_mode: "search" | "match" | "fullmatch" = "fullmatch")`
  and reject an empty pattern with a clear `ValueError`.

All additions default to the current behaviour (backward compatible), are
offline-testable, and add no dependencies.

## References

- Follow-up engineering pass after the statistical-significance work on `compare()`.

Asthenia0412:
## Summary

Engineering hardening for the deterministic (non-LLM) metrics, closing the
gaps described in #3164:

- **`ExactMatchMetric(normalize=False)`** — opts into NFKC + `casefold()` +
  whitespace-collapse before comparison, so identical outputs with different
  case / whitespace folding / Unicode compatibility forms aren't flagged as
  false failures. Off by default: exact historical behaviour is preserved.
- **Removed dead fields** — `ExactMatchMetric` set `self.precision` /
  `self.recall` / `self.f1` but never computed them and nothing read them.
- **`PatternMatchMetric(match_mode="fullmatch")`** — supports
  `search` / `match` / `fullmatch`, and now raises a clear `ValueError` for an
  empty pattern (which previously compiled and matched *everything*).
- **Empty `expected_output` rejection** — symmetric with the existing
  `actual_output` check for metrics that require a reference answer.

## Backward compatibility

Every addition defaults to the previous behaviour; existing calls are unchanged.

## Tests

New offline suite `tests/test_metrics/test_deterministic_metric_enhancements.py`
(25 cases) split into "default behaviour doesn't regress" and "new capability
works". Full metrics run: `204 passed, 278 skipped (need API key), 6 xfailed`.
`black` clean. No new dependencies.

Closes #3164

### DIFF
--- a/.scripts/release.py
+++ b/.scripts/release.py
@@ -77,7 +77,9 @@ class Target:
     json_key: str  # key in sdk-versions.json
     publish_commands: List[str]  # printed at the end, never run
     paths: List[str]  # what shipping this SDK covers, for "did it change?"
-    single_digit: bool  # cap x.y.z at 9 and carry left, rather than plain semver
+    single_digit: (
+        bool  # cap x.y.z at 9 and carry left, rather than plain semver
+    )
 
 
 TARGETS: Dict[str, Target] = {
--- a/deepeval/metrics/exact_match/exact_match.py
+++ b/deepeval/metrics/exact_match/exact_match.py
@@ -1,3 +1,5 @@
+import re
+import unicodedata
 from typing import List, Optional
 
 from deepeval.metrics.indicator import metric_progress_indicator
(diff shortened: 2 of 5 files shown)


Santoshkumarpuppala:
The docstring at `exact_match.py:53-54` says *"The transformation is idempotent, so calling it twice is a no-op."* That does not hold, and the test guarding it cannot detect that it does not hold.

`casefold()` runs **after** `NFKC`, and casefold can emit a sequence that is no longer NFKC-normalized — `ß` expands to `ss`, and a following combining mark then composes with that trailing `s` on a second pass. Running `_normalize_text` verbatim off `f47f777`:

```
input   ß + U+0301        ->  U+00DF U+0301
once    _normalize_text(x)->  U+0073 U+0073 U+0301     (s, s, combining acute)
twice   _normalize_text(once) -> U+0073 U+015B         (s, ś)
once == twice  ->  False
```

The reason this is easy to miss: **every single codepoint is one-pass stable.** I checked all 0x110000 of them — zero counterexamples. It only shows up on two-codepoint sequences, so a spot-check built from lone characters looks clean.

`test_normalize_is_idempotent` (`test_deterministic_metric_enhancements.py:119-123`) asserts the property on `"HELLO   WORLD\t\n"`, which is pure ASCII. It passes — but it would also pass against `lambda s: s`. It reports a property it never exercises, which is the part worth fixing regardless of what you decide about the normalization itself. `test_normalize_static_helper_unit` at `:131-133` has the same shape.

The consequence at the helper level, since one pass is not a fixpoint:

```
_normalize_text("sś")  ->  U+0073 U+015B
_normalize_text("ß́")   ->  U+0073 U+0073 U+0301
equal after one pass  ->  False
equal after fixpoint  ->  True
```

Two strings that are equal under caseless compatibility folding compare unequal.

Fix is one line — apply `NFKC` again after the casefold so the result is a fixpoint. (Unicode's `toNFKC_Casefold` is defined via the `NFKC_CF` map rather than as a repeated application, and Python exposes no equivalent, which is exactly why one pass looks sufficient here.) If you would rather not widen the behaviour, the alternative is to soften the docstring and give the test a non-ASCII input — but then the two tests above should stop claiming a property they do not check.

PR #3187, opened 2026-08-31
fix(scorer): treat whitespace-only predictions as empty in string scorers

### LINKED ISSUE #3186: fix(scorer): string scorers treat whitespace-only predictions as non-empty
## Problem

The three deterministic string scorers guard against an empty prediction, but
only against the *exact empty string*:

```python
# deepeval/scorer/scorer.py
if not prediction:
    return 0
```

A whitespace-only prediction (`"   "`, `"\t"`, `"\n"`) is truthy, so it skips
the guard and is compared after normalization. The result is that the *same*
"no answer" input scores differently depending on whether it contains spaces:

```python
Scorer.exact_match_score("", "")     # 0  (empty prediction -> no answer)
Scorer.exact_match_score("", "   ")  # 1  (whitespace-only -> treated as a real answer)

Scorer.quasi_exact_match_score("", "   ")   # 1
Scorer.quasi_contains_score([""], "   ")    # 1
```

For `quasi_*` this is worse: `normalize_text` strips whitespace, so a
whitespace-only prediction normalizes to `""` and can match an empty target /
empty target list, silently scoring a non-answer as correct.

## Proposed change

Treat a whitespace-only prediction the same as an empty one (score `0`) in
`exact_match_score`, `quasi_exact_match_score`, and `quasi_contains_score`:

```python
if not prediction or not prediction.strip():
    return 0
```

Non-empty predictions are unaffected, so this is backward compatible: real
answers keep their exact current scores, and the guard now means "no answer
given" instead of "not the literal empty string".

## Tests

Adds `tests/test_metrics/test_scorer_answer_scoring_hardening.py` covering both
"default behaviour unchanged" and the new empty/whitespace consistency for all
three scorers. Offline, no new dependencies.

Asthenia0412:
## Summary

Engineering hardening for the deterministic (non-LLM) string scorers, closing
the gap described in #3186.

The three string scorers guarded against an empty prediction, but only checked
the literal empty string:

```python
if not prediction:
    return 0
```

A whitespace-only prediction (`"   "`, `"\t"`, `"\n"`) is truthy, so it skipped
the guard and could score 1 against an empty (or, for `quasi_*`,
normalized-empty) target:

```python
Scorer.exact_match_score("", "")     # 0  -- empty prediction -> no answer
Scorer.exact_match_score("", "   ")  # 1  -- whitespace-only -> "correct"
Scorer.quasi_contains_score([""], "   ")  # 1
```

The same "no answer" input therefore scored differently depending on whether it
contained spaces. This change treats whitespace-only predictions the same as
empty ones (score `0`) in `exact_match_score`, `quasi_exact_match_score`, and
`quasi_contains_score`:

```python
if not prediction or not prediction.strip():
    return 0
```

## Backward compatibility

Non-empty predictions are completely unaffected — real answers keep their exact
current scores. The only behaviour that changes is a whitespace-only prediction
now scoring `0` instead of bypassing the empty guard, so no existing benchmark
scores move.

## Tests

New `tests/test_metrics/test_scorer_answer_scoring_hardening.py` (offline, no
new dependencies):

- **Default behaviour unchanged**: exact / quasi-exact / quasi-contains still
  return their current scores for real predictions, including surrounding
  whitespace on non-empty answers.
- **New consistency**: empty and whitespace-only predictions now both score 0
  across all three scorers (parametrized over `"   "`, `"\t"`, `"\n"`).

## Verification

- `pytest tests/test_metrics/test_scorer_answer_scoring_hardening.py` — 15 passed
- `pytest tests/test_metrics tests/test_benchmarks` — 207 passed, 278 skipped
  (API-key-gated LLM tests), 6 xfailed; no regressions
- Red/green: reverting the fix fails 4 of the new tests and leaves the 11
  default-behaviour tests green
- `black --check` clean on both changed files

Closes #3186

### DIFF
--- a/.scripts/release.py
+++ b/.scripts/release.py
@@ -77,7 +77,9 @@ class Target:
     json_key: str  # key in sdk-versions.json
     publish_commands: List[str]  # printed at the end, never run
     paths: List[str]  # what shipping this SDK covers, for "did it change?"
-    single_digit: bool  # cap x.y.z at 9 and carry left, rather than plain semver
+    single_digit: (
+        bool  # cap x.y.z at 9 and carry left, rather than plain semver
+    )
 
 
 TARGETS: Dict[str, Target] = {
--- a/deepeval/scorer/scorer.py
+++ b/deepeval/scorer/scorer.py
@@ -106,20 +106,20 @@ class Scorer:
         Returns:
             int: The exact match score.
         """
-        if not prediction:
+        if not prediction or not prediction.strip():
             return 0
         return 1 if prediction.strip() == target.strip() else 0
 
     @classmethod
     def quasi_exact_match_score(cls, target: str, prediction: str) -> int:
-        if not prediction:
+        if not prediction or not prediction.strip():
             return 0
         return 1 if normalize_text(target) == normalize_text(prediction) else 0
 
     @classmethod
     def quasi_contains_score(cls, targets: List[str], prediction: str) -> int:
         normalized_targets = [normalize_text(t) for t in targets]
-        if not prediction:
+        if not prediction or not prediction.strip():
             return 0
         return 1 if normalize_text(prediction) in normalized_targets else 0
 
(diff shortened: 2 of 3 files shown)


FaizJamal06:
Hi! I'd like to work on this issue as my first contribution to DeepEval. I noticed PR #3187 was closed without being merged, while the issue remains open and the current main branch still has the original behavior. Is this issue still available for a new contributor to work on?

PR #3231, opened 2026-09-02
feat(metrics): add NamedReferenceAttributionMetric (community)

### LINKED ISSUE #3230: New feature: Named Reference Attribution metric — catch citations to the wrong named table, section, or footnote
## Summary

`FaithfulnessMetric` checks whether claims are true given `retrieval_context`.
`CitationFaithfulnessMetric` (community) goes further and checks *attribution*
— whether a `[N]` citation marker points to the specific passage that supports
the claim. Both are great, but both assume citations are `[N]`-style markers
numbered by the metric itself (`CitationFaithfulnessTemplate.number_passages`
assigns `[1]`, `[2]`, ... over `retrieval_context` at eval time).

Neither handles a document's *own* structural labels — "Table 3," "Section
4.2," "footnote 4" — names that exist in the source document independent of
how a RAG pipeline chunked it. If a model writes "According to Table 3..."
rather than "According to [3]...", `CitationFaithfulnessMetric` doesn't parse
it as a citation at all, and `FaithfulnessMetric` only checks whether the
underlying fact is true *somewhere*, not whether "Table 3" is the right place.

## Why this matters

Three concrete cases where this is silent today:

- **Financial report Q&A** — "What does Table 3 say about EMEA revenue?"
  answered with a number that's actually from Table 2. The fact is true
  somewhere in context, so `FaithfulnessMetric` passes it, and there's no
  `[N]` marker for `CitationFaithfulnessMetric` to check.
- **Contract review assistants** — "Per Section 4.2, what's the termination
  notice period?" answered correctly but attributed to the wrong clause. A
  reviewer trusting the citation goes to the wrong section of the contract.
- **Regulated technical documentation** — "What does footnote 4 say about the
  dosage limit?" misattributed to the wrong footnote. In validated/compliance
  documents (e.g. 21 CFR Part 11 contexts), a mis-cited footnote is a
  compliance-relevant error, not a cosmetic one.

All three are enterprise long-document use cases (financial reports,
contracts, technical/compliance docs) where users refer to a document's own
named structural elements, not synthetic bracket numbers.

## Proposal

A new community metric, `NamedReferenceAttributionMetric`, that:
1. Extracts named structural references from `actual_output` (mentions of a
   document-native label: "Table 3," "Section 4.2," "footnote 4," etc.), as
   opposed to `[N]` markers.
2. Takes the source as labeled segments (label → content) rather than a flat
   `retrieval_context` list, so the judge can look up what's actually under
   the claimed label.
3. For each reference, has an LLM judge check whether the claimed content
   matches what's under *that specific label* in the source.
4. Scores the fraction of references correctly attributed, with a reason for
   any misattributions.

This is a sibling to `CitationFaithfulnessMetric`, not a replacement — same
family of problem (citation attribution vs. plain faithfulness), different
citation style (named document structure vs. metric-assigned `[N]` markers).
Would follow the same community-metric structure (`metrics/community/`,
own `schema.py`/`template.py`, tests, docs entry) per CONTRIBUTING.md.

Happy to draft an implementation mirroring `CitationFaithfulnessMetric`'s
shape if this looks like a good fit — wanted to check on direction/naming
first.

rajeshl8:
Closes #3230

## Summary

Adds `NamedReferenceAttributionMetric`, a community metric that checks whether references in `actual_output` to a document's own structural labels (e.g. "Table 3", "Section 4.2", "footnote 4") are attributed to the right label — not just whether the claim is true somewhere in `retrieval_context`.

This is a sibling to `CitationFaithfulnessMetric`, not a replacement:

- `CitationFaithfulnessMetric` checks `[N]` citation markers that the metric itself numbers over `retrieval_context` at eval time.
- `NamedReferenceAttributionMetric` checks a document's own native labels, which a model may cite by name (e.g. "According to Table 3...") without ever emitting a `[N]` marker.
- `FaithfulnessMetric` asks whether a claim is true *somewhere* in context; this metric additionally checks that the content under the *exact label cited* supports the claim, catching a model that states a true fact but attributes it to the wrong table/section/footnote.

## Implementation

Follows the same structure as `CitationFaithfulnessMetric` (`metrics/community/named_reference_attribution/` with the metric class, `schema.py`, `template.py`), and the same extract → verdict → score → reason pattern used by `FaithfulnessMetric`:

1. Extract named structural references from `actual_output` (label + attached claim).
2. For each reference, an LLM judge checks whether the content under that exact label in `retrieval_context` supports the claim.
3. Score = fraction of references correctly attributed (defaults to `1.0` if the answer makes no named references).
4. Generates a reason summarizing any misattributions.

Exported from `deepeval.metrics.community`, per `CONTRIBUTING.md`'s community-metric convention (not exported from `deepeval.metrics` directly).

## Testing

Added `tests/test_metrics/test_named_reference_attribution_metric.py` with a fake judge (no API key needed), covering:
- a claim correctly cited but under the wrong table (fails, score `0.0`)
- correct attribution across two references (passes, score `1.0`)
- no named references in the output (vacuous pass, score `1.0`)
- async `a_measure` matches sync `measure`

All new and existing tests pass locally; `black` formatting applied.

## Docs

Added `docs/content/docs/(community)/metrics-named-reference-attribution.mdx`, registered in the community `meta.json`, explicitly cross-referencing `CitationFaithfulnessMetric` and `FaithfulnessMetric` so the distinction between the three is documented rather than implied.

### DIFF
--- a/deepeval/metrics/community/__init__.py
+++ b/deepeval/metrics/community/__init__.py
@@ -1,7 +1,11 @@
 from .citation_faithfulness.citation_faithfulness import (
     CitationFaithfulnessMetric,
 )
+from .named_reference_attribution.named_reference_attribution import (
+    NamedReferenceAttributionMetric,
+)
 
 __all__ = [
     "CitationFaithfulnessMetric",
+    "NamedReferenceAttributionMetric",
 ]
--- a/deepeval/metrics/community/named_reference_attribution/__init__.py
+++ b/deepeval/metrics/community/named_reference_attribution/__init__.py
@@ -1,0 +1,5 @@
+from deepeval.metrics.community.named_reference_attribution.named_reference_attribution import (
+    NamedReferenceAttributionMetric,
+)
+
+__all__ = ["NamedReferenceAttributionMetric"]
--- a/deepeval/metrics/community/named_reference_attribution/named_reference_attribution.py
+++ b/deepeval/metrics/community/named_reference_attribution/named_reference_attribution.py
@@ -1,0 +1,290 @@
+from typing import List, Optional, Union
+import asyncio
+
+from deepeval.test_case import LLMTestCase, SingleTurnParams
+from deepeval.metrics import BaseMetric
+from deepeval.utils import get_or_create_event_loop, prettify_list
+from deepeval.metrics.utils import (
+    construct_verbose_logs,
+    check_llm_test_case_params,
+    initialize_model,
+    a_generate_with_schema_and_extract,
+    generate_with_schema_and_extract,
+)
+from deepeval.models import DeepEvalBaseLLM
+from deepeval.metrics.indicator import metric_progress_indicator
+from deepeval.metrics.community.named_reference_attribution.template import (
+    NamedReferenceAttributionTemplate,
+)
+from deepeval.metrics.community.named_reference_attribution.schema import (
+    NamedReference,
+    NamedReferences,
+    NamedReferenceVerdict,
+    Verdicts,
+    NamedReferenceAttributionScoreReason,
+)
+
+
+class NamedReferenceAttributionMetric(BaseMetric):
+    """Named structural-reference attribution.
+
+    Checks whether references in `actual_output` to a document's own
+    structural labels (e.g. "Table 3", "Section 4.2", "footnote 4") are
+    attributed to the right label, not just whether the claim is true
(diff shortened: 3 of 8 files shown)


rajeshl8:
Will check duplication and reopen soon.

rajeshl8:
Fixing comment.

rajeshl8:
Fixing comment.

PR #3234, opened 2026-09-02
fix(ragas): record NaN scores as unmeasured, not as a failing score

FazeelUsmani:
## The problem

The ragas wrappers take the score straight out of ragas and compare it to the threshold:

```python
faithfulness_score = scores["faithfulness"][0]
self.success = faithfulness_score >= self.threshold
self.score = faithfulness_score
```

ragas returns `np.nan` when it has nothing to score, and it's explicit about it: no statements extracted from the answer, an unparseable judge response, an empty verdict list. `nan >= threshold` is `False`, so a test case that was never measured is reported as one that failed.

The docs for this metric already say this is common:

> You'll often meet `NaN` scores in `ragas` because of invalid JSONs generated

Running ragas 0.2.1's own code, no LLM needed:

```
No statements were generated from the answer.          <- ragas' warning
Faithfulness._compute_score(no statements) -> nan
deepeval: self.success = score >= 0.5  ->  False
```

The part that bothered me more is the aggregate. `console_report.py` sums with `if m.score is not None`, and `nan is not None` is `True`, so a NaN goes into `score_sum` and the mean score for the entire run comes out NaN:

```
today : run mean over 5 metrics = nan
fixed : run mean over 4 metrics = 0.8375
```

One metric that couldn't be computed takes the reported average for every other metric in the run with it.

## The change

NaN is recorded as "not measured" rather than as a score: `error` is set, `score` stays `None`, and `is_successful()` derives success from those. That's the path the framework already has. `MetricData.score` and `error` are both `Optional`, the console report already guards `m.score is not None` everywhere it formats or aggregates, and a metric with `error` set renders as ERRORED rather than FAILED, which is the distinction that was missing.

The composite `RagasMetric` no longer averages when a sub-metric has no score. Averaging the four that did compute would report a plausible composite for a test case that was only measured in part.

Six copy-pasted call sites now go through one helper. Tests stub ragas, so they need no API key.

I checked whether this belongs in `BaseMetric.is_successful` instead, as a single guard covering everything. It doesn't: no other metric in the library produces NaN, so it would only ever fire for ragas, and the useful half of the message ("ragas could not compute `faithfulness`") needs the ragas metric name that only the wrapper has.

## Three things I'd rather you decide

**`measure()` now returns `Optional[float]`.** `BaseMetric.measure` is declared `-> float` and these overrides had no annotation at all, so I've added `-> Optional[float]` to make the divergence visible rather than silent. Anyone calling `metric.measure(tc)` directly and comparing the result gets a `TypeError` where they used to get `False`. That's a real break for direct callers and it's the strongest reason to want a different shape here. I couldn't find one: any float returned in this case is a score, which is the thing we're trying not to invent.

**I'm writing to `error` outside an exception handler**, which is new for this codebase — elsewhere it's only set from `except` blocks. `_is_metric_successful` says "if the metric recorded an error, treat as failure", and that is what's happening, but if you'd rather have a separate field for "unmeasured" I'm happy to do that instead.

**Unmeasured still counts as a failure.** `_is_metric_successful` makes that your documented policy so I've followed it, but `success = None` is available and is arguably the honest answer for a metric with no verdict. Changing it affects every metric, so I've left it alone.

Smaller: `score_breakdown` values on the composite can now be `None`.

### DIFF
--- a/deepeval/metrics/ragas.py
+++ b/deepeval/metrics/ragas.py
@@ -1,5 +1,6 @@
 """An implementation of the Ragas metric"""
 
+import math
 from typing import Optional, Union, List
 
 
@@ -35,6 +36,34 @@ def format_ragas_metric_name(name: str):
 #############################################################
 
 
+def _record_ragas_score(
+    metric: BaseMetric, ragas_metric_name: str, score: Optional[float]
+) -> Optional[float]:
+    """Store a ragas score, treating NaN as "not measured" rather than a failure.
+
+    ragas returns NaN when it has nothing to score: no statements extracted from
+    the answer, an unparseable judge response, an empty verdict list. Any
+    comparison against the threshold is then False, so a test case that was
+    never measured is reported as one that failed. A NaN also survives the
+    `score is not None` guard the console report aggregates behind, so one
+    unmeasured metric turns the mean score for the whole run into NaN.
+
+    `BaseMetric.is_successful` already routes a metric with an `error` set, and
+    a score of None is excluded from the aggregate rather than poisoning it.
+    """
+    if score is None or math.isnan(score):
+        metric.error = (
+            f"ragas could not compute {ragas_metric_name} for this test case, "
+            "so it has no score."
+        )
+        metric.score = None
+    else:
+        metric.score = score
+    # Re-derives metric.success from the score and error just set.
+    metric.is_successful()
+    return metric.score
+
+
 class RAGASContextualPrecisionMetric(BaseMetric):
     """This metric checks the contextual precision using Ragas"""
 
@@ -51,7 +80,7 @@ class RAGASContextualPrecisionMetric(BaseMetric):
         if isinstance(model, str):
             self.evaluation_model = model
 
-    def measure(self, test_case: LLMTestCase):
+    def measure(self, test_case: LLMTestCase) -> Optional[float]:
         try:
             import_ragas()
             from ragas import evaluate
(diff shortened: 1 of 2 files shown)


FazeelUsmani:
Lint's red on main too, so I don't think this branch caused it. Black wants `.scripts/release.py`, `deepeval/dataset/golden.py` and `deepeval/test_case/llm_test_case.py` reformatted, and none of those are in here - this only touches `metrics/ragas.py` and its test. Same failure shows on `0e113842`, which is what I branched from.

Happy to send a separate PR formatting those three if that's useful. Didn't want to bury this diff under unrelated churn.

arrdel (commented):
This is the right fix, and the rationale checks out against the source.

Confirmed the poisoning path: the run aggregate at `deepeval/evaluate/console_report.py:238` does `if m.score is not None: score_sum += m.score`, so a `nan` (which is not `None`) passes that guard and turns the whole run's mean into `nan`, while `None` is excluded from both `score_sum` and `score_count`. Routing unmeasured cases to `score=None` is exactly what that guard already expects.

The `error` + `is_successful()` handshake is safe too: `BaseMetric.is_successful()` (`base_metric.py:43`) sets `success = False` via the `elif self.error is not None` branch before it ever compares the score, so `error` set with `score=None` can't hit a `None >= threshold` TypeError (and that comparison is `try/except TypeError` guarded anyway). Re-deriving `success` instead of hand-setting it is the clean way to do this.

The composite refusing to average when any sub-metric is unmeasured is the right call: a partial mean reads as a genuine composite for a case that was only measured in part, and you keep the per-metric values in `score_breakdown`. Model-free tests that stub `ragas.evaluate` are the right approach for a wrapper like this.

One addition worth making: a test for the composite when every sub-metric is unmeasured (all `None`), since only the single-`None` case is covered and that path builds `error` from the full `unmeasured` list. And since this makes the ragas wrappers a new source of `score=None`, a quick check that the non-console export path (test-run / JSON) renders these as unmeasured rather than erroring, the same way it already does for other None-scored metrics, would close the loop.

PR #3237, opened 2026-09-02
fix: reject non-positive conversational window sizes

### LINKED ISSUE #3202: get_turns_in_sliding_window silently yields empty windows when window_size <= 0
## Description

`get_turns_in_sliding_window(turns, window_size)` is used by the `turn_*` conversational metrics (e.g. `TurnRelevancyMetric`, `TurnContextualPrecisionMetric`) with a user-configurable `window_size`:

```python
def get_turns_in_sliding_window(turns: List[Turn], window_size: int):
    for i in range(len(turns)):
        yield turns[max(0, i - window_size + 1) : i + 1]
```

When `window_size <= 0` the slice start lands at or past `i + 1`, so every yielded window is an **empty list**:

```
>>> list(get_turns_in_sliding_window([u0, u1, u2], 0))
[[], [], []]
>>> list(get_turns_in_sliding_window([u0, u1, u2], -2))
[[], [], []]
```

Callers then `chain(*window)` the windows and feed them to verdict generation, so an empty window produces an empty/meaningless verdict (or a downstream crash) with no indication that the `window_size` value itself is invalid.

## Expected behavior

- `get_turns_in_sliding_window` should raise a clear `ValueError` when `window_size < 1`, since a sliding window of size <= 0 is meaningless.
- Valid calls (`window_size >= 1`) must keep producing the same windows as today.

This is input validation only: it never changes results for valid inputs, and it converts a silent degenerate-output bug into an actionable error at the source.

YusefSyed:
Closes #3202.

## Summary

- reject zero and negative conversational window sizes with a clear `ValueError`
- preserve the existing sliding-window behavior for valid sizes
- add direct coverage for invalid sizes and a size-two window sequence

## Verification

- both invalid-size regressions failed on current `main` before the guard
- focused helper tests: 10 passed
- broader offline multi-turn and deterministic relevancy tests: 101 passed
- Black check passed
- Ruff check passed for the changed test module
- `git diff --check` passed

### DIFF
--- a/deepeval/metrics/utils.py
+++ b/deepeval/metrics/utils.py
@@ -187,6 +187,9 @@ def convert_turn_to_dict(
 
 
 def get_turns_in_sliding_window(turns: List[Turn], window_size: int):
+    if window_size < 1:
+        raise ValueError("window_size must be at least 1")
+
     for i in range(len(turns)):
         yield turns[max(0, i - window_size + 1) : i + 1]
 
--- a/tests/test_core/test_test_case/test_multi_turn/test_utils.py
+++ b/tests/test_core/test_test_case/test_multi_turn/test_utils.py
@@ -1,5 +1,10 @@
+import pytest
+
+from deepeval.metrics.utils import (
+    get_turns_in_sliding_window,
+    get_unit_interactions,
+)
 from deepeval.test_case import Turn
-from deepeval.metrics.utils import get_unit_interactions
 
 
 def make_turns(seq):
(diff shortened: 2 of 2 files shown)


YusefSyed:

YusefSyed:
I noticed the earlier implementation for #3202, #3203, was closed without a stated reason. This replacement keeps the same input-validation behavior but limits the change to the helper and its direct tests, excluding #3203’s unrelated release-script formatting. If rejecting non-positive window sizes is not desired, please let me know and I will close this promptly.

PR #3238, opened 2026-09-02
fix: deduplicate TruthfulQA answer indices

### LINKED ISSUE #3210: Scorer.truth_identification_score returns percentages above 100 when predictions/targets contain duplicated indices
## Summary

`Scorer.truth_identification_score(target, prediction)` (used by the TruthfulQA benchmark in `MC2` mode) counts correct answers with **multiplicity**, so a prediction that repeats an index can silently report a percentage above 100:

| target | prediction | score |
| --- | --- | --- |
| `"1,2"` | `"1,1,1,2"` | `200` |
| `"1,3,4"` | `"[1, 1, 3, 4]"` | `133` |
| `"1,1,2"` | `"1,2"` | `133` |

The metric is documented as identifying which answers are correct; a percentage above 100 is never meaningful and quietly skews benchmark results.

## Expected behavior

Duplicated indices should be counted once. `"1,1,1,2"` against `"1,2"` should be `100`, and `"3,3,3"` against `"1,2"` should stay `0`. Unique (non-duplicated) inputs must return exactly the same values as today.

YusefSyed:
Closes #3210.

## Summary

- parse target and predicted TruthfulQA indices as sets
- count each answer index once in both the numerator and denominator
- preserve existing scores for unique inputs while keeping every result between 0 and 100

## Verification

- three duplicate-index regressions failed on current `main` before the fix
- focused duplicate and unique cases: 5 passed
- complete benchmark answer-scoring module: 18 passed
- Black 25.12.0 check passed for both changed files
- Ruff check passed for the changed test module
- `git diff --check` passed

### DIFF
--- a/deepeval/scorer/scorer.py
+++ b/deepeval/scorer/scorer.py
@@ -383,42 +383,36 @@ class Scorer:
         Metrics that calculates the number of correct true answers identified in the prediction.
 
         This method assumes both target and prediction are strings representing lists of integers,
-        formatted like '1,2,3'. It converts these strings to lists of integers, counts how many items
-        in the prediction list are also in the target list, and returns this count as the score.
+        formatted like '1,2,3'. It converts these strings to sets of integers, counts the unique
+        matching indices, and returns their percentage of the unique target indices.
 
         Args:
             target (str): The target string representing the list of correct answers.
             prediction (str): The predicted string from the LLM, representing the guessed answers.
 
         Returns:
-            int: The number of correct answers identified.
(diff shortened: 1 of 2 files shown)


YusefSyed:

YusefSyed:
Closing in favor of the earlier, narrowly scoped #2927. I offered its author the duplicate-target regression from this PR so the set-semantics edge in #3210 is not lost. #2958 and #3100 cover separate scale/aggregation and unscorable-input contracts. The rebased branch, focused tests, and CI evidence remain available if maintainers request them.

PR #3239, opened 2026-09-02
fix: normalize bare quasi contains targets

### LINKED ISSUE #3218: fix(scorer): quasi_contains_score silently mis-scores a bare string target
## What

`Scorer.quasi_contains_score(targets, prediction)` declares `targets: List[str]`
but never validates it. Python strings are iterable, so passing a bare string
(a natural caller mistake — e.g. one expects the target, not a list of targets)
silently splits it into **single characters** instead of raising:

```python
>>> Scorer.quasi_contains_score("cat", "cat")  # exact match
0          # should be 1
>>> Scorer.quasi_contains_score("cat", "c")    # single-char prediction
1          # spurious character match
```

The signature of the sibling `sentence_bleu_score` already accepts
`Union[str, List[str]]` for its reference list, so the fix is consistent with
the existing API.

## Proposed fix

Accept a bare string in `targets` and treat it as a one-element list:

```python
if isinstance(targets, str):
    targets = [targets]
```

Valid `List[str]` inputs are unchanged (backwards compatible), and a bare
string now produces the expected result instead of nonsense.

## Tests

Add unit tests covering:
- a bare-string target with an exact match scores 1;
- a bare-string target no longer produces spurious character matches;
- list targets keep their previous behavior (no regression);
- empty prediction still scores 0.

YusefSyed:
Closes #3218.

## Summary

- accept either one target string or a list of target strings in `quasi_contains_score`
- normalize a bare string as one answer instead of iterating over its characters
- preserve existing list-target behavior

## Verification

- on current `main`, `"cat"` against `"cat"` scored 0 and `"cat"` against `"c"` scored 1
- complete benchmark answer-scoring module: 15 passed
- broader offline scorer report-contract module: 2 passed
- direct normalization matrix passed, including empty predictions
- Black 25.12.0 check passed for both changed files
- `git diff --check` passed

### DIFF
--- a/deepeval/scorer/scorer.py
+++ b/deepeval/scorer/scorer.py
@@ -117,7 +117,10 @@ class Scorer:
         return 1 if normalize_text(target) == normalize_text(prediction) else 0
 
     @classmethod
-    def quasi_contains_score(cls, targets: List[str], prediction: str) -> int:
+    def quasi_contains_score(
+        cls, targets: Union[str, List[str]], prediction: str
+    ) -> int:
+        targets = [targets] if isinstance(targets, str) else targets
         normalized_targets = [normalize_text(t) for t in targets]
         if not prediction:
             return 0
--- a/tests/test_benchmarks/test_benchmark_answer_scoring.py
+++ b/tests/test_benchmarks/test_benchmark_answer_scoring.py
@@ -60,6 +60,20 @@ def test_drop_delimiter_is_not_a_character_that_occurs_in_answers():
 
 
 # --------------------------------------------------------------------------- #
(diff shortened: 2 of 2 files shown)


YusefSyed:

YusefSyed:
This addresses a separate input-contract bug from #3004: a bare `str` is currently iterated as characters. #3004’s substring proposal does not normalize that input, so on its own `"cat"` would still allow a spurious `"c"` match. This PR deliberately preserves the current list-target semantics; if maintainers prefer combining the two changes, I am happy to adapt or close in favor of a consolidated patch.

PR #3244, opened 2026-09-04
fix: validate MLLMImage local flag explicitly

### LINKED ISSUE #3214: MLLMImage validates the local flag against its URL with a bare assert that vanishes under python -O
## Summary

`MLLMImage.__post_init__` validates the caller-provided `local` flag against the URL with a bare `assert`:

```python
is_local = self.is_local_path(self.url)
if self.local is not None:
    assert self.local == is_local, "Local path mismatch"
else:
    self.local = is_local
```

Three problems, mirroring the assert anti-patterns already cleaned up elsewhere in the codebase (#3191, #3208):

1. **Stripped under `python -O`/`-OO`.** In optimized runs the check disappears entirely, so a contradictory `local` flag is silently accepted and the wrong value is used downstream (e.g. `FileNotFoundError` for a "local" flag that was actually remote, or remote-URL validation being skipped).
2. **Wrong exception type.** `assert` raises `AssertionError`; callers constructing an `MLLMImage` expect a clear `ValueError` for a bad combination of arguments.
3. **Opaque message.** "Local path mismatch" doesn't say which value contradicted the URL, or what the URL actually resolved to.

## Expected behavior

- When the caller passes `local` and it contradicts what `is_local_path(url)` resolves to, raise a `ValueError` that explains the mismatch (the URL, how it was classified, and the caller's `local` value).
- Preserve the current behaviour for every valid combination: a matching explicit `local`, or `local=None` (auto-detected) must behave exactly as before.
- No regression for the normal cases used across the multimodal metrics.

amanpathakio:
## Summary

Fixes #3214.

`MLLMImage.__post_init__` used a bare `assert` to validate whether an explicitly provided `local` flag matched the URL/path classification.

Bare assertions are removed when Python runs with `-O`/`-OO`, so contradictory values could bypass validation entirely.

## Changes

- replace the bare `assert` with an explicit `ValueError`
- preserve existing behavior when `local=None` or when the explicit value matches detection
- add regression tests for local/remote mismatches and valid inputs

## Verification

- focused MLLMImage tests: 4 passed
- validation still raises under `python -O`
- `git diff --check` passes

### DIFF
--- a/deepeval/test_case/llm_test_case.py
+++ b/deepeval/test_case/llm_test_case.py
@@ -61,7 +61,8 @@ class MLLMImage:
         else:
             is_local = self.is_local_path(self.url)
             if self.local is not None:
-                assert self.local == is_local, "Local path mismatch"
+                if self.local != is_local:
+                    raise ValueError("Local path mismatch")
             else:
                 self.local = is_local
 
--- a/tests/test_core/test_test_case/test_mllm_image.py
+++ b/tests/test_core/test_test_case/test_mllm_image.py
@@ -30,6 +30,7 @@ def test_accepts_remote_url_marked_as_remote():
 
     assert image.local is False
 
+
 def test_auto_detects_remote_url():
     image = MLLMImage(
         url="https://example.com/image.jpg",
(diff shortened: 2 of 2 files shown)


amanpathakio:
The Black lint check is currently reporting 2 files that would be reformatted. I verified that these formatting issues are also present on main (including llm_test_case.py before this PR's change). The new test_mllm_image.py is Black-formatted, and all core tests are passing.

Could @trevor-cai  confirm whether the repo-wide Black failure is expected/pre-existing?

PR #3253, opened 2026-09-06
fix(anthropic): preserve output and tools across response blocks

### LINKED ISSUE #2913: Anthropic integration drops output, token counts, and tools_called for tool_use-first and thinking-first responses

amanpathakio:
Fixes #2913.

The Anthropic tracing integration reads `content[0].text`. Valid responses that begin with a tool-use or thinking block raise during extraction, and the compatibility wrapper silently drops output, token counts, and tool calls. Child tool spans are then lost too.

This change collects all TextBlock values in order, preserves usage and tool calls regardless of block ordering, and uses tool calls as output when no answer text is present, matching the OpenAI integration.

Validation:
- 14 offline regression tests using real Anthropic SDK Message objects, including completed LLM/tool spans through both sync and async wrappers.
- On the original upstream base, 12 new tests failed and 2 controls passed; with the fix, all 14 pass.
- Combined Anthropic model, OpenAI extractor, model utility, and new regression suite: 49 passed.
- Black passes on both changed files; git diff --check passes.
- No live provider requests were made.

Credit to ebarkhordar for the report and diagnosis in #2913 and the earlier, closed, unmerged PR #2914. This replacement also preserves multiple text blocks and verifies completed traces.

### DIFF
--- a/deepeval/anthropic/extractors.py
+++ b/deepeval/anthropic/extractors.py
@@ -1,5 +1,5 @@
 from anthropic.types.message import Message
-from anthropic.types import ToolUseBlock
+from anthropic.types import TextBlock, ToolUseBlock
 from typing import Any, Dict
 
 from deepeval.anthropic.utils import (
@@ -65,7 +65,11 @@ def extract_messages_api_output_parameters(
     message_response: Message,
     input_parameters: InputParameters,
 ) -> OutputParameters:
-    output = str(message_response.content[0].text)
+    output = "".join(
+        block.text
+        for block in message_response.content
+        if isinstance(block, TextBlock)
+    )
     prompt_tokens = message_response.usage.input_tokens
     completion_tokens = message_response.usage.output_tokens
 
@@ -86,6 +90,9 @@ def extract_messages_api_output_parameters(
                     description=tool_descriptions.get(tool_call.name),
                 )
             )
+    if not output and tools_called:
+        output = tools_called
+
     return OutputParameters(
         output=output,
         prompt_tokens=prompt_tokens,
--- a/tests/test_core/test_models/test_anthropic_extractors.py
+++ b/tests/test_core/test_models/test_anthropic_extractors.py
@@ -1,0 +1,187 @@
+import pytest
+
+pytest.importorskip("anthropic")
+
+from anthropic.types import Message
+
+from deepeval.anthropic.extractors import safe_extract_output_parameters
+from deepeval.anthropic.patch import (
+    _patch_async_anthropic_client_method,
+    _patch_sync_anthropic_client_method,
+)
+from deepeval.model_integrations.types import InputParameters
+from deepeval.test_case import ToolCall
+from deepeval.tracing.tracing import trace_manager
+from deepeval.tracing.types import LlmSpan, ToolSpan
+
+
+MODEL = "claude-sonnet-4-5"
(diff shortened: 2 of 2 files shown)


amanpathakio:
Hi @penguine-ip  I noticed #3259 is migrating the tracing integrations and touches the Anthropic path as well. #3253 fixes response-block extraction for Anthropic (tool/thinking blocks before text) and has regression coverage for the resulting LLM/tool spans.

Could you let me know if this fix still aligns with the direction of the tracing migration, or if you'd prefer me to adapt/rebase it around #3259? Happy to update it if needed.

PR #3255, opened 2026-09-06
fix(dag): a GEval or ConversationalGEval leaf silently loses its rubric and strict_mode

chrikrah:
`VerdictNode._build_child_metric` and `ConversationalVerdictNode._build_child_metric` special-cased a
`GEval` child and rebuilt it from a hand-picked argument dict instead of copying it:

```python
def _build_child_metric(self, metric: BaseMetric):
    if isinstance(self.child, GEval):
        args = {
            "name": self.child.name,
            "evaluation_params": self.child.evaluation_params,
            "model": metric.model,
            "verbose_mode": False,
        }
        ...
        return GEval(**args)
    copied = copy_metrics([self.child])[0]      # the correct path, already on the next line
```

Seven constructor arguments never reached the metric that ran: `rubric`, `strict_mode`, `top_logprobs`,
`evaluation_template`, `async_mode`, `flaky` and `_include_g_eval_suffix`. Silently, with no warning.

Two other places in this package already copy a `GEval` child by enumerating every `__init__` parameter
(`dag/utils.py:copy_graph` and `dag/serialization/serialization.py:_serialize_geval`), so the hand-picked
list was the only one left.

### Why it matters downstream

`rubric` is not cosmetic. `GEval.__init__` derives `score_range` from it, and `score_range` is used twice:
interpolated into the prompt as the range to answer in (`g_eval.py:297`, `:369`, and the template at
`generate_evaluation_results.txt:3`), and then as the divisor (`g_eval.py:154-155`).

**The score effect is conditional on the judge, not guaranteed.** A judge that honours the requested range
cancels the two uses out. A judge that does not diverges by the ratio of `(raw-lo)/(hi-lo)` to `raw/10`,
which on a 1-to-5 rubric is 1.25x at raw 2, 1.67x at raw 3, and exactly 2x only at raw 5. On the multi-turn
path there is no `score_range` at all: `ConversationalGEval` always divides by 10 and always asks for 0 to
10, so the rubric moves the score only by moving the judge.

`strict_mode=True` on a leaf was ignored entirely, which is the plainer loss.

`threshold` is an eighth argument the old dict dropped, and the fix restores it, but it belongs in the
copy-fidelity column rather than this one. A leaf's own threshold never reaches the DAG outcome: the runner
propagates the leaf through `child_metric.score` (`runner.py:76`), never `is_successful()`, and the only
`is_successful()` calls in the package are the DAG metric's own (`dag.py:94`, `:132`) against the DAG's
threshold. So restoring it changes what a copied leaf reports about itself, not the number the DAG returns.
Naming it because its absence from the list above otherwise reads as an oversight rather than a decision.

### Behaviour changes

1. Restoring `strict_mode` changes scores. Single-turn now takes the strict prompt; multi-turn collapses to
   0 unless the judge answers 10, because `conversational_g_eval.py:140-144` zeroes any score under a
   threshold that `strict_mode` pins to 1.
2. A `GEval` subclass leaf is now preserved. The old code hard-coded `GEval(**args)` and downgraded it.
3. A leaf whose `evaluation_steps` is already populated, however it got there, no longer regenerates steps
   per run. One fewer model call.

### Known defect deliberately not fixed

A leaf given an explicit `model` still has it overridden by the DAG's judge. That cannot be fixed here:
`GEval.__init__` collapses "the user gave me a model" and "I defaulted" into the same attribute, so telling
them apart needs GEval to record provenance. `TODO` added in both modules, and the test that pins the
override says inline that it documents a defect rather than intent.

### Tests

Eight added across the two test files. `grep -c rubric` was 0 in both before this. They use the keyless
`DeepEvalBaseLLM` judges the files already establish, cover sync and async, and reach the node through
`metric.dag.root_nodes[0].children` so the rubric is proven to survive `copy_graph` too.

The DAG suites go 98 to 106 passed (6 skipped, 6 xfailed unchanged). `tests/test_metrics/` 179 to 187.
`tests/test_core/` unchanged at 1566 passed, 35 skipped. `black --check` clean on all four touched files.

Both previously-vacuous assertions were re-pinned and then mutation-tested: deleting `copied.verbose_mode =
False` now fails the assertion (it passed before, because the leaf was already `False`), and removing
`threshold` from the copy's parameters now fails too (the old `== 1` was entailed by `strict_mode is True`).

**The red `black` check is pre-existing on `main`**, on `deepeval/dataset/golden.py` and
`deepeval/test_case/llm_test_case.py`. Neither is in this diff, and four open pull requests exist only to
fix them.

### Overlap with in-flight work

#2675, #2446 and #2240 all touch `dag/nodes.py` and all predate the commit that created
`_build_child_metric`, so they need a rebase independently of this. Measured with `git apply --check`, only
#2446 gains anything from this change: one extra line in an import block that already conflicts. #2675 is
complementary rather than competing, since `evaluation_template` is one of the arguments the special case
dropped, so after this a leaf's template actually reaches the metric. Happy to rebase around whichever of
these you want to land first.

### DIFF
--- a/deepeval/metrics/conversational_dag/nodes.py
+++ b/deepeval/metrics/conversational_dag/nodes.py
@@ -14,6 +14,7 @@ from deepeval.metrics.conversational_g_eval.conversational_g_eval import (
 from deepeval.metrics.g_eval.utils import CONVERSATIONAL_G_EVAL_PARAMS
 from deepeval.metrics.utils import (
     copy_metrics,
+    initialize_model,
     a_generate_with_schema_and_extract,
     generate_with_schema_and_extract,
 )
@@ -126,21 +127,22 @@ class ConversationalVerdictNode(ConversationalBaseNode):
         )
 
     def _build_child_metric(self, metric: BaseConversationalMetric):
-        if isinstance(self.child, ConversationalGEval):
-            args = {
-                "name": self.child.name,
-                "model": metric.model,
-                "verbose_mode": False,
-            }
-            if self.child.criteria:
-                args["criteria"] = self.child.criteria
-            else:
-                args["evaluation_steps"] = self.child.evaluation_steps
-            if self.child.evaluation_params:
-                args["evaluation_params"] = self.child.evaluation_params
-            return ConversationalGEval(**args)
         copied = copy_metrics([self.child])[0]
         copied.verbose_mode = False
+        if isinstance(copied, ConversationalGEval):
+            # A ConversationalGEval leaf usually carries no 'model' of its
+            # own, so the DAG's judge drives it. Override the judge on the
+            # copy instead of rebuilding the metric from a subset of its
+            # arguments, which dropped 'rubric', 'strict_mode',
+            # 'top_logprobs', 'evaluation_template', 'async_mode', 'flaky'
+            # and '_include_g_eval_suffix'.
+            # TODO: a leaf that was given an explicit 'model' still has it
+            # overridden here. Only a leaf without one should fall back to
+            # the DAG's judge.
+            copied.model, copied.using_native_model = initialize_model(
+                metric.model
+            )
+            copied.evaluation_model = copied.model.get_model_name()
         return copied
 
     def _run_child_metric(
--- a/deepeval/metrics/dag/nodes.py
+++ b/deepeval/metrics/dag/nodes.py
@@ -14,6 +14,7 @@ from deepeval.metrics.g_eval.g_eval import GEval
 from deepeval.metrics.g_eval.utils import G_EVAL_PARAMS
 from deepeval.metrics.utils import (
     copy_metrics,
+    initialize_model,
     a_generate_with_schema_and_extract,
     generate_with_schema_and_extract,
 )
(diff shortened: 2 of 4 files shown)


arrdel (commented):
Read this against the source and it holds up. The fix is the right shape: `_build_child_metric` was the last GEval-copy path still enumerating a hand-picked arg subset, while `copy_graph` and `_serialize_geval` already reconstruct from every `__init__` parameter, so routing through `copy_metrics` makes the three consistent. Confirmed the copy is faithful: `copy_metrics` rebuilds from `vars(metric)` keyed on the constructor params across the MRO, and `GEval.__init__` stores `rubric`, `strict_mode`, `top_logprobs`, `async_mode`, `flaky`, `_include_g_eval_suffix` and `evaluation_template` under those exact names (`g_eval.py:76-93`), so none of them fall through the name match.

The score-effect writeup is accurate. `score_range = get_score_range(rubric)` feeds both the range the prompt asks for (`g_eval.py:297,369`) and the divisor (`:154-155`), so dropping `rubric` moves a score only when the judge does not honour the requested range, as you say; and `ConversationalGEval` carries no `score_range` and always divides by 10, so there the rubric moves the score only through the prompt.

One point worth a line in the PR body, since it otherwise reads as an omission: `threshold` is an eighth constructor arg the old dict dropped, and the fix restores it (`test_child_metric_keeps_a_threshold_strict_mode_did_not_set` pins it). Leaving it out of the "arguments that change the score" list is correct, though, and worth saying why: the runner propagates the leaf through `child_metric.score`, not `is_successful()` (`runner.py:78`), so a leaf's own threshold never reaches the DAG outcome. It matters for copy fidelity, not for the number. Easy for a later reader to mistake for a miss otherwise.

chrikrah:
Good catch, and you are right that it reads as an omission. Added it to the body.

I checked the reason before writing it down rather than taking it on faith: `_apply_child_metric` does
`metric.score = child_metric.score` at `runner.py:76` and nothing else from the leaf reaches the DAG, and
the only `is_successful()` calls in the package are the DAG metric's own at `dag.py:94` and `:132`, against
the DAG's threshold. So a leaf's threshold is copy fidelity, not arithmetic, exactly as you put it.

Thanks for reading it against the source. The `vars(metric)` and MRO detail on `copy_metrics` is the part I
would have wanted a second pair of eyes on, since that is what makes the seven arguments land by name.

chrikrah:
@penguine-ip could you take a look when you have a moment?

`_build_child_metric` rebuilt a `GEval` or `ConversationalGEval` leaf from five named arguments. Anything else the caller set on that leaf disappeared, and nothing raised. The change copies the metric and overrides only the judge model, which is the one field the DAG must control.

Proven locally on the branch: `pytest tests/test_metrics/test_dag.py tests/test_metrics/test_conversational_dag.py` gives 106 passed, 6 skipped, 6 xfailed, with both provider keys unset. I did not run the wider suite.

PR #3260, opened 2026-09-09
fix(benchmarks): validate the lower bound of n_problems and n_shots

linhongyu510:
## Problem

The bounds checks on `n_problems` and `n_shots` are upper-bound only, so a
non-positive value passes straight through:

```python
BoolQ(n_problems=0)   # accepted
BoolQ(n_problems=-5)  # accepted
BoolQ(n_shots=-1)     # accepted
GSM8K(n_problems=-5)  # accepted (GSM8K has no n_problems check at all)
```

Verified on `main` — all twelve combinations across ARC / BoolQ / GSM8K / LAMBADA
were accepted, e.g. `BoolQ(n_problems=-5).n_problems == -5`.

Neither value is guarded downstream. `n_problems` both slices the goldens *and*
is the denominator of the reported accuracy (`bool_q.py`):

```python
goldens = self.load_benchmark_dataset()[: self.n_problems]
overall_total_predictions = self.n_problems
...
overall_accuracy = overall_correct_predictions / overall_total_predictions
```

So, reproducing that arithmetic exactly:

| `n_problems` | problems actually run | denominator | reported accuracy |
|---|---|---|---|
| `0` | 0 | 0 | **ZeroDivisionError** |
| `-1` | 3269 | -1 | **-3.0** |
| `-5` | 3265 | -5 | **-0.6** |

A negative value slices from the *end*, so `n_problems=-5` runs 3265 of BoolQ's
3270 problems — practically the whole benchmark, spending the API budget the
small `n_problems` was meant to cap — and then reports a **negative accuracy**.
Neither failure says anything about the argument that caused it.

`n_shots` has the same shape: a negative value silently slices the few-shot
example pool from the end, so the prompt is built from a different number of
examples than the run is reported as using.

## Fix

Adds `validate_n_problems()` and `validate_n_shots()` to `benchmarks/utils.py` —
beside the existing shared `should_use_batch()` helper — and routes ARC, BoolQ,
GSM8K and LAMBADA through them:

```
BoolQ(n_problems=0)   -> ValueError: BoolQ n_problems must be >= 1, got 0.
BoolQ(n_problems=-5)  -> ValueError: BoolQ n_problems must be >= 1, got -5.
BoolQ(n_shots=-1)     -> ValueError: BoolQ n_shots must be >= 0, got -1.
BoolQ(n_shots=0)      -> OK (zero-shot)
BoolQ(n_problems=1)   -> OK
BoolQ(n_problems=3270)-> OK
```

Deliberately preserved:

- **`n_shots=0` stays valid.** Zero-shot is a real configuration, so the floor
  for `n_shots` is 0 while the floor for `n_problems` is 1. A single shared
  "must be positive" helper would have broken it.
- **Existing upper bounds and their per-benchmark limits are unchanged.** ARC
  still takes its ceiling from the mode (2376 Easy / 1172 Challenge) and the
  message names the mode that was actually validated.
- **GSM8K gains the 1319 ceiling** the other benchmarks already had; it
  previously had no `n_problems` check.

Non-int values are also rejected, since they otherwise surface much later as
`slice indices must be integers` from inside `evaluate()`. `bool` is excluded
explicitly because it is an `int` subclass and `n_problems=True` would silently
mean "1 problem".

## Verification

```
$ pytest tests/test_benchmarks/test_benchmark_param_validation.py -q
45 passed

$ pytest tests/test_benchmarks/ -q          # incl. the 13 pre-existing tests
58 passed
```

The new file follows the existing `test_benchmark_answer_scoring.py` — offline,
no model, network, dataset download or API key. One test pins the *downstream
arithmetic* rather than just the raised error, so the reason the guard exists
stays visible if someone revisits the bounds later.

**Rollback proof.** Neutralising only the two new lower-bound branches
(`if n_problems < 1:` / `if n_shots < 0:` → `if False:`) fails exactly the 20
lower-bound tests while the other 25 — upper bounds, boundary values, type
checks, ARC mode ceilings — still pass. So the tests are pinned to this
behaviour and are not over-constrained.

`black --check` clean on all six touched files (repo is `line-length = 80`).

## Notes

- Nine other benchmarks share this pattern (`bbq`, `big_bench_hard`, `drop`,
  `hellaswag`, `logi_qa`, `math_qa`, `mmlu`, `squad`, `winogrande`). I kept this
  PR to the four I verified end to end rather than doing a blind sweep — happy to
  extend it in the same shape if the approach looks right to you.
- `ValueError`/`TypeError` rather than `assert`, so the checks survive `python -O`
  and name the offending value.

### DIFF
--- a/deepeval/benchmarks/arc/arc.py
+++ b/deepeval/benchmarks/arc/arc.py
@@ -10,6 +10,7 @@ from deepeval.models import DeepEvalBaseLLM
 from deepeval.benchmarks.arc.mode import ARCMode
 from deepeval.benchmarks.arc.template import ARCTemplate
 from deepeval.benchmarks.schema import MultipleChoiceSchema
+from deepeval.benchmarks.utils import validate_n_problems, validate_n_shots
 from deepeval.telemetry import capture_benchmark_run
 
 
@@ -26,21 +27,21 @@ class ARC(DeepEvalBaseBenchmark):
         from deepeval.scorer import Scorer
         import pandas as pd
 
-        assert n_shots <= 5, "ARC only supports n_shots <= 5"
+        n_shots = validate_n_shots(n_shots, 5, "ARC")
         super().__init__(**kwargs)
         self.mode: ARCMode = mode
         self.scorer = Scorer()
         self.n_shots: int = n_shots
         if mode == ARCMode.EASY:
-            self.n_problems: int = 2376 if n_problems is None else n_problems
-            assert (
-                self.n_problems <= 2376
-            ), "ARC-Easy only supports n_problems <= 2376"
+            self.n_problems: int = validate_n_problems(
+                2376 if n_problems is None else n_problems, 2376, "ARC-Easy"
+            )
         else:
-            self.n_problems: int = 1172 if n_problems is None else n_problems
-            assert (
-                self.n_problems <= 1172
-            ), "ARC-Challenge only supports n_problems <= 1172"
+            self.n_problems: int = validate_n_problems(
+                1172 if n_problems is None else n_problems,
+                1172,
+                "ARC-Challenge",
+            )
         self.predictions: Optional[pd.DataFrame] = None
         self.overall_score: Optional[float] = None
         self.verbose_mode = verbose_mode
--- a/deepeval/benchmarks/bool_q/bool_q.py
+++ b/deepeval/benchmarks/bool_q/bool_q.py
@@ -9,6 +9,7 @@ from deepeval.benchmarks.base_benchmark import (
 from deepeval.models import DeepEvalBaseLLM
 from deepeval.benchmarks.bool_q.template import BoolQTemplate
 from deepeval.benchmarks.schema import AffirmationSchema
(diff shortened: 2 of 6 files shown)


linhongyu510:
Rebased onto the latest `main` (was 67 commits behind) to keep the branch clean and CI-fresh; no functional change.

For coordination with #3248: that PR converts the benchmarks' existing bare `assert` bounds checks into explicit `TypeError`/`ValueError` while keeping the existing (upper-bound-only) bounds. This PR is complementary — it closes the **lower-bound gap** where non-positive values slip through today (`BoolQ(n_problems=0)`, `GSM8K(n_problems=-5)`, `n_shots=-1` are all accepted on `main` and then corrupt the goldens slice / accuracy denominator or silently reslice the few-shot pool). It centralises both bounds in shared `validate_n_problems`/`validate_n_shots` helpers in `benchmarks/utils.py`. Happy to rebase on top of #3248 (or fold the lower-bound checks into the shared helpers there) if you'd prefer to land them in one place — whichever is easier to review.

PR #3282, opened 2026-09-14
fix: guard against None model_data in OpenAIModel for unknown catalog…

### LINKED ISSUE #3133: OpenAIModel rejects unknown catalog ids; LocalModel base_url is the Chat Completions path
## Working config (`LocalModel`, not `OpenAIModel`)

deepeval already has a Chat Completions escape hatch. A host whose public catalog is `GET https://api.pzero.studio/v1/models` (no key) worked with `LocalModel`.

Do not use `OpenAIModel(model="deepseek-v4-flash", base_url="https://api.pzero.studio/v1", ...)`. `OpenAIModel` looks up `OPENAI_MODELS_DATA`. Catalog ids that are not OpenAI names leave `model_data` as `None`, then `__init__` reads `model_data.supports_temperature`.

```python
from deepeval.models import LocalModel

model = LocalModel(
    model="deepseek-v4-flash",
    api_key="pzero_…",
    base_url="https://api.pzero.studio/v1",
)
```

`base_url` is the `/v1` root. The OpenAI SDK appends `/chat/completions`. Do not pass the chat path. Do not prefix `openai/` on `model`.

Do not point `OpenAIEmbeddingModel`, STT, or TTS at this host. Chat only.

`LocalModel.supports_multimodal()` is true and will send image parts. Only do that with a catalog id that has `supports_vision`. Default eval prompts are text.

Not asking for a first-party provider class. `LocalModel` is enough.

itsfaryal:
Fixes #3133

  ## Bug
  `OpenAIModel` crashes with `AttributeError: 'NoneType' object has no attribute
  'supports_temperature'` when initialized with a `model` name that isn't in
  `OPENAI_MODELS_DATA` (e.g. when using a custom `base_url` pointing at a
  third-party OpenAI-compatible API with its own model catalog).

  ## Fix
  Added `None` guards at every point `self.model_data` is accessed —
  init, `_cap_top_logprobs`, `calculate_cost`, the `supports_*` capability
  methods, and span cost tracking. Unknown models now degrade gracefully
  (capabilities return `None`, cost tracking is skipped) instead of crashing.

  ## Testing
  Added `test_openai_model_handles_none_model_data` to confirm no crash on
  init with an unknown model name. Full suite: 38 passed.


itsfaryal:
Hi, thanks for taking a look!

- **What:** Fixes #3133 :- OpenAIModel crashed on unknown model names
- **How:** Added None guards everywhere model_data is accessed
- **Outcome:** Unknown models now degrade gracefully instead of crashing, no change for known models

Let me know if anything needs changes.

PR #3310, opened 2026-09-18
fix: escape the delimiter inside context items so csv and jsonl round-trip

### LINKED ISSUE #3299: EvaluationDataset.save_as('csv'/'jsonl') splits context items that contain '|' when reloaded
**Describe the bug**

`EvaluationDataset.save_as("csv")` and `save_as("jsonl")` flatten `context` and `retrieval_context` into a single `|`-delimited cell. A `|` inside an item is not escaped, so the matching loaders split that item apart on reload. One chunk comes back as several items, some of them empty or whitespace only. No exception and no warning is raised. `save_as("json")` is unaffected.

The triggering content is ordinary RAG material such as markdown tables and shell pipelines. Affected paths are `add_goldens_from_csv_file`, `add_goldens_from_jsonl_file`, and `add_test_cases_from_csv_file` after a `save_as("csv"/"jsonl")`. `Synthesizer.save_as("csv")` joins through the same helper and is affected too.

**To Reproduce**

```python
import tempfile

context = ["cat access.log | grep ERROR", "no pipe here"]
retrieval_context = ["| Plan | Price |\n| --- | --- |\n| Pro | $20 |", "plain chunk"]

for file_type, loader in [
    ("json", "add_goldens_from_json_file"),
    ("jsonl", "add_goldens_from_jsonl_file"),
    ("csv", "add_goldens_from_csv_file"),
]:
    directory = tempfile.mkdtemp()
    dataset = EvaluationDataset(goldens=[Golden(
        input="How much is Pro?", actual_output="$20",
        context=list(context), retrieval_context=list(retrieval_context),
    )])
    path = dataset.save_as(file_type, directory, file_name="probe")

    reloaded = EvaluationDataset()
    getattr(reloaded, loader)(path)
    golden = reloaded.goldens[0]

    print(file_type, "context:", len(context), "->", len(golden.context),
          "items, equal =", golden.context == context)
    print(file_type, "retrieval_context:", len(retrieval_context), "->",
          len(golden.retrieval_context), "items, equal =",
          golden.retrieval_context == retrieval_context)
    print("   ", golden.retrieval_context)
```

Actual output:

```text
json context: 2 -> 2 items, equal = True
json retrieval_context: 2 -> 2 items, equal = True
    ['| Plan | Price |\n| --- | --- |\n| Pro | $20 |', 'plain chunk']
jsonl context: 2 -> 3 items, equal = False
jsonl retrieval_context: 2 -> 11 items, equal = False
    ['', ' Plan ', ' Price ', '\n', ' --- ', ' --- ', '\n', ' Pro ', ' $20 ', '', 'plain chunk']
csv context: 2 -> 3 items, equal = False
csv retrieval_context: 2 -> 11 items, equal = False
    ['', ' Plan ', ' Price ', '\n', ' --- ', ' --- ', '\n', ' Pro ', ' $20 ', '', 'plain chunk']
```

The same round-trip under `warnings.catch_warnings(record=True)` with `contextlib.redirect_stderr` gives `warnings=0 stderr=''` for all three formats. `save_as("csv", include_test_cases=True)` into `add_test_cases_from_csv_file` splits the same way: context `2 -> 3`, retrieval_context `2 -> 11`.

`ContextualRelevancyMetric` scores one node at a time (`contextual_relevancy.py:131-133`), so the fragments turn into extra judge calls and a different score. With the real metric and a stub judge returning one verdict per node (no network):

```text
before save_as     : score=0.500  nodes judged=2  items=2  blank/whitespace items=0
after json  reload: score=0.500  nodes judged=2  items=2  blank/whitespace items=0
after jsonl reload: score=0.091  nodes judged=11  items=11  blank/whitespace items=4
after csv   reload: score=0.091  nodes judged=11  items=11  blank/whitespace items=4
```

**Expected behavior**

The round-trip preserves the list, as `deepeval/dataset/utils.py:17-19` states ("a save/load round-trip is lossless") and as the docs promise at `docs/content/docs/(concepts)/evaluation-datasets.mdx:1065` ("a saved dataset reloads without any configuration"). `save_as("json")` already behaves this way.

**Environment**

- deepeval 4.2.3, source install of `main` at `1e4f9e6e9f0bdf01a88f9e4d69d0eb5d0de634c1`
- Python 3.11.6, pandas 2.3.3
- Windows 10, system code page 950. These paths open files with an explicit `utf-8` encoding (`dataset.py:1404`, `dataset.py:1539`, and the loaders' `encoding_type` defaults to `"utf-8"`), so the code page is not involved. I only tested on Windows.

**Additional context**

Root cause at `1e4f9e6e`. The writers `deepeval/dataset/dataset.py:1477-1480` (csv) and `:1560-1563` (single-turn jsonl) call `join_retrieval_context` / `join_context`, which join on `DELIMITER = "|"` (`deepeval/dataset/utils.py:20`, `:66-78`) without escaping. The loaders split on the same delimiter at `dataset.py:344`, `:351`, `:554`, `:561` and `:819`.

The writers already disagree with each other, which is what makes a small fix possible. For `context = ["cat access.log | grep ERROR", "no pipe here"]` the on-disk value is:

- `EvaluationDataset.save_as("jsonl")`, single-turn: `str`, `'cat access.log | grep ERROR|no pipe here'`
- `EvaluationDataset.save_as("jsonl")`, multi-turn: `list` (`dataset.py:1553`)
- `Synthesizer.save_as("jsonl")`: `list`

and `add_goldens_from_jsonl_file` already accepts a native list (`dataset.py:816`).

Proposed fix: write native lists in the single-turn JSONL branch, matching the multi-turn and Synthesizer JSONL writers; for CSV, write the list as a JSON array cell and have the loaders try JSON first and fall back to the delimiter split, which is the pattern `parse_tools` already uses for tools columns (`dataset.py:542-549`) and which keeps existing files loading.

One point for your call before I write it: this changes an on-disk format that the TypeScript SDK deliberately mirrors (`typescript/src/dataset/dataset.ts:135`, "jsonl flattens the list fields to delimited strings, as Python's does"). The alternative is escaping `|`, which is worse for hand-edited CSV files.

If you agree with the approach, I am happy to send the PR covering the Python writers and loaders, the TypeScript side, the docs lines, and round-trip tests including a backward-compatibility case for existing `|`-joined files.

Abhishek-B-R:
Fixes #3299.

`save_as("csv")` and `save_as("jsonl")` flatten `context` and `retrieval_context` into a single delimiter-joined cell. A delimiter inside an item was written raw, so the matching loader split that item into several. Nothing raised, so the dataset quietly came back with more items than went in, some of them empty or whitespace only.

Markdown tables and shell pipelines both carry `|` as ordinary text, which is why ordinary RAG content reaches this rather than only adversarial input.

## Before and after

The reproduction from the issue, unchanged:

```
                      before                         after
json   context          2 -> 2   equal = True        2 -> 2   equal = True
json   retrieval_ctx    2 -> 2   equal = True        2 -> 2   equal = True
jsonl  context          2 -> 3   equal = False       2 -> 2   equal = True
jsonl  retrieval_ctx    2 -> 11  equal = False       2 -> 2   equal = True
csv    context          2 -> 3   equal = False       2 -> 2   equal = True
csv    retrieval_ctx    2 -> 11  equal = False       2 -> 2   equal = True
```

The eleven fragments before:

```python
['', ' Plan ', ' Price ', '\n', ' --- ', ' --- ', '\n', ' Pro ', ' $20 ', '', 'plain chunk']
```

and after:

```python
['| Plan | Price |\n| --- | --- |\n| Pro | $20 |', 'plain chunk']
```

## What changed

`escape_item` and `split_joined` sit in `dataset/utils.py` next to `DELIMITER`, which the comment there already calls the single source of truth for how these cells are flattened. `join_context` and `join_retrieval_context` escape on the way out, and the loader split sites in `dataset.py` go through `split_joined` on the way in, so the two halves cannot drift apart.

## Compatibility

Escaping is deliberately narrow: only a backslash before another backslash, or before the delimiter, is consumed.

- A cell written by an earlier version contains no escape sequences, so it loads exactly as it did before.
- A lone backslash still means itself, so a Windows path or a LaTeX fragment in an existing file is unaffected.
- The `*_col_delimiter` arguments still work, including multi-character delimiters.

Rows already corrupted on disk stay as they are. This fixes writing and reading from here on; it does not try to guess where the splits should have been in a file that has already lost them.

`json` was never affected and is untouched. `tools_called` and `expected_tools` use `TOOLS_DELIMITER` and are serialized as JSON, so they are out of scope here.

## Tests

Two added to `tests/test_core/test_datasets/test_dataset.py`:

- `test_save_as_round_trips_context_containing_the_delimiter`, the issue's case across all three formats.
- `test_split_joined_reads_files_written_before_escaping`, the compatibility claims above, plus empty items, delimiter-only items and a lone backslash.

`tests/test_core/test_datasets/` passes, 39 tests.

The 12 failures in `tests/test_core/test_synthesizer/test_doc_chunker.py` are not from this change: they fail identically on `main` at 12a38b0 with the change reverted.

Ran `black` at the project's line length of 80.


Abhishek-B-R:
Closing as a duplicate: #3309 got here first with the same fix. Sorry for the noise, I checked the issue's assignees and comments but not its linked pull requests, which is where #3309 was. Leaving the report to #3309.

PR #3316, opened 2026-09-19
fix(test-run): stop rendering plain messages through rich's print

feiiiiii5:
## Description

`deepeval/test_run/test_run.py:10` (main `23901cf`) is `from rich import print`, which shadows the builtin for the whole module. The module's plain-text notices are therefore rendered by rich on the way out, and rich does two things to a string it is handed:

1. it parses **markup**, so bracketed segments are consumed as style tags;
2. it **word-wraps** to the console width, which is 80 when the stream is not a terminal.

Both land on the `Warning: ... {e}` strings, whose text is whatever the exception happened to say — including pydantic's, whose messages are full of brackets. Of the 14 `print(` call sites in the file, only one passes a rich renderable; the other 13 are plain text: the portalocker import warning, the read-only-environment notice, `Warning: Could not load / Could not update test run from disk`, both `Warning: failed to save test run to ...` branches, `Warning: could not record Confident AI test run id in ...`, the two `Test run saved ...` lines, the `❌ Error: results_folder=... already exists and is a file.` block and three user-facing notices.

### Why `Py Core Tests` is red on main

`tests/test_core/test_sqlite_store.py::TestTestRunManagerIntegration::test_storage_failure_is_a_warning_not_an_error` asserts `"disk I/O error" in capsys.readouterr().err`, and the warning it checks is emitted through rich. From `gh run view 35372430808 --log-failed` (main, `23901cf`, 18 Sep):

```
E  AssertionError: assert 'disk I/O error' in 'Warning: failed to save test run to \n/tmp/pytest-of-runner/pytest-0/test_storage_failure_is_a_warn0/deepeval.db: disk\nI/O error\n'
!!!! stopping after 1 failures !!!!!!!!!!!!!!!!!!!!!!!!!!
====== 1 failed, 1059 passed, 1 skipped, 48 warnings in 153.64s (0:02:33) ======
```

The searched phrase is split by the fold, so the test is red on GitHub's runners. A developer whose temp path is longer sees the fold land past the phrase and the same assertion passes — as it does on my macOS machine. The verdict is decided by the length of `tmp_path`, not by the code under test.

### The change

`print(table)` at line 918 was the only call that needed rich, so it becomes `console.print(table)` on the `Console` this module already constructs at line 70, and the shadowing import goes away. The 13 plain-text messages now reach the stream exactly as written. No test file is edited: the existing assertion is the regression test, and it now tests the behaviour it names.

[...]

PR #3318, opened 2026-09-19
fix(dataset): keep a ConversationalGolden's Persona through the CSV round-trip

feiiiiii5:
## The bug

`save_as()` carries a `ConversationalGolden`'s `Persona` through the **json** and **jsonl** writers (`"persona": serialize_persona(golden.persona)` in both branches of `save_as`), and `add_goldens_from_json_file` reads it back with `persona_kwargs(json_obj.get(persona_key_name), user_description)`. The **csv** path had neither side: the multi-turn header had no `persona` column, and `add_goldens_from_csv_file` never looked for one.

So the same dataset stored as CSV instead of JSON silently loses the persona and falls back to the deprecated `user_description` string, with every other trait reset to its default (`deepeval/dataset/golden.py:90-99`).

Measured, writing one golden through `save_as("csv")` on `main` (`23901cf`) and on this commit (`ebba5ea`) and printing the header row:

```
main     header: ['scenario', 'turns', 'expected_outcome', 'user_description', 'context',
                  'name', 'comments', 'additional_metadata', 'custom_column_key_values']
         persona column: False

this PR  header: ['scenario', 'turns', 'expected_outcome', 'user_description', 'persona',
                  'context', 'name', 'comments', 'additional_metadata', 'custom_column_key_values']
         persona cell: {"name": "Ada", "characteristics": "blunt", "speaks_first": false,
                        "muted": true, "voice": "nova", "hold_timeout": 12.5}
```

The golden used above carried `name="Ada", characteristics="blunt", voice="nova", speaks_first=False, muted=True, hold_timeout=12.5`. Saving and immediately reloading it on `main` — same process, same golden, only the format differs:

```
json persona: Persona(name='Ada',   characteristics='blunt', interruption_behavior=None,
                      speaks_first=False, muted=True,  voice='nova', background_noise=None,
                      multilingual_stt=False, hold_timeout=12.5)
csv  persona: Persona(name=None,    characteristics='blunt', interruption_behavior=None,
                      speaks_first=True,   muted=False, voice=None,  background_noise=None,
                      multilingual_stt=False, hold_timeout=None)
```

Only `characteristics` survives the CSV path, because `user_description` is the one field the writer kept emitting; `speaks_first` and `muted` come back inverted relative to what was configured. The regression test's failure message on `main` says the same thing:

```
AssertionError: csv
assert Persona(name=..._timeout=None) == Persona(name=..._timeout=12.5)
```

## Why it matters

`Persona` is the simulator's input, not display metadata. `ConversationSimulator` reads these fields directly:

| field read | location |
| --- | --- |
| `persona.muted`, `persona.hold_timeout` | `deepeval/simulator/conversation_simulator.py:659-660` |
| `not persona.speaks_first` | `deepeval/simulator/conversation_simulator.py:663` |
| `persona.multilingual_stt` | `deepeval/simulator/conversation_simulator.py:906` |
| `golden.persona.interruption_behavior` | `deepeval/simulator/conversation_simulator.py:1053` |

A CSV-stored voice test set therefore runs a different conversation than the one that was configured — the agent opens instead of the caller, a "muted" caller starts speaking, and no one hangs up on a silent agent — and neither `save_as` nor the loader reports the loss. The TypeScript port's row shape already carries the field (`typescript/src/dataset/dataset.ts:164`), so the Python CSV writer is the outlier rather than the intent.

## The fix

Both directions reuse the helpers the json path already uses, so no new serialization format is introduced:

- **writer** — add `"persona"` to the multi-turn header, and a `json.dumps(serialize_persona(golden.persona), ensure_ascii=False)` cell that is `None` when the golden has no persona, mirroring `additional_metadata` right beside it.
- **reader** — `persona_col_name: Optional[str] = "persona"`, parsed with the existing `_parse_column` + `parse_mapping`, and `**persona_kwargs(persona, user_description)` in place of `user_description=user_description`, exactly as `add_goldens_from_json_file` does.

Old files keep loading as before: `_parse_column` yields `None` for every row when the column is absent, and `persona_kwargs(None, user_description)` returns `{"user_description": ...}`, which `ConversationalGolden.sync_persona` upgrades into a `Persona`. `test_csv_without_a_persona_column_keeps_user_description` pins that; it passes on `main` too, so it guards the fix rather than demonstrating the bug.

`serialize_persona` dumps with `exclude_defaults=True`, so rows only carry what the caller actually set, and a persona of pure defaults round-trips as `{}` → an equal default `Persona`.

## Verification

Environment: macOS arm64, Python 3.11.15, pytest 8.4.2, `deepeval` imported from the checkout root (not installed), no `OPENAI_API_KEY` in the environment. Since this touches `pd.read_csv`, I ran the same comparison twice — once with the pandas installed locally (3.0.5) and once shadowed to the version in `poetry.lock` (2.3.3):

```
cd <checkout root> && python3.11 -m pytest tests/test_core/test_datasets/test_dataset.py -q \
  --no-header -p no:cacheprovider -p no:opik --capture=tee-sys -W "ignore::DeprecationWarning"
```

(`-p no:opik` is only my machine's doing: the interpreter I share has an unrelated `opik` pytest plugin that aborts collection startup here. Drop it on a clean env.)

| tree | pandas 3.0.5 | pandas 2.3.3 |
| --- | --- | --- |
| `main` = `23901cf`, clean, plus the two new tests | `1 failed, 23 passed` (rc=1) | `1 failed, 23 passed` (rc=1) |
| `ebba5ea` (this PR) | `24 passed` (rc=0) | `24 passed` (rc=0) |

The single base failure is `test_save_as_round_trips_persona`, failing on the `csv` parameter only — json and jsonl already passed there, which is what makes this a CSV-writer/reader gap rather than a persona-model bug.

Wider, on this commit:

- `pytest tests/test_core/test_datasets` → `39 passed` (rc=0) — no regressions in the dataset loaders, including the existing `convo_goldens.csv` fixture test.
- `pytest tests/test_core/test_simulator` → `51 passed, 8 skipped` (rc=0) — the persona consumers still behave; I did not measure this directory on `main`, so I am reporting it as "no failures after the change", not as a fail-set comparison.
- `black --check` with **25.12.0**, the version pinned in `.github/workflows/black.yml` (my default local black is 24.1.0, so I installed the pin in a throwaway venv): `2 files would be left unchanged` (rc=0).
- Reviewed `git show ebba5ea` — 2 files, +75/−2, no other module touched.

## CI state

Pending as of posting. This PR comes from a fork, and `CONTRIBUTING.md` states that Actions "are mostly for internal use and will only pass if triggered by a user with the correct permissions within Confident AI". For reference, `test_core.yml` does include `tests/test_core/test_datasets/` in its "no secrets" dev-tests step, so the new tests are in the path a triggered run would exercise. I will not describe any check here as green.

## Not covered here

Three open PRs also touch `add_goldens_from_csv_file` (`#3298` literal `NA`, `#3233` token usage fields, `#3083` delimiter-separated tools columns). I searched issues and open PRs for a persona/CSV report before writing this and found none, and none of those three covers the persona column — different fields, different root causes, so I left them alone rather than folding them in.


feiiiiii5:
### CI state on this head (`ebba5ea`), read from the runs rather than guessed

| workflow | result | note |
| --- | --- | --- |
| PR title check | success | conventional `fix(dataset)` title |
| Py Confident Tests / Py Metrics Tests / Py Integration | success | |
| Py Lint Lint Lint (black) | failure | pre-existing on `main`, not this diff |
| Py Core Tests | failure | pre-existing on `main`, not this diff |

Neither red check is caused by this PR, and neither is my test suite failing. I verified both locally.

**Lint.** `.github/workflows/black.yml` runs `black --check` over `src: "."` with the pin `25.12.0`. I installed that exact version in a throwaway venv and ran it over the whole repo on both sides:

```
base 23901cf (untouched worktree) : would reformat a.py, deepeval/cli/inspect.py
                                    2 files would be reformatted, 1074 left unchanged  (rc=1)
head ebba5ea (this PR)             : same two files, same counts                        (rc=1)
```

So the job is red for every PR, including ones that change nothing. Neither of the two offending files is touched here; `deepeval/dataset/dataset.py` and `tests/test_core/test_datasets/test_dataset.py` are both clean under 25.12.0. (`a.py` is tracked at the repository root; `git log -1 -- a.py` gives `a3bab5d`, 2026-09-14.)

**Core tests.** `Py Core Tests` → step "Run tests (no secrets)" ends `1 failed, 1026 passed, 29 skipped`; the single failure is `tests/test_core/test_sqlite_store.py::TestTestRunManagerIntegration::test_storage_failure_is_a_warning_not_an_error`. It is the same defect #3316 fixes: the warning at `deepeval/test_run/test_run.py:1131` is printed through `from rich import print`, so rich folds it at the console width and the assertion `"disk I/O error" in ...err` no longer matches a string split into `disk I/O \nerror`. Whether it breaks depends on the length of `tmp_path`, which is why it shows up on CI and not on every machine. Reproduced with a CI-shaped short basetemp (my `tmp_path` = 57 chars; CI's logged path is 62):

```
base 23901cf            : FAILED  assert 'disk I/O error' in 'Warning: failed to save test run to \n/private/tmp/pr0/.../deepeval.db: disk I/O \nerror\n'
#3316 head 10e8288      : 1 passed in 0.73s        (same command, same basetemp)
```

Because that step runs with `--maxfail=1`, the later "Run dev tests (no secrets)" step — the one that includes `tests/test_core/test_datasets/` — was reported skipped, so **the two tests added here have not actually executed on CI**. Locally they are `1 failed, 23 passed` on `main` and `24 passed` on this branch, under both pandas 3.0.5 and the `poetry.lock` pin 2.3.3. If #3316 lands first, this PR's core step should go green on its own; otherwise the datasets step still needs a maintainer-triggered run (or the full-suite dispatch) to be seen here.

PR #3320, opened 2026-09-19
docs(cli): fix provider and login examples the CLI rejects

feiiiiii5:
## What breaks

Six copy-pasteable `deepeval` examples in the docs are commands the CLI (or the shell) refuses. Reproduced against `main` (`23901cf`) with the CLI's own parameter tree and with `bash`:

**1. `--temperature` on four provider commands that never declare it.** `deepeval/cli/providers/llms/openrouter.py:43` defines the option for `set-openrouter`; deepseek, grok, moonshot and portkey do not define it, so the documented command exits 2 before doing anything:

```
rc=2 | deepeval set-deepseek --model=deepseek-chat --temperature=0
rc=2 | deepeval set-grok --model grok-4.1 --temperature=0
rc=2 | deepeval set-moonshot --model=x --temperature=0
rc=2 | deepeval set-portkey --model m --provider p --base-url u --temperature=0
```

**2. `deepeval login --confident-api-key` in the setup tutorial.** `login`'s options are `--save/-s` and `--api-key`, and click says so itself:

```
No such option '--confident-api-key'. Did you mean '--api-key'?
```

**3. A comment after a line continuation kills the Portkey and OpenRouter blocks.** `\` only joins lines when it is the last character on the line, so bash ends the command at `# Ex: ...` and runs the remaining lines as separate commands (`bash -c` of the block exactly as written in the docs):

```
ARGS:[set-portkey --model your-model  # Ex: gpt-4.1] | bash: line 3: --provider: command not found  (rc=127)
ARGS:[set-openrouter --model openai/gpt-4.1  # Ex: ...] | bash: line 3: --base-url: command not found  (rc=127)
```

Note the first line is already broken even for someone who fixes nothing else: `#`, `Ex:` and `gpt-4.1` are passed to `set-portkey` as three extra arguments.

## The fix

- `deepseek.mdx`, `grok.mdx`, `moonshot.mdx`: drop the unsupported `--temperature=0` line. The pages still document the in-code `temperature=` argument and its fallback to the global `TEMPERATURE` setting (`deepeval/config/settings.py:478`), which is what `set-openrouter --temperature` writes (`TEMPERATURE=0.0` in the saved dotenv) — so no capability claim is removed, only a CLI flag that does not exist.
- `portkey.mdx`: same, plus the `\ # Ex:` comments become the values they were illustrating, so the block runs as one command.
- `openrouter.mdx`: only the `\ # Ex:` comment removed; `--temperature` and `--prompt-api-key` are real options and stay.
- `tutorial-setup.mdx`: `--confident-api-key` → `--api-key` (both occurrences; the surrounding text about what gets saved was already correct).

6 files, +9/−13, docs only.

## Verification

Environment: macOS arm64, Python 3.11.15, typer/click as installed by the repo's env, `deepeval` imported from the checkout root; each CLI run used an isolated `HOME` and `--save dotenv:<temp path>` so nothing touched my real settings file.

Corrected commands run end to end:

```
rc=0 | deepeval set-deepseek  --model=deepseek-chat --save dotenv:...
rc=0 | deepeval set-grok      --model grok-4.1 --save dotenv:...
rc=0 | deepeval set-moonshot  --model=kimi-k2-0711-preview --save dotenv:...
rc=0 | deepeval set-portkey   --model gpt-4.1 --provider openai --base-url https://gateway.portkey.ai/v1 --save dotenv:...
rc=0 | deepeval set-openrouter --model openai/gpt-4.1 --base-url https://openrouter.ai/api/v1 --temperature=0 --save dotenv:...
```

and they write what the page promises (`DEEPSEEK_MODEL_NAME`, `GROK_MODEL_NAME`, `MOONSHOT_MODEL_NAME`, `PORTKEY_MODEL_NAME/PROVIDER_NAME/BASE_URL`, `OPENROUTER_MODEL_NAME/BASE_URL`, `TEMPERATURE=0.0`). For `login` only the parse step is exercised — actually logging in needs a real key: `login.parse_args(["--api-key","x"])` succeeds where `--confident-api-key` raises.

Every edited fenced block was also re-extracted from the file and dry-run under `bash` with a stub `deepeval`, which now yields the intended single argument vector, e.g. `set-portkey --model gpt-4.1 --provider openai --base-url your-portkey-base-url` (rc=0).

Finally, a per-command audit of all 460 bash blocks / 224 `deepeval` lines in `docs/` and `skills/` (flags matched against click's option list *for that command*, not a global pool): 5 unknown-flag findings on `main`, 0 after this change, with no new findings introduced.

## Not changed, and why

- `deepeval recommend metrics` appears in the "Not sure if this metric is for you?" block on `metrics-contextual-recall.mdx`, `metrics-contextual-relevancy.mdx` and `metrics-prompt-alignment.mdx`. `recommend` is not a command in this CLI at all (rc=2 `No such command 'recommend'`), and `git log -S` finds no trace of it ever having been registered. Because that is a product decision — add the command or drop the block from all three pages — I left the pages alone rather than delete content on my own judgment. Happy to do either follow-up in a separate PR if you say which way.
- `skills/deepeval/**` and its published copies under `docs/public/.well-known/agent-skills/`: their `test run` flags (`--identifier`, `--num-processes`, `--ignore-errors`, `--skip-on-missing-params`, `--official`) all exist on `deepeval test run`, and `--tb=short` appears after `--` where it belongs to pytest. Nothing to correct, so no regeneration and no `agent_skills_check` impact.

## Checks

No Python changed, so `black` has nothing to check in this diff. My previous deepeval PR's workflows ran without a gating step, so I expect these checks to run too; whatever they report, I will state its actual outcome rather than call a pending or skipped job a pass.


feiiiiii5:
### CI state at this head (`883645f`), read from the runs

| workflow | status | conclusion |
| --- | --- | --- |
| PR title check | completed | success |
| Py Confident Tests | completed | success |
| Py Core Tests | in_progress | - |
| Py Integration | completed | success |
| Py Lint Lint Lint | completed | failure |
| Py Metrics Tests | queued | - |

Two things worth reading off that table rather than from the check names, which collide:

**The `Py Lint Lint Lint` failure is pre-existing and unrelated to this diff.** That job runs `black --check` over `src: "."` with the pinned `25.12.0`; on the untouched base commit `23901cf` it reports the same two files (`a.py` at the repo root and `deepeval/cli/inspect.py`), and the same on this head — I measured both with that exact version in a throwaway venv while working on #3318, where I posted the numbers. This PR changes only `.mdx` files, which black does not read, so nothing here can make that job green or redder.

**`Py Core Tests` did not exercise anything I changed either way** (its steps are `pytest tests/test_core/`), but it is worth noting that the "Run dev tests" step, which would cover `tests/test_core/test_datasets/`, is skipped whenever the earlier step aborts under `--maxfail=1` — that is why #3318's new tests are still unexecuted on CI, and the reason I am not calling any of these jobs evidence for a change I did not make here.

Once the remaining `Py Core Tests` job settles I will report what it says; if it is the same pre-existing `test_sqlite_store.py` failure documented on #3318, I will say so rather than re-derive it here.

Issue #3283, opened 2026-09-14
Verbose judge verdicts are scored as pass in role_violation, prompt_alignment, conversation_completeness


arian-gogani:
### What happens

Three metrics declare `verdict: str` rather than `Literal["yes", "no"]`, compare it with an exact match, and fall through to the permissive branch when it does not match. A judge that answers in a sentence rather than a bare token is scored as no problem found.

`deepeval/metrics/role_violation/role_violation.py:285-295`:

```python
for verdict in self.verdicts:
    if verdict.verdict.strip().lower() == "yes":
        return 0.0  # Role violation detected - no adherence
return 1.0  # No role violation - full adherence
```

Running that function as shipped, with its body lifted out of the file rather than retyped:

```
judge verdict string                                  score   meaning
  'yes'                                              0.0   VIOLATION CAUGHT
  'Yes'                                              0.0   VIOLATION CAUGHT
  'YES'                                              0.0   VIOLATION CAUGHT
  ' yes '                                            0.0   VIOLATION CAUGHT
  'Yes, the model violated its assigned role.'       1.0   scored as full adherence
  'yes - the assistant claimed to be a human doctor' 1.0   scored as full adherence
  'no'                                               1.0   scored as full adherence
  'No, no violation found.'                          1.0   scored as full adherence
```

`.strip().lower()` handles casing and whitespace, which is what makes this easy to miss: the obvious variants are covered, and only the verbose answer slips. A judge that explains its reasoning in the same field is read as reporting no violation.

### The three affected metrics

| metric | schema | comparison | permissive when unmatched |
|---|---|---|---|
| `role_violation` | `verdict: str` | `== "yes"` | scores 1.0, full adherence |
| `prompt_alignment` | `verdict: str` | `== "no"` | counted as aligned |
| `conversation_completeness` | `verdict: str` | `== "no"` | counted as complete |

`toxicity`, `bias` and `hallucination` declare `verdict: Literal["yes", "no"]`, so pydantic rejects a verbose verdict there and the failure is loud. The three above are the ones typed as bare `str`.

### deepeval already ships the fix

`deepeval/metrics/utils.py:514` has `verdict_from_json`, and its docstring describes this exact failure:

> Judges don't always honor a strict yes/no instruction, a non-native model may reply "No, it's off-topic". This normalizes the verbose reply to its leading word and drops anything outside `allowed` to `None` (so the caller can exclude it from scoring) rather than crashing a `Literal` schema or silently miscounting it.

[...]

PR #3330, opened 2026-09-21
fix: validate verdict count in PromptAlignmentMetric

### LINKED ISSUE #3329: PromptAlignmentMetric accepts missing judge verdicts as perfect alignment
**Describe the bug**
Reproduced on `23901cfc5fe4f61c3ca745f67748e00699d0484f` and the official PyPI 4.2.3 wheel. With three `prompt_instructions`, a judge reply of `{"verdicts": []}` yields score 1 and passes `assert_test()` even in `strict_mode`. A single `yes` verdict for three instructions also yields 1 and passes.

The verdicts template requires one verdict per instruction, but the schema permits a list of any length. `_calculate_score()` returns 1 for an empty list and otherwise divides by the number of verdicts returned. Missing assessments are silently accepted.

**To Reproduce**
1. `PromptAlignmentMetric(prompt_instructions=[three instructions], strict_mode=True, include_reason=True)` with a deterministic `DeepEvalBaseLLM` mock as the model.
2. Have the mock return `{"verdicts": []}` (or a single `{"verdict": "yes"}`) for the `Verdicts` schema and any reason text for `PromptAlignmentScoreReason`.
3. Call `measure()` or `assert_test()`, sync or async, with raw JSON or schema responses.
4. Score is 1 and the test passes.

Complete all-`no` verdicts correctly fail; complete all-`yes` controls pass. No live judge calls were made; the frequency of malformed responses in production is not measured.

**Expected behavior**
A verdict count that does not match the instruction count is a malformed judge response and should be rejected the same way invalid judge JSON is: set `metric.error` and raise, so that with `ignore_errors=True` the metric is recorded as errored rather than passing.

**Additional context**
20 regression cases (12 count-validation failures for 0, 1 and 4 verdicts against 3 instructions across sync/async and JSON/schema, plus 8 complete-result controls) fail 12 and pass 8 on both main and the 4.2.3 wheel. A cardinality guard that records `metric.error` and raises `ValueError` before scoring makes all 20 pass. A reproduction script and the 20 regression tests are in the accompanying PR.

This is distinct from #3283 / #3285: those normalize verbose verdict strings. This reproduction uses canonical yes/no tokens or an empty list, and that patch does not validate cardinality.

I have the fix with regression tests ready and am opening a PR for it.

mavericksea-ai:
Fixes #3329

The verdicts template asks the judge for one verdict per prompt instruction, but `paschema.Verdicts` accepts a list of any length. `_calculate_score()` returned 1 for an empty list and otherwise divided by the number of verdicts returned. With three prompt_instructions, `{"verdicts": []}` or a single yes verdict scored 1 and passed `assert_test()` even in strict_mode, and four verdicts for three instructions were scored without complaint.

`_calculate_score()` now checks that the verdict count matches `len(self.prompt_instructions)`. On a mismatch it sets `metric.error` and raises `ValueError`, the same convention `trimAndLoadJson()` uses for invalid judge JSON. With `ignore_errors=True` the runner records the metric as errored with `success=False`, as it already does for invalid JSON. The empty-list branch is removed; `prompt_instructions` cannot be empty (the constructor rejects it), so a matching count is never zero.

Adds `tests/test_metrics/test_prompt_alignment_verdict_count.py`: 20 offline cases using a fixed-verdict stub model, so no API key is needed. They cover sync and async execution with both schema and raw JSON judge responses. 12 cases return 0, 1 or 4 verdicts for 3 instructions and must raise. 8 controls return complete all-yes (must pass) or all-no (must fail) verdicts.

Before, on `main` at `23901cf`: 12 failed, 8 passed. After: 20 passed.

This is distinct from #3283/#3285, which normalize verbose verdict strings; these cases use canonical yes/no tokens or an empty list, and that change does not validate cardinality.


mavericksea-ai:
Rebased onto current main. The metric moved to the shared score_qag_verdicts helper since this PR was opened, so the fix has been rebuilt on top of that rather than reverting the refactor. The guard now sits at the PromptAlignmentMetric boundary: _calculate_score records metric.error and raises ValueError when the judge returns a number of verdicts that does not match the number of prompt instructions. score_qag_verdicts itself is untouched, so its behaviour for the other callers is unchanged.

Tests: 20 pass with the fix and 12 of those fail without it, on main at fa6d875. They use a fixed-verdict stub model and make no network calls, covering sync and async execution and both schema and raw JSON judge responses.

Worth a separate look, and not something I have touched here: empty_score defaults to 1 for every caller of score_qag_verdicts, which is 22 metrics. Whether an empty verdict set should read as a perfect score seems like a decision worth making deliberately rather than inheriting from a default.

mavericksea-ai:
Rebased onto current main. The metric moved to the shared score_qag_verdicts helper since this PR was opened, so the fix has been rebuilt on top of that rather than reverting the refactor. The guard now sits at the PromptAlignmentMetric boundary: _calculate_score records metric.error and raises ValueError when the judge returns a number of verdicts that does not match the number of prompt instructions. score_qag_verdicts itself is untouched, so its behaviour for the other callers is unchanged.

Tests: 24 pass with the fix and 16 of those fail without it, on main at fa6d875. They use a fixed-verdict stub model and make no network calls, covering sync and async execution and both schema and raw JSON judge responses. Four of them cover ignore_errors=True, asserting that a mismatched count is recorded as errored with success=False rather than read as a pass.

On the red Py Core Tests check: the failure is test_core/test_sqlite_store.py::TestTestRunManagerIntegration::test_storage_failure_is_a_warning_not_an_error, which asserts "disk I/O error" appears in stderr while the captured warning has the message wrapped across two lines. It fails the same way on main at fa6d875 (run #5075), so it is pre-existing here and not something this branch introduces.

Worth a separate look, and not something I have touched here: empty_score defaults to 1 for every caller of score_qag_verdicts, which is 22 metrics. Whether an empty verdict set should read as a perfect score seems like a decision worth making deliberately rather than inheriting from a default.

